PR #73, opened 2021-05-24
New feature for more access within QUA of baked waveforms 

arthurostrauss:
This PR proposes a series of getters (operation name and length) to retrieve easily the baked operations within QUA and tune them in real time (amplitude, duration, truncation modulations for example).
This feature is illustrated well with the new version of the 1 qubit RB using the baking and the Cayley table for the Clifford group to recover efficiently the inverse operation.


bakery/RB_1qb_baking_2.py
```diff
@@ -0,0 +1,88 @@
+from qm import SimulationConfig
+from qm.QmJob import QmJob
+from qm.QuantumMachinesManager import QuantumMachinesManager
+from RB_1qb_configuration import config, pulse_len
+from rb_1_with_table_guide import *
+import matplotlib.pyplot as plt
+
+d_max = 20  # Maximum RB sequence length
```
liorella-qm, line 8:
```suggestion
d_max = 20  # Maximum RB sequence length
```

liorella-qm, line 8: all of this up to line 44 should be refactored into a class that the user inits with d_max, K, the name of the qubit, the name of the ops, and then a class is created which contains everything he needs in order to run the program: K_list, duration tracker, revert op etc.

bakery/RB_1qb_baking_2.py
```diff
@@ -0,0 +1,88 @@
+from qm import SimulationConfig
+from qm.QmJob import QmJob
+from qm.QuantumMachinesManager import QuantumMachinesManager
+from RB_1qb_configuration import config, pulse_len
+from rb_1_with_table_guide import *
```
liorella-qm, line 5: please avoid import * here because it's hard to know what is being used from there (makes the code less readable)

bakery/rb_1_with_table_guide.py
```diff
@@ -0,0 +1,78 @@
+import pandas as pd
```
liorella-qm, line 1: what is the purpose of this file? Is it a sort of library for 1 qubit RB? If so let's create a python file which is really a 1 qubit RB library which allows the user to import everything he needs in order to run 1 qubit RB in an easy manner. This then should not include functions like "measure state" which are not related to the RB itself.

bakery/RB_1qb_baking_2.py
```diff
@@ -0,0 +1,88 @@
...
+import matplotlib.pyplot as plt
```
liorella-qm, line 6: this file should be an example file for usage of the library file (see comment on rb_1_with_table_guide.py

liorella-qm:
hi Arthur,

I have several general comments:

1. there is no clear separation between the code used for the baking and the code that runs specific things with the baking. We want the baking to become a packaged feature and so we have to separate this very clearly. Think about what you want to be packaged, put it in a separate folder and import from there. All the scripts that you import to should be in a seprate folder of "applications of baking". (Do it however you think is right...)
2. the purpose of benchmarking is to benchmark existing things. Think about "benchmark" in the context of testing which computer is fastest - you don't build a computer from scratch but take an exisitng one and benchmark its performance. The same should go for gates with RB. That's why we need a 1 qubit RB library as I explained in the comments.

```
modified_config,  = create_rb_sequences(existing_config, 
```

bakery/RB_1qb_baking_2.py
arthurostrauss: Import is now more explicit

bakery/RB_1qb_baking_2.py
arthurostrauss: Details of the class to be determined/fixed : 
- how to cope with configuration files (so far we stay in the spirit of the baking and update the input config with all necessary addings, like a black box)
- ensure the writing of the convention in the config of existing operations to be able to play seamlessly all Cliffords (basic operations should for now be given with the right name X, Y, X/2,...

bakery/baking_usage_examples/RB_1_qubit/RB_utils.py
```diff
@@ -0,0 +1,110 @@
...
+
+]
+
+# Cayley table corresponding to above Clifford group structure
+
+c1_table = pd.read_csv('c1_cayley_table.csv').to_numpy()[:, 1:]
+
+
+class RB_one_qubit:
+    def __init__(self, config: dict, d_max: int, K: int, qubit: str, cliffords: List = None,
```
liorella-qm, line 42: not sure I understand how cliffords can be a list. What is the meaning of every elementn in the list? i.e how do I know that entry zero corresponds to 'X', entry 1 to 'Y' and so on? I think it should be a dict no? e.g:
```
RBOneQubit(..., cliffords = {'X': 'my_x', 'Y': 'my_y', ....} )
```

liorella-qm (changes requested):
hi some small changes, but overall looks much better!

bakery/baking_usage_examples/RB_1_qubit/RB_utils.py
arthurostrauss: cliffords should be a list whose index is related to which index should we look for in the provided Cayley table (we move from one Clifford to another, keeping track on the change using indices). For now it is true that we could move forward with a dictionary, but in the hope of moving everything to QUA later I do think we shall rely on this moving related to indices. Main issue with dict is that to my knowledge it has no numerical ordering (unless we do an enumerate but then we are coming back to a list).

bakery/baking_usage_examples/RB_1_qubit/RB_utils.py
```diff
@@ -0,0 +1,110 @@
...
+        Class to retrieve easily baked RB sequences and their inverse operations
+        :param config Configuration file
+        :param d_max Maximum length of desired RB sequence
+        :param K Number of RB sequences to be generated
+        :param qubit Name of the quantum element designating the qubit
+        :param cliffords List of Clifford operations chosen, by default c1_ops defined above
+        :param cayley_table Table indicating group structure associated to Cliffords provided
+        """
+
+        if cliffords is None:
```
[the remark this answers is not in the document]
arthurostrauss: Because Cliffords should be indexed according to what is inside the Cayley table, the two inputs should be related. So I can remove the possibility for the user to tune up its set of Cliffords, but I cannot give access to the Clifford set without knowing how to navigate through the Cayley table.

liorella-qm (approved):
looks very good!

PR #95, opened 2021-09-05
2 qubit RB example using baking

arthurostrauss:
This PR proposes a new class abstraction for the running of a full 2 qubit RB experiment. An example of how to run it is provided. The post-processing is currently not implemented, those scripts only do allow the realization of the experiment until the retrieval of experimental data.
According to estimations, the number of Cliffords runnable oscillates between 100 and 180 Cliffords, depending on the length of the different generators of the Clifford.


examples/randomized-benchmark/two-qubit-rb/readme.md
```diff
@@ -0,0 +1,86 @@
...
+The pattern behind RB is always the same: 
+- Generate a random sequence composed of Cliffords
+- Append to the sequence one Clifford reverting the action of the random sequence, such that the identity circuit is played
+- Measure the survival probability
+- Repeat prior steps for varying number of Cliffords within the sequence
+
+In order to have an easy way to derive all Cliffords of the two qubit Clifford group, 
+we use the decomposition illustrated in the following reference: https://arxiv.org/pdf/1210.7011.pdf.
+This decomposition allows an easy derivation of the group from the Clifford group for one qubit, and uses three main two qubit gates
+(that shall be decomposed later into a native set of gates by the user depending on its own hardware):
```
liorella-qm, line 25: their own hardware

examples/randomized-benchmark/two-qubit-rb/readme.md
```diff
@@ -0,0 +1,86 @@
...
+- CNOT
+- SWAP
```
liorella-qm, line 27: who is playing SWAP? As far as I know there is no architecutre that can directly implement SWAP, if there is I'd be happy to know

examples/randomized-benchmark/two-qubit-rb/readme.md
```diff
@@ -0,0 +1,86 @@
...
+- iSWAP
+
```
liorella-qm, line 29: what about CPHASE?

examples/randomized-benchmark/two-qubit-rb/readme.md
```diff
@@ -0,0 +1,86 @@
...
+- a *config* dictionary, used to open a Quantum Machine and to store the waveforms randomly generated
+- *N_Clifford*, the number of desired Clifford per sequence. Here an array of different values can be provided to retrieve different truncations
+- *N_sequences*: Number of random sequences to be generated for averaging
+- *two_qb_gate_macros*: Dictionary containing macros taking a baking object as input and using
+ baking play statements on necessary quantum elements to perform the three 2 qubit gates mentioned above ("CNOT", "SWAP" and "iSWAP" which are also keys of the dictionary).  The native two-qubit gate macro shall always contain a baking align statement before and after the playing statements
+- *single_qb_macros*: Optional dictionary in the case where a single qubit gate is not simply a QUA play statement on one quantum element, one can specify a dictionary of 
+a set of macros. Note that the dictionary shall then contain macros for all single qubit generators, i.e "I", "X", "Y", "X/2", "Y/2", "-X/2", "-Y/2".
+- *qubit_register*: Tuple containing names of two target qubits
```
liorella-qm, line 40: in what format are those defined? We don't have a definition of a "qubit" in QUA

examples/randomized-benchmark/two-qubit-rb/readme.md
```diff
@@ -0,0 +1,86 @@
...
+- *single_qb_macros*: Optional dictionary in the case where a single qubit gate is not simply a QUA play statement on one quantum element, one can specify a dictionary of 
+a set of macros. Note that the dictionary shall then contain macros for all single qubit generators, i.e "I", "X", "Y", "X/2", "Y/2", "-X/2", "-Y/2".
+- *qubit_register*: Tuple containing names of two target qubits 
+- *seed*: Random seed
+
+
+# Run the experiment
+Before running the experiment, the user shall define all baking macros for computing Clifford generators for each of the two target qubits specified in the *qubit_register*.
+To help the user retrieve the correct quantum elements to be addressed when specifying one of the two qubits, we propose the implementation of the *Resolver* class, which aims to facilitate the retrieval of various quantum elements that might need to be involved for computing a specific quantum gate on the "logical qubits".
```
liorella-qm, line 46: "logical qubits" -> qubits

examples/randomized-benchmark/two-qubit-rb/readme.md
```diff
@@ -0,0 +1,86 @@
...
+One could in fact see a qubit as a set of quantum elements in the configuration, that do depend on the operation to be played on. Recall that within the *RBTwoQubits* class, the macros will be called with the following arguments: the baking object of reference (which shall always be 
+a parameter of the baking macros, as shown in the example), and the target qubit(s) (the two names given in the *qubit_register* argument of the class) on which the macros shall be applied. Single qubit gate macros only require one target qubit out of the two provided, whereas the two_qubit_gate macros require the two qubits of the register (a control and a target, always chosen to be in the same order for this experiment).
+
+The example shows briefly how one can run the experiment using the class. The method *run* uses one single argument,
+which is the main QUA program the user should define with its own measurement commands and its own stream_processing, such that the retrieval and fitting of results can be done directly. 
+The user shall build its QUA program according to two important guidelines:
```
liorella-qm, line 52: there is a very differnet writing style here than in other places in qua-libs. For example, Gal's writing style is much more informal. I suggest discussing with him about the tone

examples/randomized-benchmark/two-qubit-rb/example_2qb_rb.py
```diff
@@ -0,0 +1,232 @@
...
+from qm.QuantumMachinesManager import SimulationConfig
+from entropylab_qpudb import Resolver
+
+qmm = QuantumMachinesManager()
+
+"""
+For the required two qubit gates to generate the 4 classes (see Supplementary info of this paper:
+https://arxiv.org/pdf/1210.7011.pdf), we require the user to complete the following macros below according
+to their own native set of qubit gates, that is perform the appropriate decomposition and convert the pulse sequence
+in a sequence of baking play statements (amounts to similar structure as QUA, just add the prefix b. before every statement)
```
liorella-qm, line 12: in -> into

examples/randomized-benchmark/two-qubit-rb/example_2qb_rb.py
```diff
@@ -0,0 +1,232 @@
...
+In this example, we have the simplest case where "q0" and "q1" are defined as the two qubits directly in the config,
+and there is one single coupler named "coupler". Readout resonators coupled to each qubit are named "rr1" and "rr2"
+"""
+
+
+class RBTwoQubitResolver(Resolver):
+    def __init__(self, aliases: Union[dict, None] = None):
+        super().__init__(aliases)
+
+    def q(self, qubit, channel=None):
```
liorella-qm, line 41: what does this function actually do? It seems to just return qubit

examples/randomized-benchmark/two-qubit-rb/example_2qb_rb.py
```diff
@@ -0,0 +1,232 @@
...
+            that might be required)
+            """
+            return qubit
+
+        elif qubit == "q1":
+            return qubit
+        else:
+            return qubit
+
+    def res(self, resonator):
```
liorella-qm, line 55: same question as above

examples/randomized-benchmark/two-qubit-rb/example_2qb_rb.py
```diff
@@ -0,0 +1,232 @@
...
+to the two qubits you want to benchmark.
+The Resolver allows you to retrieve easily the names of the quantum elements associated to your target qubits, couplers,
+and other elements you might need to carry (fluxlines, etc...).
+
+In this example, we have the simplest case where "q0" and "q1" are defined as the two qubits directly in the config,
+and there is one single coupler named "coupler". Readout resonators coupled to each qubit are named "rr1" and "rr2"
+"""
+
+
+class RBTwoQubitResolver(Resolver):
```
liorella-qm, line 37: the resolver should be specified for a QPU. I.e. if you have say a 10 qubit system with flux tuning and a parameteric coupler, you'd define two channels for the qubit ("xy" and "z"), RR, and a coupler. Then you'd use this mapping in your code.

In this example I'm not entirely sure what the resolver is assisting with

examples/randomized-benchmark/two-qubit-rb/example_2qb_rb.py
```diff
@@ -0,0 +1,232 @@
...
+
+# Baking Macros required for two qubit gates
+
+# Here is my assumed native two qubit gate
+
+
+resolve = RBTwoQubitResolver()
+
+
+def CZ(b_seq: Baking, ctrl: str = "q0", tgt: str = "q1"):
```
liorella-qm, line 80: why provide here default arguments?

examples/randomized-benchmark/two-qubit-rb/example_2qb_rb.py
```diff
@@ -0,0 +1,232 @@
...
+
+
+def SWAP(b_seq: Baking, ctrl: str = "q0", tgt: str = "q1"):
+    CNOT(b_seq, ctrl, tgt)
+    CNOT(b_seq, tgt, ctrl)
+    CNOT(b_seq, ctrl, tgt)
+
+
+"""
+In  what follows, q_tgt should be the main target qubit for which should be played the single qubit gate.
```
liorella-qm, line 118: not sure I understand this comment

examples/randomized-benchmark/two-qubit-rb/example_2qb_rb.py
```diff
@@ -0,0 +1,232 @@
...
+
+def mX_2(b: Baking, q_tgt):
+    b.play("-X/2", resolve.q(q_tgt))
+
+
+def mY_2(b: Baking, q_tgt):
+    b.play("-Y/2", resolve.q(q_tgt))
+
+
+two_qb_gate_macros = {
```
liorella-qm, line 152: how do I know if I implemented all the required gates? Perhaps this should be a dataclass or namedtuple which forces the user to provide all required arguments when instantiating?

examples/randomized-benchmark/two-qubit-rb/example_2qb_rb.py
```diff
@@ -0,0 +1,232 @@
...
+    "CNOT": CNOT,
+    "iSWAP": iSWAP,
+    "SWAP": SWAP
+}
+
+single_qb_gate_macros = {
```
liorella-qm, line 158: same comment as for `two_qb_gate_macros`

liorella-qm (changes requested):
Only looked at interface, did not check implementation.

very good work overall.

Please see my comments.

examples/randomized-benchmark/two-qubit-rb/readme.md
arthurostrauss: Assuming no one can, the idea is that those gates are the ones allowing the easy class decomposition depicted in the reference given. That is, regardless of two qubit native gates that the user wants to implement, the selection of the Clifford will always be done w.r.t to the class type defined by the presence of a CNOT/SWAP/iSWAP in the circuit. For this reason and to make the code hardware agnostic, we ask the user to provide a macro for each of those gates (if it is native to the hardware, it's the best but if it's not, he will simply have to play a bunch of single qubit gates as well to pass from one two qubit gate to another according to appropriate circuit identities). 
I will explain it better here

examples/randomized-benchmark/two-qubit-rb/readme.md
arthurostrauss: CPHASE is irrelevant, as it is not a gate that allows the selection of a Clifford according to its class type. In the example, we assume that CPHASE is the native gate, but the user shall put a macro for his own native two qubit gate (but not mandatory) and from there derive sequences of instructions in those three indicated baking macros to yield the intended gates that do create the class separation

examples/randomized-benchmark/two-qubit-rb/readme.md
```diff
@@ -0,0 +1,86 @@
...
+- CNOT
+- SWAP
+- iSWAP
+
+We introduce a specific class that allows an easy generation and execution of multiple random Clifford sequences based on 
+few generic inputs the user shall provide:
+- a  *QuantumMachinesManager* instance, which will be used to launch  the series of jobs necessary to play all synthezised sequences 
+- a *config* dictionary, used to open a Quantum Machine and to store the waveforms randomly generated
+- *N_Clifford*, the number of desired Clifford per sequence. Here an array of different values can be provided to retrieve different truncations
+- *N_sequences*: Number of random sequences to be generated for averaging
```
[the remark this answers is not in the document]
arthurostrauss: No, by N_Clifford you provide a set of truncation indices common to all random sequences

examples/randomized-benchmark/two-qubit-rb/readme.md
arthurostrauss: I know, it is defined to yield access through the Resolver to the specific quantum elements. How do you call "q0" when calling resolve.q("q0", channel="X)?

examples/randomized-benchmark/two-qubit-rb/readme.md
arthurostrauss: True, but usually readmes of other examples are not as long and detailed as here, maybe should we add it as a tool as we did with one Qb RB in py-qua-tools?

examples/randomized-benchmark/two-qubit-rb/readme.md
```diff
@@ -0,0 +1,86 @@
...
+Moreover, we use the *add_compiled* feature of the QOP to be able to load random sequences successively, without having to compile a new program at every iteration. This scheme involves two main limitations that are described below.
+
+## 1. Maximum number of Clifford limited on lengths of gates 
+The first consequence of this method is that the number of maximum Clifford operations playable in the experiment is limited by the waveform memory, and is therefore highly dependent on the length of the gates the user has calibrated.
+Assuming all single qubit gates have the same length, and using one single native two qubit gate with another (longer) length, we have performed the following approximative benchmark of how many Clifford operations are playable.
+Note that this number may slightly vary as all Clifford operations do not carry the same number of gates and there is therefore some randomness on the actual number that can be played. The benchmark done below indicates the maximum number of playable Clifford operations in one single random sequence with certainty.
```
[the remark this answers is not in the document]
arthurostrauss: I agree, but here we are not interested in the average quantity but rather the maximum amount of Cliffords you can play with certainty without saturation of waveform memory rather than on average.

examples/randomized-benchmark/two-qubit-rb/example_2qb_rb.py
arthurostrauss: Yes, for the case of our simple example it  simply t does hat, because our configuration is basically one qubit= one quantum element with the same name.
But the user shall redefine those methods according to his own setup and access the required quantum elements associated to his target qubit and the desired channel.

examples/randomized-benchmark/two-qubit-rb/example_2qb_rb.py
arthurostrauss: Same answer as above, simplicity of our example config yields no specific case for resolver usage. Maybe we should pick a more complex architecture to showcase better the resolver usage. We should discuss which architecture should be addressed (Google perhaps?)

examples/randomized-benchmark/two-qubit-rb/example_2qb_rb.py
arthurostrauss: Same remark as above, it is not helpful for the specific usecase, but it's put there as a generic expression for this "macro" conception which points out the need to map from circuit level first for Clifford selection before coming back to pulse level

examples/randomized-benchmark/two-qubit-rb/example_2qb_rb.py
arthurostrauss: To showcase what format should be put there, but I can remove it

examples/randomized-benchmark/two-qubit-rb/example_2qb_rb.py
arthurostrauss: What about "In  what follows, q_tgt is the name of the target qubit on which the gate is applied."
It just means it shall indicate the name of the "q0" or "q1" that you put as argument in the resolver through resolve.q("q0", ...). The name of this argument should be specified. Overall, we need a true documentation for the Resolver use case I figure.

examples/randomized-benchmark/two-qubit-rb/example_2qb_rb.py
arthurostrauss: I could add further assertion in the class to make sure the correct gates are provided. Not sure to understand what you mean by dataclass or namedtuple in this case.

PR #100, opened 2021-12-14
add examples for config builder tool

SatyaBade12:


examples/config-builder/resonator.py
```diff
@@ -0,0 +1,48 @@
...
+res.lo_frequency = 4e9
+
+wfs = [
+    ArbitraryWaveform("wf1", np.linspace(0, -0.5, 16).tolist()),
+    ArbitraryWaveform("wf2", np.linspace(0, -0.5, 16).tolist()),
+]
+
+ro_pulse = MeasurePulse("ro_pulse", wfs, 16)
+ro_pulse.add(
+    Weights(ConstantIntegrationWeights("integ_w1_I", cosine=1, sine=0, duration=16))
```
yomach, line 24: change integration weights names to:

cos
sin
minus_sin

examples/config-builder/resonator.py
```diff
@@ -0,0 +1,48 @@
...
+)
+ro_pulse.add(
+    Weights(ConstantIntegrationWeights("integ_w2_I", cosine=0, sine=1, duration=16))
+)
+ro_pulse.add(
+    Weights(ConstantIntegrationWeights("integ_w2_Q", cosine=1, sine=0, duration=16))
+)
+
+res.add(Operation(ro_pulse))
+
```
yomach, line 37: We need a few comments.
At this part, I forgot that the "cont" is the controller...

At the beginning, write something like:
First we would define the controller, elements and operations

Then here write something like:
Now that we have everything ready, we can create the configuration. First we create a ConfigBuilder object, and then we add the controller and elements to it.

examples/config-builder/resonator.py
```diff
@@ -0,0 +1,48 @@
...
+cb = ConfigBuilder()
+cb.add(cont)
+
+# here the two wfs are already added to resonator object, but still it is possible to add them to the setup
```
yomach, line 41: This part is confusing, why would we want to add them manually?

I would just remove this

examples/config-builder/three_qb_system.py
```diff
@@ -0,0 +1,317 @@
...
+from qualang_tools.config.configuration import *
+from qualang_tools.config.components import *
+from qualang_tools.config.builder import ConfigBuilder
+
+cnot_coupler = ArbitraryWaveform(
+    "cnot_coupler",
+    (
+        2.0
+        * np.pi
+        * np.array(
```
yomach, line 12: Err, I hate the way black handles this...
What are these numbers? We need to put here functions that create them and not the raw numbers

examples/config-builder/three_qb_system.py
```diff
@@ -0,0 +1,317 @@
...
+                -0.15096289009237282,
+                -0.110322391784403,
+                -0.08696095347060069,
+                -0.0801578668706695,
+                -0.07016164311782273,
+                -0.053739705459874915,
+                -0.035772043817697045,
+            ]
+        )
+    ).tolist(),
```
yomach, line 32: Also, why do we need tolist()? Can't the ArbitraryWaveform handle numpy?

examples/config-builder/three_qb_system.py
```diff
@@ -0,0 +1,317 @@
...
+                0.045723392279059444,
+                0.02312523234819297,
+                0.009823882427570495,
+                0.0019644568243603654,
+                -0.0008139980959453693,
+            ]
+        )
+    ).tolist(),
+)
+
```
yomach, line 199: The ordering here is different compared to the previous example...
I actually think this makes a bit more sense, but again, add a few comments to explain it.

yomach (changes requested):

examples/config-builder/resonator.py
SatyaBade12: is there any documentation available for `py-qua-tools` that I can refer here?

examples/config-builder/resonator.py
SatyaBade12: it was an example written for gui based config builder, there we need to support addition of any object, unnecessary here...

examples/config-builder/three_qb_system.py
SatyaBade12: the precise numbers are not important for this example, switching to random numbers

examples/config-builder/three_qb_system.py
SatyaBade12: GUI config builder required all waveforms to be in serializable JSON format so I intentionally converted them to python lists, ArbitraryWaveform can handle numpy arrays

examples/config-builder/resonator.py
yomach: There is nothing for py-qua-tools in general. Baking has this:
https://github.com/qua-platform/py-qua-tools/tree/main/qualang_tools/bakery
The only other "big" tool is the config builder...

SatyaBade12: I added few comments, but we need a detailed guide on "how to use ConfigBuilder?" in `py-qua-tools`

yomach: I agree...

yomach (approved):

examples/config-builder/resonator.py
SatyaBade12: there will be a new PR in py-qua-tools

PR #101, opened 2021-12-15
DRAG waveforms correction

KevinAVR:
I read the cited reference that found the DRAG waveforms had typos. I corrected them.

### DIFF
--- a/examples/randomized-benchmark/DRAG-optimization/configuration.py
+++ b/examples/randomized-benchmark/DRAG-optimization/configuration.py
@@ -1,25 +1,53 @@
 import numpy as np
 
-gauss_len = 100
+gauss_len = 16
 gaussian_amp = 0.2
 
 
 def gauss(amplitude, mu, sigma, delf, length):
-    t = np.linspace(-length / 2, length / 2, length)
+    t = np.linspace(0, length, length)
     gauss_wave = amplitude * np.exp(-((t - mu) ** 2) / (2 * sigma ** 2))
-    # Detuning correction Eqn. (4) in Chen et al. PRL, 116, 020501 (2016)
-    gauss_wave = gauss_wave * np.exp(2 * np.pi * delf * t)
     return [float(x) for x in gauss_wave]
 
 
 def gauss_der(amplitude, mu, sigma, delf, length):
-    t = np.linspace(-length / 2, length / 2, length)
+    t = np.linspace(0, length, length)
     gauss_der_wave = (
-        amplitude * (-2 * (t - mu)) * np.exp(-((t - mu) ** 2) / (2 * sigma ** 2))
+        amplitude
+        * (-2 * 1e9 * (t - mu) / (2 * sigma ** 2))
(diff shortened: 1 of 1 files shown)


examples/randomized-benchmark/DRAG-optimization/configuration.py
```diff
@@ -1,40 +1,72 @@
 import numpy as np
 
-gauss_len = 100
+gauss_len = 16
 gaussian_amp = 0.2
 
 
 def gauss(amplitude, mu, sigma, delf, length):
-    t = np.linspace(-length / 2, length / 2, length)
```
yomach, line 8: I think I like the symmetric version better, because then mu=0 means you get a normal gaussian. This is also the convention for scipy and other packages...
(I would actually remove the "mu" parameter, nobody ever changes it!)

But I'm open to discussion, convince me otherwise!

examples/randomized-benchmark/DRAG-optimization/configuration.py
```diff
@@ -1,40 +1,72 @@
 import numpy as np
 
-gauss_len = 100
+gauss_len = 16
```
yomach, line 3: why 16? because it is being scanned in the file?

examples/randomized-benchmark/DRAG-optimization/configuration.py
```diff
@@ -1,40 +1,72 @@
...
+        amplitude
+        * (-2 * 1e9 * (t - mu) / (2 * sigma ** 2))
+        * np.exp(-((t - mu) ** 2) / (2 * sigma ** 2))
     )
-    # Detuning correction Eqn. (4) in Chen et al. PRL, 116, 020501 (2016)
-    gauss_der_wave = gauss_der_wave * np.exp(2 * np.pi * delf * t)
     return [float(x) for x in gauss_der_wave]
 
 
+def gauss_det(amplitude, mu, sigma, delf, length):
```
yomach, line 23: Too many functions with confusing names!
gauss -> Ok, simple
gauss_der -> I guess derivative, OK.
gauss_det ??
gauss_der_det ??

What are they?
I would change the names to be longer (and informtive), and while you're at it, add docstrings!

yomach (changes requested):

examples/randomized-benchmark/DRAG-optimization/configuration.py
KevinAVR: I put a short one because usually DRAG pulses are used when the length is small, not really in the range of 100 ns

examples/randomized-benchmark/DRAG-optimization/configuration.py
KevinAVR: I was following what was reported in the paper Chen in the comments. It is really not relevant for gaussian, but for cosine you would need to do it from 0 to lenght

examples/randomized-benchmark/DRAG-optimization/configuration.py
KevinAVR: Improved, thanks!

PR #102, opened 2021-12-20
DRAG config updated with better comments and names

KevinAVR:
Improved things after your feedback! thanks

### DIFF
--- a/examples/randomized-benchmark/DRAG-optimization/configuration.py
+++ b/examples/randomized-benchmark/DRAG-optimization/configuration.py
@@ -24,39 +24,26 @@ def gauss_derivative(amplitude, sigma, length):
     return [float(x) for x in gauss_derivative_wave]
 
 
-# I-quadrature component for gaussian shaped detuned pulses
-def gauss_detuned(amplitude, sigma, del_f, length):
-    t = np.linspace(-length / 2, length / 2, length)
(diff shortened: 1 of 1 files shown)


examples/randomized-benchmark/DRAG-optimization/configuration.py
```diff
@@ -1,40 +1,83 @@
...
-gauss_len = 100
-gaussian_amp = 0.2
+gauss_len = 16  # pulse length in ns
+gaussian_amp = 0.2  # pulse amplitude in volts
 
+# Definition of pulses follow Chen et al. PRL, 116, 020501 (2016)
 
-def gauss(amplitude, mu, sigma, delf, length):
+
+# I-quadrature component for gaussian shaped pulses
```
yomach, line 9: If you're at it, put the comments in the docstrings!
Actually, it would be great to add these to qualang_tools, and then just import it!!!!

But we can do it in 2 stages, first update this file to the github, then prepare the qualang_tools, then update this file again. Whatever you prefer.

yomach (approved):

yomach:
Don't close the previous PR and open a new one. This is very confusing.
Github allows reviewers to see changes from the previous reviews, and this is very important! I also want to be able to see my old comments...
1 Branch == 1 PR

KevinAVR (commented):
DRAG now in exponential form

PR #110, opened 2022-05-04
Use Case 1 - Saffman Lab - Atom Sorting

TheoLaudatQM:
lower case everywhere
Fix typos and add docstrings
results section
fix typo
Fix figure
Improve data occupation matrix transfer
Correct pulse sequence
Add intro according to Niv
Add docstring to config and remove saves


Quantum-Control-Applications/AMO/Use Case 1 - Saffman Lab - Atom Sorting/Array_sorting.py
```diff
@@ -0,0 +1,497 @@
...
+from qm import SimulationConfig
+from qm import LoopbackInterface
+from Config_array_sorting import *
+import matplotlib.pyplot as plt
+
+
+################################
+# Open quantum machine manager #
+################################
+qmm = QuantumMachinesManager()
```
yomach, line 12: add host= and ip= and add them to the config (same as SC config, check there)

Quantum-Control-Applications/AMO/Use Case 1 - Saffman Lab - Atom Sorting/Array_sorting.py
```diff
@@ -0,0 +1,497 @@
+from qm.qua import *
+from qm.QuantumMachinesManager import QuantumMachinesManager
+from qm import SimulationConfig
+from qm import LoopbackInterface
+from Config_array_sorting import *
```
yomach, line 5: change file names also to be lower case

Quantum-Control-Applications/AMO/Use Case 1 - Saffman Lab - Atom Sorting/Array_sorting.py
```diff
@@ -0,0 +1,497 @@
...
+target_frequencies_full_python = [[int(column_if[i]) if atom_target_list[j][i] else 0 for i in range(len(atom_target_list[0]))]
+                           for j in range(number_of_rows)]
+target_frequencies_1D = [j for sub in target_frequencies_full_python for j in sub]
+
+#############################################
+# Macros used to derive relevant parameters #
+#############################################
+
+
+def find_number_of_tweezers():
```
yomach, line 60: need to change Macros to match our "best practice guide"
https://qm-docs.qualang.io/guides/best_practices#macros

yomach:
See my comments, it's mostly done. I didn't go too deeply into it, just took a look a the basic code structure.
Maybe someone should take a closer look at the readme? I can try to do it tomorrow

yomach:
Oh, and you need to do the black formating as well

Quantum-Control-Applications/AMO/Use Case 1 - Saffman Lab - Atom Sorting/Array_sorting.py
TheoLaudatQM: Done

Quantum-Control-Applications/AMO/Use Case 1 - Saffman Lab - Atom Sorting/Array_sorting.py
TheoLaudatQM: Done

Quantum-Control-Applications/AMO/Use Case 1 - Saffman Lab - Atom Sorting/Array_sorting.py
TheoLaudatQM: I can do this of course, but I thought that the use-cases should correspond to the program that did run on a specific experiment without modifications as stated in Schluster use-case...

yomach: hmm, I understand, but I'm not sure we want to postcode that doesn't correspond to our own practices...
I think that the main thing is that use-cases are not meant to be "runnable".

It's not a functional change, not different than changing a variable name.

@tomerfqm WDYT?

Quantum-Control-Applications/AMO/Use Case 1 - Saffman Lab - Atom Sorting/Array_sorting.py
```diff
@@ -2,14 +2,14 @@
 from qm.QuantumMachinesManager import QuantumMachinesManager
 from qm import SimulationConfig
 from qm import LoopbackInterface
-from Config_array_sorting import *
+from config_array_sorting import *
```
yomach, line 5: you also need to rename the file(s)

Quantum-Control-Applications/AMO/Use Case 1 - Saffman Lab - Atom Sorting/Array_sorting.py
yomach: Oh, I see that you actually already made the change?

yomach:
Black formating still fails, I can help with that

Quantum-Control-Applications/AMO/Use Case 1 - Saffman Lab - Atom Sorting/Array_sorting.py
TheoLaudatQM: I understand but to me it is a bit more than just changing a variable name, because before the variables were global and could change from anywhere in the program. Now they are local and so the logic may have changed which will result in a different program... I tested it on Bordeaux lab and it seems to behave properly...
In my opinion there are two ways of presenting the use-cases: either (1) they are codes that ran on real experiment and so it is sure that it works but it is hardware specific and so adaptations need to be made if someone else wants to implement it and we cannot update them if our best practice changed for instance, or (2) they are general (not hardware specific) programs that do not correspond to any experiment and will need adaptations before being implemented, but they will follow our best practice and can be modified if needed.
If think we need to discuss it and choose the most relevant case while we do not have too many use-cases...

TheoLaudatQM: > Oh, I see that you actually already made the change?

Yes but I saved the two versions

Quantum-Control-Applications/AMO/Use Case 1 - Saffman Lab - Atom Sorting/Array_sorting.py
TheoLaudatQM: I did but for some reason github did not update them...

Quantum-Control-Applications/AMO/Use Case 1 - Saffman Lab - Atom Sorting/Array_sorting.py
yomach: Yes, you're right in principle about the macros. The thing is, if they were used as global (changed outside of the macro), then it is very very confusing...
I think that we are going for (1), and we definitely will not update them if we update our best practices. Maybe we can just add a note regarding it...
In this specific case, I would say that the main problem is that at Saffman they were written without the best practice :)

I'm also ok with keeping the macros as they were, and adding a note about it in the readme file, something like:
"Note that the macros are not written according to the best practice and are using global variables, use with care"
whatever you prefer.

TheoLaudatQM:
I think it was because of the config, hope it works now

Quantum-Control-Applications/AMO/Use Case 1 - Saffman Lab - Atom Sorting/Array_sorting.py
TheoLaudatQM: Done

Quantum-Control-Applications/AMO/Use Case 1 - Saffman Lab - Atom Sorting/Array_sorting.py
TheoLaudatQM: Ok I understand, I 100% prefer to upload the best practice one with the modified macro because it is indeed much clearer.
By the way, I think I can also add macros for get the locations and frequencies of each row and another one for reading the occupation matrix from the analog input to make the main program more concise

yomach: This is up to you :) 
From my POV it is ready.

I still haven't looked at the readme too deeply, but we can release and fix it afterward if any changes are needed...

yomach (approved):

yomach:
Everything passed, you can merge it when you're ready

PR #117, opened 2022-05-19
changing names of operations in ALLXY

MichalGQM:


Quantum-Control-Applications/Superconducting/Single Fixed Transmon/allxy.py
```diff
@@ -332,26 +333,26 @@
...
+    ("y180", "x90"),
+    ("x90", "x180"),
+    ("x180", "x90"),
+    ("y90", "y180"),
+    ("y", "y90"),
+    ("x180", "I"),
+    ("y180", "I"),
+    ("x90", "x90"),
+    ("y90", "y90"),
 ]
```
yomach, line 356: Hmm, I'm not sure if we want to change the table as well or not... convince me?

MichalGQM:
We actually want them to get used to the new names .. So why would we leave
here the old ones ... These names will be shown in the x axes in the graph
right ? So we would want them to see the new names to get used to it.
And - just to make it all in the same names ... makes it more clear.. don't
you think so?

yomach (approved):

PR #119, opened 2022-05-22
Exmples in timing

MichalGQM:


yomach:
There are many small issues:
1. `qe` everywhere
2. `QMm` instead of `qmm`
3. If you run it, the plots will mess up with one another
4. Issues in the readme file, the notes are not correctly written.

Maybe build the site locally and take a look? and also run the file and adjust it.
We can do a session to go over stuff.

MichalGQM:
Hey,
Regarding 1, you mean qe instead of qe1?
Regarding 4 , I built it locally... I don't understand what does not
correctly mean ... .

Thanks!

yomach (changes requested):
1.

Tutorials/intro-to-timing/timing_in_qua.py
```diff
@@ -35,44 +35,26 @@
...
-
-plt.subplots_adjust(left=0.1, bottom=0.1, right=0.9, top=0.9, wspace=0.4, hspace=0.4)
-
+samples.con1.plot()
+plt.legend(["qubit", "resonator"])
 
 # Simulate and plot 2nd example
 job = qmm.simulate(config, example2, SimulationConfig(int(1000)))  # in clock cycles, 4 ns
 samples = job.get_simulated_samples()
+plt.title("example 1 - Two pulses from different elements")
```
yomach, line 44: This is in the wrong place and is not being shown, is that on purpose?

MichalGQM: Indeed.. it was in the wrong place but it is not being shown on purpose since the title is actually written in the sentence above each figure...

yomach: Cool, then move all of the titles to the correct place and then we can merge

PR #122, opened 2022-05-31
SC & NV improvements

yomach:


Quantum-Control-Applications/Superconducting/Single Fixed Transmon/time_of_flight.py
```diff
@@ -1,3 +1,7 @@
+"""
+A script used to look at the raw ADC data, this allows checking that the ADC is not saturated, correct for DC offsets
```
TheoLaudatQM, line 2: Shall we rename it raw_adc_traces.py for consistency?

Quantum-Control-Applications/Superconducting/Single Fixed Transmon/configuration.py
```diff
@@ -64,7 +64,6 @@ def readout_macro(threshold=None, state=None, I=None, Q=None):
 #############
```
TheoLaudatQM, line 64: For the future when we'll have more of them, I think it is better to put the QUA macros in a separated file called macros.py rather than in the configuration

Quantum-Control-Applications/NV-Centers/Confocal-Setup/manual_mixer_calibration.py
```diff
@@ -0,0 +1,31 @@
...
+qm = qmm.open_qm(config)
+
+job = qm.execute(cw_output)
+
+# When done, the halt command can be called and the offsets can be written directly into the config file.
+
+# job.halt
+
+# These are the 2 commands used to correct for mixer imperfections. The first is used to set the DC of the `I` and `Q`
+# channels to compensate for the LO leakage. The 2nd command is used to correct for the phase and amplitude mismatches
```
TheoLaudatQM, line 22: I would add something like this since several customers seemed confused: 
"Since this compensation depends on the I & Q powers, it is advised to run this step with no input power, so that there is no LO leakage while the pulses are not played."

Quantum-Control-Applications/Superconducting/Single Fixed Transmon/manual_mixer_calibration.py
```diff
@@ -0,0 +1,31 @@
...
+
+job = qm.execute(cw_output)
+
+# When done, the halt command can be called and the offsets can be written directly into the config file.
+
+# job.halt
+
+# These are the 2 commands used to correct for mixer imperfections. The first is used to set the DC of the `I` and `Q`
+# channels to compensate for the LO leakage. The 2nd command is used to correct for the phase and amplitude mismatches
+# between the channels.
```
TheoLaudatQM, line 23: I would add something like this since several customers seemed confused: 
"Since this compensation depends on the I & Q powers, it is advised to run this step with no input power, so that there is no LO leakage while the pulses are not played."

Quantum-Control-Applications/Superconducting/Single Fixed Transmon/configuration.py
yomach: Yea I agree.
But not for this PR

Quantum-Control-Applications/NV-Centers/Confocal-Setup/manual_mixer_calibration.py
yomach: good point

Quantum-Control-Applications/NV-Centers/Confocal-Setup/configuration.py
```diff
@@ -75,10 +74,10 @@ def IQ_imbalance(g, phi):
     },
     "elements": {
         "NV": {
-            "mixInputs": {"I": ("con1", 1), "Q": ("con1", 2), "lo_frequency": NV_LO_freq, "mixer": "mixer_qubit"},
+            "mixInputs": {"I": ("con1", 1), "Q": ("con1", 2), "lo_frequency": NV_LO_freq, "mixer": "mixer_NV"},
             "intermediate_frequency": NV_IF_freq,
             "operations": {
-                "const": "cw_pulse",
+                "cw": "const_pulse",
```
TheoLaudatQM, line 80: Just to understand and also modify the flux tunable transmon files if needed, why do you want to use "cw" instead of "const"?

yomach: Yes, I wanted to make it consistent, it was different between the NV and the SC.
I randomly decided that the operation will be "cw".

PR #127, opened 2022-06-12
Codes for ESR experimental bring up

KevinAVR:
fine tuned from the installation in Darmouth


Quantum-Control-Applications/ESR/T1.py
```diff
@@ -0,0 +1,165 @@
...
+# Open quantum machine #
+########################
+
+qm = qmm.open_qm(config)
+
+###################
+# The QUA program #
+###################
+
+pi_len = int(320//4)
```
yomach, line 25: no need for `int`. The `//` already keeps it as an int

Quantum-Control-Applications/ESR/T1.py
```diff
@@ -0,0 +1,165 @@
...
+pi_half_len = int(160//4)
+
+wait_min = int(1500//4)
+wait_max = int(20000//4)
+dwait = 1000//4
+
+waits = np.arange(wait_min, wait_max + dwait/2, dwait)
```
yomach, line 32: it's better to do it `+ 0.1` instead of `+ dwait/2`, otherwise it can give different results from QUA.
Not very critical as today we'll probably import the loop tools and then we can just use them.

Quantum-Control-Applications/ESR/T1.py
```diff
@@ -0,0 +1,165 @@
...
+    Q = declare(fixed)
+    I_st = declare_stream()
+    Q_st = declare_stream()
+
+    with for_(n, 0, n < n_avg, n + 1):
+
+        with for_(wait_len, wait_min, wait_len <= wait_max, wait_len + dwait):
+
+            # initialization
+            # wait(wait_len)
```
yomach, line 55: remove commented out code

Quantum-Control-Applications/ESR/T1.py
```diff
@@ -0,0 +1,165 @@
...
+
+            reset_phase('ensemble')
+            reset_phase('resonator')
+            reset_frame('ensemble')
+
```
yomach, line 60: why do you need to reset phase and frame?

Quantum-Control-Applications/ESR/T1.py
```diff
@@ -0,0 +1,165 @@
...
+            echo = declare_stream(adc_trace=True)
+
+            assign(pulse_delay, 2000//4 - Cast.mul_int_by_fixed(pi_half_len, 0.5) - Cast.mul_int_by_fixed(pi_len, 0.5))
+            assign(readout_delay, 2000//4 - Cast.mul_int_by_fixed(pi_len, 0.5) - Cast.mul_int_by_fixed(readout_len, 0.125) - 5)
+
```
yomach, line 65: Everything here is constant and python... So why?

1. Cast.mul_int_by_fixed ?
2. Why not just calculate this in advance? this doesn't change between iterations...

Quantum-Control-Applications/ESR/T2.py
```diff
@@ -0,0 +1,165 @@
...
+
+            reset_phase('ensemble')
+            reset_phase('resonator')
+            reset_frame('ensemble')
+
+            echo = declare_stream(adc_trace=True)
+
+            assign(pulse_delay, delay_len - Cast.mul_int_by_fixed(pi_half_len, 0.5) - Cast.mul_int_by_fixed(pi_len, 0.5))
+            assign(readout_delay, delay_len - Cast.mul_int_by_fixed(pi_len, 0.5) - Cast.mul_int_by_fixed(readout_len, 0.125) - 5)
+
```
yomach, line 65: here delay_len changes.
But still, no need for Cast.mul_int_by_fixed

Quantum-Control-Applications/ESR/T1.py
```diff
@@ -0,0 +1,165 @@
...
+            wait(8, 'switch_1', 'switch_2')
```
yomach, line 66: why wait(8)?

Quantum-Control-Applications/ESR/configuration.py
```diff
@@ -0,0 +1,537 @@
...
+    'version': 1,
+
+    'controllers': {
+
+        'con1': {
+
+            'type': 'opx1',
+
+            'analog_outputs': {
+                1: {'offset': +0.0},  # qubit I
```
yomach, line 103: the comments here are wrong, either update or remove

Quantum-Control-Applications/ESR/configuration.py
```diff
@@ -0,0 +1,537 @@
...
+                'x180': 'x180_pulse',
+                'x90': 'x90_pulse',
+                '-x90': '-x90_pulse',
+                'y180': 'y180_pulse',
+                'y90': 'y90_pulse',
+                '-y90': '-y90_pulse',
+            },
+        },
+
+        'ensemble2': {
```
yomach, line 152: what is ensemble2?

Quantum-Control-Applications/ESR/T1.py
```diff
@@ -0,0 +1,165 @@
...
+# Open quantum machine manager #
+################################
+
+qmm = QuantumMachinesManager(host=qop_ip, port=qop_port)
+
+########################
+# Open quantum machine #
+########################
+
+qm = qmm.open_qm(config)
```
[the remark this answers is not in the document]
KevinAVR: Thanks

Quantum-Control-Applications/ESR/T1.py
KevinAVR: True. I have also noticed that when I do type(1e6//4) it returns me a float which is annoying when I do update frequency but I will save it for thoses cases

KevinAVR: For example, in the same code I have `cooldown_time = 10e6//4` and it is of type float but `wait()` requires type int

Quantum-Control-Applications/ESR/T1.py
KevinAVR: These comments are place holders where you are going to put the type of inizialization (in the insatllation was for hyperpolarization with a laser) and part of T1 experiments is using wait()

Quantum-Control-Applications/ESR/T1.py
KevinAVR: good point!

Quantum-Control-Applications/ESR/T2.py
KevinAVR: Is it no need because it is done in python? so then it know to do well (int * float)?

KevinAVR: Nice! I checked with the simulation and the calculations comes out right thanks!

Quantum-Control-Applications/ESR/T1.py
KevinAVR: because `play()` uses `duration` and is analog it takes 8 cycles to come out later in comparison to the `play` of a digital pulses that uses `duration`

Quantum-Control-Applications/ESR/configuration.py
KevinAVR: cool

Quantum-Control-Applications/ESR/configuration.py
KevinAVR: it was specific to the experiment but now I see that is it not needed and it overcrowds only

Quantum-Control-Applications/ESR/configuration.py
```diff
@@ -0,0 +1,537 @@
...
+            'mixInputs': {
+                'I': ('con1', 3),
+                'Q': ('con1', 4),
+                'lo_frequency': ensemble_LO,
+                'mixer': 'mixer_ensemble',
```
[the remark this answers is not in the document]
KevinAVR: deleted ensemble2

Quantum-Control-Applications/ESR/cpmg.py
```diff
@@ -0,0 +1,206 @@
...
+
+qm = qmm.open_qm(config)
+
+###################
+# The QUA program #
+###################
+
+tau_min = 2000//4  # tau is half the time between pi pulses
+tau_max = 5e4//4  # tau is half the time between pi pulses
+tau_array = np.round(np.exp(np.linspace(np.log(tau_min), np.log(tau_max), 30)))  # outputs results in float
```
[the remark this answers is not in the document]
KevinAVR: I played with it and keep running into the same issue that I need to add np.log() and change the base to np.exp(1). np.logspace starts at (base**tau_min) not at tau_min, if it were to start at tau_min and end at tau_max then do logspacing it would be fantastic

Quantum-Control-Applications/ESR/cpmg.py
```diff
@@ -0,0 +1,206 @@
...
+
+tau_array_int = [i for i in range(len(tau_array))]
+
+for i in range(len(tau_array)):
+    tau_array_int[i] = int(tau_array[i])
```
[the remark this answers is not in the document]
KevinAVR: cool! it works

Quantum-Control-Applications/ESR/cpmg.py
```diff
@@ -0,0 +1,206 @@
...
+
+num_taus = len(tau_array)
+
+latestecho = 50e6//4  # this is the maximum amount of time I will run CPMG
+
+numecho_array = [i for i in range(num_taus)]
+
+for i_tau in range(0, num_taus):
+    numecho_array[i_tau] = int(np.floor(latestecho/(2*tau_array[i_tau])))  # already in clock cycles
```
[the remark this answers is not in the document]
KevinAVR: did modification as above, np.array(tau_array) --- tau_array is alread np.array

Quantum-Control-Applications/ESR/cpmg.py
```diff
@@ -0,0 +1,206 @@
...
+tau_array = np.round(np.exp(np.linspace(np.log(tau_min), np.log(tau_max), 30)))  # outputs results in float
+
+tau_array_int = [i for i in range(len(tau_array))]
+
+for i in range(len(tau_array)):
+    tau_array_int[i] = int(tau_array[i])
+
+num_taus = len(tau_array)
+
+latestecho = 50e6//4  # this is the maximum amount of time I will run CPMG
```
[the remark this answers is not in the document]
KevinAVR: good sugg

Quantum-Control-Applications/ESR/cpmg.py
```diff
@@ -0,0 +1,206 @@
...
+            # initialization
+            # wait(cooldown_time)
+
+            reset_phase('ensemble')
+            reset_phase('resonator')
+            reset_frame('ensemble')
+
+            assign(tau, tau_array_q[i_tau])
+
+            wait(11, 'switch_1', 'switch_2')
```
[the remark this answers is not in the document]
KevinAVR: duration in play() of digital pulse comes earlier than in the case of analog pulse, so I have to make it wait

Quantum-Control-Applications/ESR/cpmg.py
```diff
@@ -0,0 +1,206 @@
...
+            play('activate', 'switch_1', duration=pulse1_len)
+            play('activate', 'switch_2', duration=pulse1_len)
+            play('const', 'ensemble', duration=pulse1_len)
+
+            frame_rotation_2pi(-0.5, 'ensemble')
+
+            wait(tau-Cast.mul_int_by_fixed(pulse1_len, 1.5)-27, 'ensemble')
```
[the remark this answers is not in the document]
KevinAVR: so I am doing from ecneter of pi_half pulse to the center of pi_pulse. Here the difference between the two are related to the duration, so pi_half is half the duration of pi_half. Then I need reduce the tau by 1.5*pulse1_len. Then the over head of calcualtions in real time requires me to add 27, I found this number by doing assetions in the simulation part

Quantum-Control-Applications/ESR/cpmg.py
```diff
@@ -0,0 +1,206 @@
...
+
+                play('activate_resonator', 'switch_receiver')
+                measure('readout', 'resonator', echo, dual_demod.full('cos', 'out1', 'sin', 'out2', I),
+                        dual_demod.full('minus_sin', 'out1', 'cos', 'out2', Q))
+                save(I, I_st)
+                save(Q, Q_st)
+                save(tau, tau_st)
+                save(i_echo, i_echo_st)
+
+                wait(tau - pulse1_len - Cast.mul_int_by_fixed(readout_len, 0.125) - 66, 'switch_receiver', 'resonator')
```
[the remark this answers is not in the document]
KevinAVR: same as above `66` due to real time calculationand the foor loop time it takes to enter, these numbers I verified with asseriton to analog wavefmors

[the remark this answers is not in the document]
KevinAVR: Note that all of these is so that the distance between pulse from cneter to center is tau the same for readout so that the echo is centered

Quantum-Control-Applications/ESR/T1.py
yomach: Then write it in a comment.
Again, this is very weird for someone who doesn't know it...

Quantum-Control-Applications/ESR/cpmg.py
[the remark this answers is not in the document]
yomach: Cool, then just add a comment about it :)
Maybe we can add qrun soon to make it clearer

Quantum-Control-Applications/ESR/cpmg.py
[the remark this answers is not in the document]
yomach: These numbers have to be explained

Quantum-Control-Applications/ESR/cpmg.py
```diff
@@ -0,0 +1,206 @@
...
+            align()
+
+            with for_(i_echo, 0, i_echo < numecho_array_q[i_tau], i_echo + 1):
+                play('activate', 'switch_1', duration=pulse2_len)
+                play('activate', 'switch_2', duration=pulse2_len)
+                play('const', 'ensemble', duration=pulse2_len)
+
+                align()  # global align
+
+                wait(tau-pulse1_len - Cast.mul_int_by_fixed(readout_len, 0.125) - 5, 'switch_receiver', 'resonator')
```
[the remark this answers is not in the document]
yomach: Same thing

Quantum-Control-Applications/ESR/cpmg.py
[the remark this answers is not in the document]
yomach: 👍

Quantum-Control-Applications/ESR/T1.py
TheoLaudatQM: Yes because 10e6 is a float, what we did in the other codes is to set qubit_T1 = int(10e6) and then declare cooldown_time = qubit_T1 // 4

Quantum-Control-Applications/ESR/README.md
```diff
@@ -0,0 +1,18 @@
+# ESR experiments on an ensemble
+
+## Basic Files
+These files showcase various ESR experiments that can be done on an ensemble.
```
TheoLaudatQM, line 4: I would specify what ESR means (Electron Spin Resonance), I didn't know the acronym...
And also specify the setup, is it for SC (fixed or tunable freq), QD, NV?

Quantum-Control-Applications/ESR/T1.py
```diff
@@ -0,0 +1,172 @@
...
+dwait = 1000 // 4
+
+waits = np.arange(wait_min, wait_max + 0.1, dwait)
+
+cooldown_time = int(10e6 // 4)
+
+n_avg = 100
+
+pulse_delay_p = int(2000 // 4 - pi_half_len * 0.5 - pi_len * 0.5)
+readout_delay_p = int(2000 // 4 - pi_len * 0.5 - readout_len * 0.125 - 5)
```
TheoLaudatQM, line 28: I would explain in comments why we have to wait these specific values, especially why 2000ns? and why readout_len*0.125-5?

Quantum-Control-Applications/ESR/T1.py
```diff
@@ -0,0 +1,172 @@
...
+            # initialization
+            # wait(wait_len)
+
+            reset_phase("ensemble")
+            reset_phase("resonator")
+            reset_frame("ensemble")
+
+            echo = declare_stream(adc_trace=True)
+
+            wait(8, "switch_1", "switch_2")
```
TheoLaudatQM, line 55: Why wait 8 cc here ? is it set-up specific or due to OPX internal processing?

Quantum-Control-Applications/ESR/T1.py
```diff
@@ -0,0 +1,172 @@
...
+        duration=2000,
+        include_analog_waveforms=True,
+        simulation_interface=LoopbackInterface(([("con1", 3, "con1", 1), ("con1", 4, "con1", 2)]), latency=180),
+    )
+    job = qmm.simulate(config, T1, simulate_config)  # do simulation with qmm
+    job.get_simulated_samples().con1.plot()  # visualize played pulses
+
+    analog_wf = job.simulated_analog_waveforms()
+
+    ver_t1 = (
```
TheoLaudatQM, line 113: I think it would be useful to add a comment here explaining what this is for

Quantum-Control-Applications/ESR/T1.py
```diff
@@ -0,0 +1,172 @@
...
+            pass
+
+    plt.cla()
+    I = I_handle.fetch_all()
+    Q = Q_handle.fetch_all()
+    iteration = iteration_handle.fetch_all()
+    print(f"{round(iteration/n_avg * 100)}%")
+
+    plt.plot(waits * 4, I)
+    plt.plot(waits * 4, Q)
```
TheoLaudatQM, line 172: To make it nicer, I would directly put xlabel, ylabel, title, tight_layout and so on...

Quantum-Control-Applications/ESR/T1.py
```diff
@@ -0,0 +1,172 @@
+from qm import SimulationConfig
```
TheoLaudatQM, line 1: I would add a docstring at the beginning of each file explaining what it does. I know it is on the readme but when you downloaded the files and start working with it, it is more convenient to have it here as well :)

Quantum-Control-Applications/ESR/T1.py
```diff
@@ -0,0 +1,172 @@
...
+            play("activate", "switch_2", duration=pi_half_len)
+            play("const", "ensemble", duration=pi_half_len)
+
+            wait(pulse_delay, "ensemble", "switch_1", "switch_2")
+
+            frame_rotation_2pi(-0.5, "ensemble")
+            play("activate", "switch_1", duration=pi_len)
+            play("activate", "switch_2", duration=pi_len)
+            play("const", "ensemble", duration=pi_len)
+
```
TheoLaudatQM, line 66: I am not sure I understand the sequence, you play first pi/2 and then after waiting some fixed time you play pi, then wait again and measure. Why this sequence of pi/2 and pi and why don't you use wait_len?

Quantum-Control-Applications/ESR/T1.py
```diff
@@ -0,0 +1,172 @@
...
+        except Exception as e:
+            pass
+
+    plt.cla()
+    I = I_handle.fetch_all()
+    Q = Q_handle.fetch_all()
+    iteration = iteration_handle.fetch_all()
+    print(f"{round(iteration/n_avg * 100)}%")
+
```
TheoLaudatQM, line 170: Since you're also acquiring the raw traces, I would plot them to save some time next time we'll need it and comment the lines if we don't want these plots

Quantum-Control-Applications/ESR/T1.py
yomach: We need to fix `wait()` to accept a float :\ 
(duration accepts it)

Quantum-Control-Applications/ESR/T2.py
```diff
@@ -0,0 +1,172 @@
...
+            play("activate", "switch_1", duration=pi_half_len)
+            play("activate", "switch_2", duration=pi_half_len)
+            play("const", "ensemble", duration=pi_half_len)
+
+            wait(pulse_delay, "ensemble", "switch_1", "switch_2")
+
+            frame_rotation_2pi(-0.5, "ensemble")
+            play("activate", "switch_1", duration=pi_len)
+            play("activate", "switch_2", duration=pi_len)
+            play("const", "ensemble", duration=pi_len)
```
TheoLaudatQM, line 64: Same comment as for T1, why do we use this pi/2 - wait - pi - wait sequence? isn't T2 measured with a Ramsey sequence?

Quantum-Control-Applications/ESR/T2.py
```diff
@@ -0,0 +1,172 @@
...
+
+    ver_t1 = (
+        analog_wf["elements"]["ensemble"][2]["timestamp"] + (analog_wf["elements"]["ensemble"][2]["duration"] / 2)
+    ) - (analog_wf["elements"]["ensemble"][0]["timestamp"] + (analog_wf["elements"]["ensemble"][0]["duration"] / 2))
+    ver_t3 = (
+        analog_wf["elements"]["resonator"][0]["timestamp"] + (analog_wf["elements"]["resonator"][0]["duration"] / 2)
+    ) - (analog_wf["elements"]["ensemble"][2]["timestamp"] + (analog_wf["elements"]["ensemble"][2]["duration"] / 2))
+
+    print("center to center time between 1st and 2nd pulse", ver_t1)
+    print("center to center time between readout and 2nd pulse", ver_t3)
```
TheoLaudatQM, line 121: Is this done before executing on hardware to get the readout and pulse delays?, If so I would write in a comment

Quantum-Control-Applications/ESR/cpmg.py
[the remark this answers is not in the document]
TheoLaudatQM: You could write like this: `np.round(np.logspace(np.log10(tau_min), np.log10(tau_max),30))` which gives you the exact same results

Quantum-Control-Applications/ESR/cpmg.py
```diff
@@ -0,0 +1,213 @@
...
+    tau = declare(int)
+    i_tau = declare(int)
+
+    num_echo_array_q = declare(int, value=num_echo_array)
+    i_echo = declare(int)
+
+    echo = declare_stream(adc_trace=True)
+
+    pulse1_len = declare(int, value=240 // 4)  # 250 cycles = 1 us
+    pulse2_len = declare(int, value=480 // 4)
```
TheoLaudatQM, line 53: Where do these values come from?

Quantum-Control-Applications/ESR/cpmg.py
```diff
@@ -0,0 +1,213 @@
...
+
+    tau_st = declare_stream()
+    i_echo_st = declare_stream()
+
+    tau_array_q = declare(int, value=tau_array_int)
+    tau = declare(int)
+    i_tau = declare(int)
+
+    num_echo_array_q = declare(int, value=num_echo_array)
+    i_echo = declare(int)
```
TheoLaudatQM, line 48: Is this variable the same as 'iteration' that you used in the SC codes ? If so, I would keep iteration which seems clearer

Quantum-Control-Applications/ESR/cpmg.py
```diff
@@ -0,0 +1,213 @@
...
+            # initialization
+            # wait(cooldown_time)
+
+            reset_phase("ensemble")
+            reset_phase("resonator")
+            reset_frame("ensemble")
+
+            assign(tau, tau_array_q[i_tau])
+
+            wait(11, "switch_1", "switch_2")
```
TheoLaudatQM, line 68: Where does this value come from?

Quantum-Control-Applications/ESR/cpmg.py
```diff
@@ -0,0 +1,213 @@
...
+            play("activate", "switch_1", duration=pulse1_len)
+            play("activate", "switch_2", duration=pulse1_len)
+            play("const", "ensemble", duration=pulse1_len)
+
+            frame_rotation_2pi(-0.5, "ensemble")
+
+            wait(tau - Cast.mul_int_by_fixed(pulse1_len, 1.5) - 27, "ensemble")
```
TheoLaudatQM, line 75: Where do these values come from?

Quantum-Control-Applications/ESR/cpmg.py
```diff
@@ -0,0 +1,213 @@
...
+            align()
+
+            with for_(i_echo, 0, i_echo < num_echo_array_q[i_tau], i_echo + 1):
+                play("activate", "switch_1", duration=pulse2_len)
+                play("activate", "switch_2", duration=pulse2_len)
+                play("const", "ensemble", duration=pulse2_len)
+
+                align()  # global align
+
+                wait(tau - pulse1_len - Cast.mul_int_by_fixed(readout_len, 0.125) - 5, "switch_receiver", "resonator")
```
TheoLaudatQM, line 86: Where do these values come from ?

Quantum-Control-Applications/ESR/cpmg.py
```diff
@@ -0,0 +1,213 @@
...
+            play("activate", "switch_2", duration=pulse1_len)
+            play("const", "ensemble", duration=pulse1_len)
+
+            frame_rotation_2pi(-0.5, "ensemble")
+
+            wait(tau - Cast.mul_int_by_fixed(pulse1_len, 1.5) - 27, "ensemble")
+
+            align()
+
+            with for_(i_echo, 0, i_echo < num_echo_array_q[i_tau], i_echo + 1):
```
TheoLaudatQM, line 79: I thought CPMG was wait (tau/2)-Y-wait(tau)-Y-wait(tau/2) is it what you're doing here?

Quantum-Control-Applications/ESR/cpmg.py
```diff
@@ -0,0 +1,213 @@
...
+    i_echo = declare(int)
+
+    echo = declare_stream(adc_trace=True)
+
+    pulse1_len = declare(int, value=240 // 4)  # 250 cycles = 1 us
+    pulse2_len = declare(int, value=480 // 4)
+
+    with for_(n, 0, n < n_avg, n + 1):
+
+        with for_(i_tau, 0, i_tau < num_taus, i_tau + 1):
```
TheoLaudatQM, line 57: Here tau does not take the values in tau_array, is it normal?

Quantum-Control-Applications/ESR/input_calibration.py
```diff
@@ -0,0 +1,75 @@
+from qm import SimulationConfig
```
TheoLaudatQM, line 1: I think this file is the same as 'raw_adc_traces.py' in the SC folder maybe we can rename it for consistency

Quantum-Control-Applications/ESR/input_calibration.py
```diff
@@ -0,0 +1,75 @@
...
+with program() as input_cal:
+
+    n = declare(int)
+
+    adc_st = declare_stream(adc_trace=True)
+
+    with for_(n, 0, n < 5000, n + 1):
+        play("activate", "switch_receiver", duration=500)
+        wait(150, "resonator")
+        reset_phase("resonator")
```
TheoLaudatQM, line 22: I assume that you want to measure the offsets after playing 'activate' on 'switch_receiver' because that's how you will measure data in the real sequences, right ?
Does the wait value matter here? because it seems to be different from what you used in the other sequences

Quantum-Control-Applications/ESR/input_calibration.py
```diff
@@ -0,0 +1,75 @@
...
+
+    plt.figure()
+    plt.title("Single run (Check ADCs saturation)")
+    plt.plot(adc1_single_run / 2**12)
+    plt.plot(adc2_single_run / 2**12)
+
+    plt.figure()
+    plt.title("Averaged run")
+    plt.plot(adc1 / 2**12)
+    plt.plot(adc2 / 2**12)
```
TheoLaudatQM, line 75: Printing the mean values as in the SC codes is actually quite useful to quickly extract the mean values

Quantum-Control-Applications/ESR/mixer_cal.py
```diff
@@ -0,0 +1,43 @@
...
+        duration=2000,
+        simulation_interface=LoopbackInterface(([("con1", 3, "con1", 1), ("con1", 4, "con1", 2)]), latency=180),
+    )
+    job = qmm.simulate(config, mixer_cal, simulate_config)  # do simulation with qmm
+    job.get_simulated_samples().con1.plot()  # visualize played pulses
+
+else:
+    qm = qmm.open_qm(config)
+
+    job = qm.execute(mixer_cal)  # execute QUA program
```
TheoLaudatQM, line 43: I would add this after the job to save some time:
# When done, the halt command can be called and the offsets can be written directly into the config file.

# job.halt

# These are the 2 commands used to correct for mixer imperfections. The first is used to set the DC of the `I` and `Q`
# channels to compensate for the LO leakage. Since this compensation depends on the I & Q powers, it is advised to
# run this step with no input power, so that there is no LO leakage while the pulses are not played.
# The 2nd command is used to correct for the phase and amplitude mismatches between the channels.
# The output of the IQ Mixer should be connected to a spectrum analyzer and values should be chosen as to minimize the
# unwanted peaks.
# If python can read the output of the spectrum analyzer, then this process can be automated and the correct values can
# found using an optimization method such as Nelder-Mead:
# https://docs.scipy.org/doc/scipy/reference/optimize.minimize-neldermead.html

# qm.set_output_dc_offset_by_element('ensemble', ('I', 'Q'), (-0.001, 0.003))
# qm.set_mixer_correction('mixer_ensemble', int(ensemble_IF), int(ensemble_LO), IQ_imbalance(0.015, 0.01))

Quantum-Control-Applications/ESR/pi_pulse_calibration.py
```diff
@@ -0,0 +1,175 @@
...
+    I = declare(fixed)
+    Q = declare(fixed)
+    I_st = declare_stream()
+    Q_st = declare_stream()
+
+    with for_(n, 0, n < n_avg, n + 1):
+
+        with for_(pulse1_len, pulse1_min, pulse1_len <= pulse1_max, pulse1_len + dpulse1):
+
+            # wait(cooldown_time)
```
TheoLaudatQM, line 40: Why is this commented?

Quantum-Control-Applications/ESR/pi_pulse_calibration.py
```diff
@@ -0,0 +1,175 @@
...
+            reset_phase("ensemble")
+            reset_phase("resonator")
+            reset_frame("ensemble")
+
+            echo = declare_stream(adc_trace=True)
+
+            assign(pulse2_len, 2 * pulse1_len)
+            assign(
+                pulse_delay,
+                2000 // 4 - Cast.mul_int_by_fixed(pulse1_len, 0.5) - Cast.mul_int_by_fixed(pulse2_len, 0.5) - 8,
```
TheoLaudatQM, line 51: this `Cast.mul_int_by_fixed(pulse2_len, 0.5)` is actually pulse1_len no?
To get these values, you first rune the simulator right?

Quantum-Control-Applications/ESR/signal_test.py
```diff
@@ -0,0 +1,147 @@
...
+
+        # wait(cooldown_time)
+
+        reset_phase("ensemble")
+        reset_phase("resonator")
+        reset_frame("ensemble")
+
+        echo = declare_stream(adc_trace=True)
+
+        wait(8, "switch_1", "switch_2")
```
TheoLaudatQM, line 37: Why do you sometimes wait 8 cc or 11 cc ?

TheoLaudatQM (changes requested):
I made a lot of comments, sorry for this but these scripts are made to be used also by us and since I don't know the set-up and physics involved I had a lot of trouble understanding the sequences and I would not be able to use them in an installation as it is, there is no such things as to many comments :)

Quantum-Control-Applications/ESR/T1.py
KevinAVR: I like your point of view - done

Quantum-Control-Applications/ESR/T1.py
KevinAVR: good point

Quantum-Control-Applications/ESR/cpmg.py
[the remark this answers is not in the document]
KevinAVR: thanks

Quantum-Control-Applications/ESR/cpmg.py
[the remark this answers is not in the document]
KevinAVR: done

Quantum-Control-Applications/ESR/T2.py
KevinAVR: In a Ramsey sequence in SC qubits you need to apply a return pi/2 to go back to the z-axis to perform a measuremnet. In ESR you perform your measurment in XY plane. In ESR there are three T2, T2* (related to pi/2 and measure), T2_echo (related to pi/2 - variable delay - pi -echo measurement0, and T2_DD (rleate dto T2 measured by XY8 or CPMG)

Quantum-Control-Applications/ESR/cpmg.py
KevinAVR: arbitrary values but I put comments

Quantum-Control-Applications/ESR/cpmg.py
KevinAVR: it is not iteration, I have `n` for iteration. In this case i_echo is the variable that I loop over to create a CPMG that has a variable number of pi pulses in the sequence. We called it i_echo because is related to the number of echos you will collect in a single run

Quantum-Control-Applications/ESR/mixer_cal.py
KevinAVR: useful

Quantum-Control-Applications/ESR/configuration.py
```diff
@@ -0,0 +1,439 @@
...
+
+#############
+# VARIABLES #
+#############
+
+qop_ip = "127.0.0.1"
+qop_port = 80
+
+# Frequencies
+resonator_if = -30e6  # in Hz
```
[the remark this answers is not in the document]
KevinAVR: thanks

Quantum-Control-Applications/ESR/signal_test.py
KevinAVR: added docstrings

Quantum-Control-Applications/ESR/pi_pulse_calibration.py
KevinAVR: now I have put the laser to initizliaze

Quantum-Control-Applications/ESR/input_calibration.py
KevinAVR: ty

Quantum-Control-Applications/ESR/T2.py
KevinAVR: hes

Quantum-Control-Applications/ESR/T1.py
KevinAVR: cool

Quantum-Control-Applications/ESR/input_calibration.py
KevinAVR: does not matter

Quantum-Control-Applications/ESR/input_calibration.py
KevinAVR: the problem is that it does not measure anything meaningful just the offsets

Quantum-Control-Applications/ESR/cpmg.py
KevinAVR: so we want to cover 5 ms of measurement and we chose the values to do the experiment on so that they are equal space in log scale spannign these 5 ms that is why

Quantum-Control-Applications/ESR/cpmg.py
KevinAVR: I put docstrings

Quantum-Control-Applications/ESR/T1.py
```diff
@@ -105,20 +120,27 @@
         include_analog_waveforms=True,
         simulation_interface=LoopbackInterface(([("con1", 3, "con1", 1), ("con1", 4, "con1", 2)]), latency=180),
     )
+    # the simulation is uses to assert the pulse positions and to make final adjustments
```
TheoLaudatQM, line 123: the simulation is use**d**

Quantum-Control-Applications/ESR/cpmg.py
TheoLaudatQM: Ok I see, this will be solved with the loops tools :)

Quantum-Control-Applications/ESR/cpmg.py
TheoLaudatQM: The 1.5 and -27 but now I get it, you calibrate them using the simulator, thanks!

Quantum-Control-Applications/ESR/T1.py
TheoLaudatQM: Ok, thanks I'll definitely need to read about these techniques before the next NV installation...
However it seems that the duration in the wait statements is not iterating, is it normal? How do you get T1 if the time is not scanned?

Quantum-Control-Applications/ESR/cpmg.py
TheoLaudatQM: Yes correct, my bad...

Quantum-Control-Applications/ESR/cpmg.py
TheoLaudatQM: Cool thanks !

Quantum-Control-Applications/ESR/T1.py
TheoLaudatQM: Oh yes got it thanks !

TheoLaudatQM (approved):

PR #135, opened 2022-06-26
small fixes to the tutorials

tomerfqm:


Tutorials/frame-and-phase-intro/reset_phase_demo.py
```diff
@@ -15,7 +14,7 @@
 from configuration import *
 from scipy.optimize import leastsq
 
-qmm = QuantumMachinesManager(port=9510)
+qmm = QuantumMachinesManager()
```
yomach, line 17: I only wonder about this...
WDYT?

yomach: Not about removing the port (this is needed!), but about not having IP...

tomerfqm: What do you want to write there? `qop_ip`

tomerfqm: I removed it because it was a local thing on my end and I wanted it to be general..

TheoLaudatQM: I am a bit confused, so now we don't need to specify the IP at all ?

yomach: We need to specify an IP. I'm just saying that we should have a default value inside the `QuantumMachinesManager()` such that the syntax will be correct

TheoLaudatQM (approved):

PR #139, opened 2022-07-11
Small fixes to SC qubit

yomach:


Quantum-Control-Applications/Superconducting/Single Fixed Transmon/IQ_blobs.py
```diff
@@ -114,4 +74,44 @@
...
+# count = declare(int)
+# cont_condition = declare(bool)
+# assign(cont_condition, ((I > threshold) & (count < 3)))
+# with while_(cont_condition):
+#     play("pi", "qubit")
+#     align("qubit", "resonator")
+#     measure("readout", "resonator", None,
+#                 dual_demod.full("rotated_cos", "out1", "rotated_sin", "out2", I))
+#     assign(count, count + 1)
+#
```
TheoLaudatQM, line 116: I think we need to add `assign(cont_condition, ((I > threshold) & (count < 3)))` in the while statement so that it is updated

Quantum-Control-Applications/Superconducting/Single Fixed Transmon/IQ_blobs.py
```diff
@@ -114,4 +74,44 @@
...
+# with while_(I > threshold):
+#     play("pi", "qubit")
+#     align("qubit", "resonator")
+#     measure("readout", "resonator", None,
+#                 dual_demod.full("rotated_cos", "out1", "rotated_sin", "out2", I))
+#
+#########################################
+#
+# # Repeat until success active reset, up to 3 iterations
+# count = declare(int)
```
TheoLaudatQM, line 107: When we declare a value like this, it is automatically assigned to 0?

yomach: yes, but I added it for clarity

Quantum-Control-Applications/Superconducting/Single Fixed Transmon/IQ_blobs.py
```diff
@@ -114,4 +74,46 @@
...
+# cont_condition = declare(bool)
+# assign(cont_condition, ((I > threshold) & (count < 3)))
+# assign(count, 0)
```
TheoLaudatQM, line 110: It is not crucial but it would be more clear to place it before the cont_condition assignment where the value of count is needed

TheoLaudatQM (approved):

PR #147, opened 2022-07-21
NV center py_qua_tools update

TheoLaudatQM:


Quantum-Control-Applications/NV-Centers/Confocal-Setup/cw_odmr.py
```diff
@@ -28,7 +29,7 @@
         with for_(f, f_min, f <= f_max, f + df):  # Notice it's <= to include f_max (This is only for integers!)
             update_frequency("NV", f)  # update frequency
             align()  # align all elements
-
+            wait(20, "NV", "AOM")  # Used to synchronize AOM and SPCM (from simulation)
```
yomach, line 32: And delay has to be calibrated and added to the configuration, AOMs have delays that can be 400ns long... so getting the pulses to be the same in the simulator doesn't really help.
I suggest removing it

Quantum-Control-Applications/NV-Centers/Confocal-Setup/raw_adc_traces.py
```diff
@@ -10,26 +11,51 @@
...
+# Execute program
 job = qm.execute(TimeTagging_calibration)
-job.result_handles.wait_for_all_values()
+# create a handle to get results
 res_handles = job.result_handles
-raw_data = res_handles.get("raw_adc").fetch_all()
-plt.plot(raw_data)
-plt.title("ADC Trace Check ADCs saturation and define threshold")
+# Wait untill the program is done
+res_handles.wait_for_all_values()
```
yomach, line 41: You're using the "normal" method because there is only one?

Quantum-Control-Applications/NV-Centers/Confocal-Setup/time_rabi.py
```diff
@@ -46,46 +47,38 @@
...
-    plt.pause(0.1)
-
-    b_cont = res_handles.is_processing()
-    b_last = not (b_cont or b_last)
-
-print("")
+simulate = True
+if simulate:
+    simulation_config = SimulationConfig(
+        duration=28000, simulation_interface=LoopbackInterface([("con1", 3, "con1", 1)])
```
yomach, line 53: Any specific reason for this loopback interface? Will we see something here?

Quantum-Control-Applications/NV-Centers/Confocal-Setup/cw_odmr.py
TheoLaudatQM: Yes I removed it

Quantum-Control-Applications/NV-Centers/Confocal-Setup/raw_adc_traces.py
TheoLaudatQM: From what I understood, the idea is that we continue to teach customers how to use standard QUA and fetch data "normally" in the first scripts and then show then the tools progressively. This is the reason why the first scripts don't use the py_qua_tools but the others do

Quantum-Control-Applications/NV-Centers/Confocal-Setup/time_rabi.py
TheoLaudatQM: It is not really used but this way the syntax is already there in case they need it

yomach: I see. I would remove it, but I don't have a strong opinion.

yomach (approved):

Quantum-Control-Applications/NV-Centers/Confocal-Setup/time_rabi.py
TheoLaudatQM: I don't either so as a trade-off I removed it from the code except in hello_qua that can serve as a reference and that way we can show it to customers at the beginning of the session so that they are aware of it

yomach (approved):

PR #151, opened 2022-07-28
typos

MichalGQM:


yomach (approved):

TheoLaudatQM:
Maybe it is also worth adding the file that you modified with the slow_chirping to display the lines on the scope. I would put it in the main folder with the configuration file you use in order not to modify the existing use-case, what do you think @MichalGQM ?

MichalGQM:
Hey yes I think it's a good idea to have the slow motion option. But why not to modify the existing one?

TheoLaudatQM:
Because it is a use-case as we used it at Saffman's so I prefer not to modify it. But if think it's okay to have another file for future installations, maybe with the simplified config and the other files you made for your last installation, what do you think

MichalGQM:
Oh I see.. ok. sounds fair :) 
I am actually waiting for them to test the input stream from their camera. Hopefully it will be next week, So maybe I will make this second file after I check this - so the new file will have the slow motion and the input stream option. 
What do you think ?

TheoLaudatQM:
Perfect, it sounds good !

PR #155, opened 2022-08-12
DRAG and AC stark shift files

KevinAVR:
modification to RB so that we don't need to repeat the pulses every step -- this is relevant if RB is too long of a sequence

updated DRAG and AC stark shift pulse sequences


Quantum-Control-Applications/Superconducting/Single Fixed Transmon/DRAG_calibration_Google.py
```diff
@@ -0,0 +1,99 @@
...
+    while results.is_processing():
+        # Fetch results
+        I, Q, state, iteration = results.fetch_all()
+        # Progress bar
+        progress_counter(iteration, n_avg, start_time=results.get_start_time())
+        # Plot results
+        plt.cla()
+        plt.pcolor(iters, amps * drag_coef, state, cmap="magma")
+        plt.xlabel("Number of iterations")
+        plt.ylabel("Drag coef")
```
TheoLaudatQM, line 99: Maybe it is worth adding the line cut in python as well to get the alpha coeff easily, what do you think?

TheoLaudatQM: Same for the Yale method

Quantum-Control-Applications/Superconducting/Single Fixed Transmon/rb.py
```diff
@@ -12,11 +12,12 @@
 
 inv_gates = [int(np.where(c1_table[i, :] == 0)[0][0]) for i in range(24)]
 max_circuit_depth = int(3 * qubit_T1 / x180_len)
-delta_depth = 1
+delta_depth = 1  # must be one because the random generation of Cliffords is done 1 by 1
 num_of_sequences = 50
 n_avg = 20
 seed = 345324
 cooldown_time = 5 * qubit_T1 // 4
+delta_clifford = 10
```
TheoLaudatQM, line 20: The idea of this change is to be able to do RB with a step > 1 right ?

Quantum-Control-Applications/Superconducting/Single Fixed Transmon/rb.py
```diff
@@ -125,67 +126,84 @@ def play_sequence(sequence_list, depth):
...
         with for_(depth, 1, depth <= max_circuit_depth, depth + delta_depth):
-            with for_(n, 0, n < n_avg, n + 1):
-                # Replacing the last gate in the sequence with the sequence's inverse gate
-                # The original gate is saved in 'saved_gate' and is being restored at the end
-                assign(saved_gate, sequence_list[depth])
-                assign(sequence_list[depth], inv_gate_list[depth - 1])
+            # Replacing the last gate in the sequence with the sequence's inverse gate
+            # The original gate is saved in 'saved_gate' and is being restored at the end
+            assign(saved_gate, sequence_list[depth])
+            assign(sequence_list[depth], inv_gate_list[depth - 1])
```
TheoLaudatQM, line 140: This is actually what prevents us from setting delta_depth > 1 right ?
Maybe we can hardcode delta_depth=1 then, what do you think?

TheoLaudatQM (commented):
Very nice, I just added a few questions and comments

TheoLaudatQM (approved):

PR #161, opened 2022-09-08
Update QCA and qua-libs readme

TheoLaudatQM:
Fix the qua-libs readme and add a readme to QCA to promote the usecases


Quantum-Control-Applications/README.md
```diff
@@ -0,0 +1,40 @@
+# Quantum Control Applications
+
+This folder contains scripts and examples for different qubits types, all the way from basic to advanced protocols. 
+It also includes various examples and results from labs as listed below.
+
+Note that for them to work, you would need to download the latest version of the [py-qua-tools](https://github.com/qua-platform/py-qua-tools#installation).
+
+## Superconducting Qubits
```
yomach, line 8: I would add here first a link to all of the normal files (with linked to all of the readme files).
And then start with all of the usecases, maybe under a title such as "advanced use-cases"?

Quantum-Control-Applications/Superconducting/Single Flux Tunable Transmon/configuration.py
```diff
@@ -46,45 +46,95 @@ def IQ_imbalance(g, phi):
...
+minus_x90_Q_wf = minus_x90_der_wf
 # No DRAG when alpha=0, it's just a gaussian.
 
 y180_len = x180_len
 y180_sigma = y180_len / 5
 y180_amp = x180_amp
-y180_wf, y180_der_wf = np.array(drag_gaussian_pulse_waveforms(y180_amp, y180_len, y180_sigma, alpha=0, delta=1))
+y180_wf, y180_der_wf = np.array(
+    drag_gaussian_pulse_waveforms(
+        y180_amp, y180_len, y180_sigma, alpha=drag_coef, anharmonicity=anharmonicity, detuning=AC_stark_detuning
```
yomach, line 104: I think it's better to remove the `alpha=`, `anharmonicity=` and `detuning=`.

yomach (commented):
Very nice!
see my comments

Quantum-Control-Applications/README.md
TheoLaudatQM: I indeed wanted to do that but I got lazy... I will do it now :)

Quantum-Control-Applications/Superconducting/Single Flux Tunable Transmon/configuration.py
TheoLaudatQM: I guess that this is subjective, because I personally prefer to have the parameters names written explicitly otherwise I don't know which one is anharmonicity or detuning... I know that it is written in the docstring, but most of the time people don't look at it especially when running experiments if you want to change parameters quickly.
Maybe @KevinAVR can pick a side...

yomach: I'm ok either way, don't really care. 
Your arguments are good, I'll give mine, but then just decide whatever way you prefer :)
1. I think that if the names of the variables are meaningful (as they are here), then it's ok. It does say: `alpha=drag_coef` and `anharmonicity=anharmonicity` so the information there is doubled.
2. Not having it makes it backward compatible with older versions of `qualang_tools`. So if some customer goes now to qualibs and downloads the files (without updating `qualang_tools`), it would work for him. (Currently it won't).

TheoLaudatQM: Ok you convinced me, in this case since the variable names are explicit we don't need to keep the parameter's names :)

PR #163, opened 2022-09-20
Fix QCA links and typos

TheoLaudatQM:


yomach (commented):
Very nice!
I've seen that in the Swap use-case the link to the cryoscope is broken:
https://github.com/qua-platform/qua-libs/tree/main/Quantum-Control-Applications/Superconducting/Multiple%20Flux%20Tunable%20Transmons/Use%20Case%201%20-%20Two%20qubit%20gate%20optimization%20with%20cryoscope#two-qubit-swap-spectroscopy-improved-with-pre-distortion-digital-filters

TheoLaudatQM:
Good catch, I fixed it !

yomach (approved):

PR #164, opened 2022-10-12
Octave Demo Script

yuvalqm:
This is the code from the documentation page of octave.
With this script you should be able to operate a single RF output/input of the octave at a certain LO and IF.

Added the code and a configuration file to support it

### DIFF
--- a/Tutorials/octave/octave_demo.py
+++ b/Tutorials/octave/octave_demo.py
@@ -5,9 +5,8 @@ from qm.octave import *
 from qm.qua import *
 from configuration import *
 
-opx_ip = "xxx.xxx.xxx.xxx"
+qop_ip = "xxx.xxx.xxx.xxx"
 opx_port = 80
-octave_ip = "xxx.xxx.xxx.xxx"
 octave_port = 80
 
 octave_config = QmOctaveConfig()
--- a/Tutorials/octave/octave_demo.py
+++ b/Tutorials/octave/octave_demo.py
@@ -1,0 +1,50 @@
+import os
+import time
+
+from qm.QuantumMachinesManager import QuantumMachinesManager
+from qm.octave import *
+from qm.qua import *
+import numpy as np
+from configuration import *
+
+opx_ip = 'xxx.xxx.xxx.xxx'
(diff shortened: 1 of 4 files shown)


Tutorials/octave/octave_demo.py
```diff
@@ -0,0 +1,48 @@
...
+
+from qm.QuantumMachinesManager import QuantumMachinesManager
+from qm.octave import *
+from qm.qua import *
+from configuration import *
+
+opx_ip = "xxx.xxx.xxx.xxx"
+opx_port = 80
+octave_ip = "xxx.xxx.xxx.xxx"
+octave_port = 80
```
yomach, line 11:
```suggestion
qop_ip = "xxx.xxx.xxx.xxx"
opx_port = 80
octave_port = 80
```
We set up the devices to have the same IP but different ports. 
Of course, these changes also need to be reflected down.

Tutorials/octave/octave_demo.py
```diff
@@ -0,0 +1,48 @@
...
+element = "q1"
+
+qm.octave.set_lo_frequency(element, lo_freq)
+qm.octave.set_rf_output_gain(element, -10)
+qm.octave.set_rf_output_mode(element, RFOutputMode.on)
+
+qm.octave.set_lo_source(element, OctaveLOSource.LO1)
+
+qm.octave.calibrate_element(element, [(lo_freq, if_freq)])
+qm = qmm.open_qm(config)  # Calibration closes the QM so another one should be opened again after calibration is done
```
yomach, line 40: just realized that you have to open the qm twice.
very ugly :(

Tutorials/octave/octave_demo.py
yuvalqm: OK, I wanted it to be as general as possible in that sense

Tutorials/octave/octave_demo.py
yuvalqm: you need to open one for each calibration you do.. it is very ugly and sadly there was no attention to it in recent months :(

PR #167, opened 2022-11-03
black + rb_log

KevinAVR:
rb with log spacing in the steps

### DIFF
--- a/Quantum-Control-Applications/Superconducting/Single Fixed Transmon/rb_log_xaxis.py
+++ b/Quantum-Control-Applications/Superconducting/Single Fixed Transmon/rb_log_xaxis.py
@@ -1,0 +1,233 @@
+"""
+Performs a 1 qubit randomized benchmarking to measure the 1 qubit gate fidelity
+"""
+from qm.qua import *
+from qm.QuantumMachinesManager import QuantumMachinesManager
+from scipy.optimize import curve_fit
+from configuration import *
+import matplotlib.pyplot as plt
+import numpy as np
+from qualang_tools.bakery.randomized_benchmark_c1 import c1_table
+from macros import readout_macro
+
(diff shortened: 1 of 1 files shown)


Quantum-Control-Applications/Superconducting/Single Fixed Transmon/rb_log_xaxis.py
```diff
@@ -0,0 +1,233 @@
...
+    state_st = declare_stream()
+    d_cliffords = declare(int, value=delta_cliffords)
+    r = declare(int)
+
+    with for_(m, 0, m < num_of_sequences, m + 1):
+        sequence_list, inv_gate_list = generate_sequence()
+
+        assign(r, 0)
+
+        with for_(depth, 1, depth <= max_circuit_depth, depth + 1):
```
yomach, line 138: This is also relevant to the original code:
Why do we have this for loop with +1 and then have the if statement inside?
Isn't it easier just to change the for loop?

This would make it much easier for both codes...

Quantum-Control-Applications/Superconducting/Single Fixed Transmon/rb_log_xaxis.py
```diff
@@ -0,0 +1,233 @@
+"""
```
yomach, line 1: In general, it's good. I wondered why not use a single file for both (just having a linear or log option) and this led to my 2nd comment.
With the change in the `for` loop, that should be easy...

In any case, if we do put two files, then we need to update the readme file as well

TheoLaudatQM:
@KevinAVR Do you still want this? Otherwise I will close it

KevinAVR:
Let's close it @TheoLaudatQM  -- at some point it seemed relevant to a single customer but has not showed up in recurrent people ask to us -- so probably no need to invest time on it

PR #168, opened 2022-12-17
quasiparticle poisoning in charge sensitive qubits

KevinAVR:
the readme is missing. Files I tested in the recent install of Syracuse with the respective quam


MichalGoldenshtein:
read message

PR #172, opened 2023-03-28
changes in octave tutorial

MichalGQM:


Tutorials/intro-to-octave/set_octave.py
```diff
@@ -161,24 +160,25 @@ def octave_settings(qmm, qm, prog, octave_config, external_clock=False, calibrat
...
+        element_i = qm.elements[elements]
+        if isinstance(element_i, ElementWithOctave):
+            # This assumes that: FR1in measures RF1's output (which is connected to Analog output 1 and 2), FR2in measures RF2's output (which is connected to Analog output 3 and 4)
+            if (element_i.q_port == 1 or element_i.q_port == 2) and 'outputs' in \
+                    config['elements'][octave_elements[i]].keys():
+                qm.octave.set_qua_element_octave_rf_in_port(octave_elements[i], "octave1", 1)
+                qm.octave.set_downconversion(octave_elements[i])
+                qm.octave.set_downconversion(octave_elements[i],
+                                             lo_source=RFInputLOSource.Internal)  # Can change to Dmd1LO
+                qm.octave._set_downconversion_if_mode(octave_elements[i], if_mode_i=IFMode.direct,
```
TheoLaudatQM, line 173: This is still a protected API (with "_"), is it normal?

Tutorials/intro-to-octave/octave_introduction.py
```diff
@@ -319,59 +335,61 @@
 ###########################
 external_clock = False
 if external_clock == "10MHz":
-    qmm.octave_manager.set_clock("octave1", ClockType.External, ClockFrequency.MHZ_10)
+    qmm.octave_manager.set_clock(octave, ClockType.External, ClockFrequency.MHZ_10)
```
TheoLaudatQM, line 338: It is written in Pycharm "Do not use OctaveManager, it will be removed in the next version", it is normal right?

Tutorials/intro-to-octave/octave_introduction.py
```diff
@@ -23,7 +24,8 @@
 opx_port = 80
 octave_ip = "172.0.0.1"
 octave_port = 50
-
+con = "con1"
+octave = "octave1"
```
TheoLaudatQM, line 28: Parametrizing the controller name makes sense, but this is just a name that the user can define the way he/she wants, so why do we need to parametrize it here? (I just tested and I can do octave="superman")
If you want to test a 2nd Octave, then you need to change the port, no?

TheoLaudatQM (commented):
Just a general comment on top of what I wrote, how do you suggest to modify the Octave parameters if the user wants to set the output gain to -10dB for a specific port for instance? Should one do it in set_octave.py or directly in hello_octave.py?

yuvalqm:
@MichalGQM Please delete the Calibration_Database file it might cause conflicts for customers

Tutorials/intro-to-octave/configuration.py
```diff
@@ -1,7 +1,8 @@
-opx_ip = "172.0.0.1"
+# opx_ip = "172.0.0.1"
+opx_ip = "172.16.33.100"
 opx_port = 80
 
-octave_ip = "172.0.0.1"
+octave_ip = "172.16.33.100"
```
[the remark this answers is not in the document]
MichalGQM: You're right! Thanks

Tutorials/intro-to-octave/set_octave.py
MichalGQM: This is still working so I guess it's fine... @yuvalqm ?

Tutorials/intro-to-octave/octave_introduction.py
MichalGQM: I think it's fine. @yuvalqm ?

Tutorials/intro-to-octave/octave_introduction.py
MichalGQM: If you have an other octave so you need to change the octave name and port.. Did you try changing only the port and it worked?

MichalGQM:
Regarding the general command - If it's a global gain - i.e. you use the same with all the scripts so I think it's better to do the modifications to the `set_octave` file.

Tutorials/intro-to-octave/set_octave.py
yuvalqm: yes this is internal. you should use the parameters at qm.octave.set_downconversion for the if I/Q modes for example: 
`        qm.octave.set_downconversion(elements[0], lo_source=RFInputLOSource.Internal,if_mode_i= IFMode.direct, if_mode_q=IFMode.direct)
`

Tutorials/intro-to-octave/octave_introduction.py
yuvalqm: yes, It is going for depracation

Tutorials/intro-to-octave/set_octave.py
MichalGQM: Changed it. Thanks!

Tutorials/intro-to-octave/octave_introduction.py
TheoLaudatQM: I don't remember if I tested it or not, but I don't see any reasons why it shouldn't because the name can be anything (it is not linked to the port)

TheoLaudatQM (approved):
It looks good to me, you can merge :)

Tutorials/intro-to-octave/octave_introduction.py
MichalGQM: I think you're right

PR #174, opened 2023-05-10
Small fixes

MichalGQM:
I changed two small things that I found today...

### DIFF
--- a/Tutorials/intro-to-octave/set_octave.py
+++ b/Tutorials/intro-to-octave/set_octave.py
@@ -134,7 +134,7 @@ def octave_settings(qmm, qm, prog, config, octave_config, external_clock=False, 
         # Change to the relevant external frequency
         qmm.octave_manager.set_clock("octave1", ClockType.External, ClockFrequency.MHZ_10)
         # If using a clock from the OPT, use this command instead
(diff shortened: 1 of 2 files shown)


Tutorials/intro-to-octave/set_octave.py
```diff
@@ -134,7 +134,7 @@ def octave_settings(qmm, qm, prog, config, octave_config, external_clock=False,
         # Change to the relevant external frequency
         qmm.octave_manager.set_clock("octave1", ClockType.External, ClockFrequency.MHZ_10)
         # If using a clock from the OPT, use this command instead
-        # qmm.octave_manager.set_clock(octave, ClockType.Buffered, ClockFrequency.MHZ_1000)
+        # qmm.octave_manager.set_clock("octave1", ClockType.Buffered, ClockFrequency.MHZ_1000)
```
TheoLaudatQM, line 137: In this file (in the Octave configuration) the user has the possibility to add a second Octave so maybe it would be good to have the possibility to also configure the second Octave by looping over the octaves from the octave_config, what do you think?

MichalGQM: I intended to make it with more octaves but didn't get to it... You're right. :)

PR #180, opened 2023-06-09
state tomography typo fixed

KevinAVR:
I just updated the gate -x90 -> -y90 to be fixed, it was too annoying xD


TheoLaudatQM:
Done in the big qua-libs SC fixes

PR #191, opened 2023-07-21
The code needs adaptation

KevinAVR:
It is for 9 qb 9 resonators, in 2 octave 2 opxs, but easilty can be modified so smaller subet.

I think the correct subset will be 4qb 4 resonator. 

The motivation to have this add hand is that Lincoln labs is distribution candle devices 6qb 6resonators to people to test their fridges and this is something we can reuse in the future.


TheoLaudatQM:
@KevinAVR What is the status of this?

KevinAVR:
Hi Theo, I'm closing this pull request because the original intention was to have a code base that can have multiple qubit multiple resonators and it used the rapid quam but there is really no need for it because we are trasitioning to new quam, and now that we will have two qubit -- fixed and flux then we can expand when needed

PR #197, opened 2023-08-21
fix: maxtime for timetagging is too short

roiabirQM:
wrong measure len causes compilation error


TheoLaudatQM (approved):
Good catch, thanks!

PR #198, opened 2023-08-22
config fix

KevinAVR:
Hi Theo I just removed one element from the config which was part of installation but no needed anymore, the rest is only BLACK change


TheoLaudatQM (approved):
Alright, I didn't check but I trust you  :-)

PR #203, opened 2023-09-05
Uniformization of SC scripts

TheoLaudatQM:
- Fix some bugs (rotation angle, opt weights, scripts)
- Fix after testing workflow and all scripts
- Add comments in each scripts
- Update according to QOP220


maximilianqm (approved):
only quickly scanned and as it is just parameter changes and slight adjustments of the nodes no reason for in-depth review

PR #209, opened 2023-12-08
Quantum Dots folder

TheoLaudatQM:
Two folders:
1) Single spin with EDSR (Octave or not)
2 Singlet Triplet qubit with exchange interaction

I created a class (defined in the config) to define the pulse sequence using sticky elements and (kind of) virtual gates with the ability to sweep the parameters in QUA and play ramps, so that the QUA program looks fairly simple. Please tell me what you think!


Quantum-Control-Applications/Quantum-Dots/Single_Spin_EDSR/02_raw_adc_traces.py
KevinAVR, line 1: when writing 

`measure('readout' * amp(0), 'TIA', ...)`

I think it would be useful to write the comment ->

# *amp(0) sets the amplitude to zero, thus the collected trace is the digitization of the analog input
# in the majority of cases the DC bias to the sensor is applied with an external DC source

Quantum-Control-Applications/Quantum-Dots/Single_Spin_EDSR/03_time_of_flight_RF.py
KevinAVR, line 1: The stream processing in this program refers to ANALOG INPUT 1, however, in the previous program raw adc traces, RF refers to input2, and in the next program spectorscopy the demodulation is using `out1` -- is that ok?

Quantum-Control-Applications/Quantum-Dots/Single_Spin_EDSR/06_charge_stability_map_opx_and_dc_source.py
KevinAVR, line 1: Hi Theo, I remeber somewhat we had a conversation about the fast pulse by the OPX whether to do it with `set_dc_offset()` or to do it with sticky elements.

As you and I know people is picky about sending pulses without ramps or at the end to ramp down to zero without any wait, but I think given that the AC arm has much more attenuation this should not be a problem actually.

Let's see how it flies when we test it with the next customer indeed.

experimentally maybe one thing to take care of is to make sure the DC measurement lenght is not so huge as to be away from the high-pass bandwidht -- like it is possible to increase the number of averages instead of sitting too long with DC measurement at each voltage right?

Quantum-Control-Applications/Quantum-Dots/Single_Spin_EDSR/03_time_of_flight_RF.py
TheoLaudatQM: Fixed it

Quantum-Control-Applications/Quantum-Dots/Single_Spin_EDSR/06_charge_stability_map_opx_and_dc_source.py
TheoLaudatQM: dc_offset doesn't work because of gaps so it is done with sticky elements instead.
The function `get_filtered_voltage()` at the beginning of the file will indicate whether the AC pulse is too long or not, then I think it is to the appreciation of the user to estimate if the distortion is okay or not.
To me it is just important to be aware of this limitation

Quantum-Control-Applications/Quantum-Dots/Singlet_Triplet_Qubit/03_time_of_flight_RF.py
KevinAVR, line 1: same comment as in the EDSR folder -- RF reflectometry is connected to `input2` but in this file it shows `input1()`

KevinAVR (approved):
I like it a lot! I'm sure is more than 90% ready and it just needs some in the field testing.

I hope I get to go to a single triplet installation to test the more advanced code tha tyou suggested :D -- thank you Theo you are amazing!

Quantum-Control-Applications/Quantum-Dots/Single_Spin_EDSR/06_charge_stability_map_opx_and_dc_source.py
TheoLaudatQM: You are right, but it is only in this script because usually people don't care too much about gaps.
If needed, the script can be modified to add ramps and sticky elements instead, but so far I never needed to use it, so I suggest to see how customers will react before changing it

Quantum-Control-Applications/Quantum-Dots/Singlet_Triplet_Qubit/03_time_of_flight_RF.py
TheoLaudatQM: Yes, I fixed it but actually the tank circuit and TIA element are connected to both inputs in the config to have more flexibility

PR #210, opened 2023-12-13
Fix typo in active reset condition.

deanpoulos:
Here it looks like a copy paste caused the condition for the x180 active reset pulse to refer to a variable `I`, instead of the intended variable `I_reset`.


TheoLaudatQM (approved):
Indeed good catch :)

PR #211, opened 2023-12-13
Add documentation hint for internally configured octave.

deanpoulos:
Add hint in documentation for how to configure the octave when in network configurations C or D.


Quantum-Control-Applications/Superconducting/Single-Fixed-Transmon/configuration_with_octave.py
```diff
@@ -32,6 +32,10 @@
 octave_1 = OctaveUnit("octave1", qop_ip, port=11050, con="con1")
 # octave_2 = OctaveUnit("octave2", qop_ip, port=11051, con="con1")
 
+# If the instruments or QM router are connected directly to this machine, use the local octave IP and port 80.
```
TheoLaudatQM, line 35: Here you mean that the control PC (what you call machine) is connected directly to the internal network of the router, right?
If so I would write it explicitly because otherwise it can be confusing since the PC can also be connected to port 1 of the router...

TheoLaudatQM (changes requested):

Quantum-Control-Applications/Superconducting/Single-Fixed-Transmon/configuration_with_octave.py
TheoLaudatQM: @deanpoulos ?

deanpoulos: Are you able to comment here a clearer wording from your perspective? Then I will add a repo-wide search and replace. I'm only asking you to do this because you have broader experience of all the different network configurations than I do...

TheoLaudatQM: Suggestion:
`If the control PC or local network is connected to the internal network of the QM router (port 2 onwards) or directly to the Octave (without QM the router), use the local octave IP and port 80.`

deanpoulos: Great, on it

TheoLaudatQM (approved):
I am happy :)

PR #221, opened 2024-04-14
Added XEB script

arthurostrauss:
This script implements Cross-Entropy Benchmarking which leverages real-time processsing features of the OPX. Namely, the random circuit generation and execution is done on the OPX directly. We also propose tools from Qiskit (should be installed) and Cirq (no need to be installed) for enriching and strengthening the post-processing.


Quantum-Control-Applications/Superconducting/Two-Flux-Tunable-Transmons/Use Case 3 - Two-Qubit Cross-Entropy Benchmarking/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+The protocol relies on the ability to perform random circuits on the quantum computer and compare the results with the ideal simulation. 
+The advantage of using QUA is that the randomization of the circuits can be done in parallel to circuit execution through real-time processing and random sampling.
+
+The user can choose which gate set to use to generate random unitaries. Usually, the experiment is performed with layers of random single-qubit gates followed by one fixed two-qubit gate. 
+The script will generate random circuits with the chosen gate set and run them on the quantum computer. The script will then calculate the cross-entropy between the ideal and actual probability distributions to estimate the layer fidelity (from which the two-qubit gate fidelity can be inferred).
+
+As opposed to Randomized Benchmarking, we do not invert the randomly generated circuit by applying an inverse, but we rather perform a fidelity estimation over the statistics of the outcomes when measuring the system in the computational basis.
+
+There are therefore three steps in the script:
```
deanpoulos, line 19: four?

deanpoulos:
Since I'm unable to comment on the notebook itself, I will add comments roughly cell-by-cell.

deanpoulos:
**Cell 1** 
I don't think you should run `!pip install qiskit` by default, since this could mess with a user's conda fragile environment. Better to remove the try-except even though it's helpful, and just leave the pip command as a comment, e.g.,
```
# !pip install qiskit
from qiskit.circuit import QuantumCircuit
from qiskit.circuit.library import get_standard_gate_name_mapping as gate_map, UnitaryGate, RYGate
```

deanpoulos:
**Cell 3**
Just a thought for the future, when there are many inputs into the protocol (which you call hyperparameters), you can put them into a dataclass, e.g.,

```
@dataclass
class XEBConfig:
    save_dir: str = ""  # Directory where the data will be saved
    should_save_data: bool = True  # Whether to save the data
    generate_new_data: bool = True  # Whether to generate new data
    seqs: int = 100  # Number of random sequences to run per depth
    max_depth: int = 3  # Max depth of the XEB sequence
    step: int = 1  # Step for choosing which depths to run up to max_depth
    depths: np.ndarray = field(init=False)  # Array of depths to iterate through
    n_shots: int = 101  # Number of averages per sequence
    impose_0_cycle: bool = False  # Whether to impose the first gate at 0-cycle
    apply_two_qb_gate: bool = False

    def __post_init__(self):
        self.depths = np.arange(1, self.max_depth + 1, self.step)
```

The benefit to doing this is that you can easily save the config parameters to file, or serialize into a dictionary, allowing you to add it directly to the experiment results in e.g., part of the params in an `xarray` DataArray:
```
from dataclasses import asdict
config = XEBConfig(...)
d = config.to_dict()
```

This isn't usually done in ML, but can be very handy in physics where you share around data for portable analysis by other researchers, or where you do so many runs and forget with what parameters you performed the run.

For now it's fine, but I thought I'd mention it because this is what I'd do if I was writing a protocol from scratch.

deanpoulos:
**Cell 4**
```
Path().cwd()
```
is this needed?

deanpoulos:
**Cell 5**
Here you generate the gatesets, and offer as a chose 1 or 2. I suggest you move all this logic into a separate python file behind a function which, e.g.,

**gatesets.py**
```python
gateset = generate_gateset(gateset="supremacy")
```

The notebook is currently designed to be very transparent which is good, but for a new user looking to just run the notebook and see the results, this result gives too much detail.

I also suggest putting the gateset choices as a string to make make things more explicit.

deanpoulos:
**Cell 6**
Here you seem to define some critical parameters for running the experiment. I think it's important to prompt the user directly to change what is there, and keep the templates clear, e.g.,

```python
# change this so that it follows the name of your qubit elements.
#  for example, this is how it would look if your element was "qubit1$xy"
qubit_elements = [f"qubit{i}$xy" for i in qubits]  # qubit elements associated to qubits

# change this so that it follows the name of your elements.
readout_elements = [f"qubit{i}$rr" for i in qubits]  # readout resonators elements associated to qubits

# change this to configure CZ operations on different qubit pairs
#  key: tuple of qubit indices to perform CZ
#  value: tuple following element, operation
CZ_operations = {
    (0, 2): ("qubit2$z", "Cz$flux_pulse_control_qubit0"),
    # (1, 2): ("qubit2$z", "Cz$flux_pulse_control_qubit1"),
    # (2, 1): ("qubit2$z", "Cz$flux_pulse_control_qubit1"),
    # (2, 0): ("qubit2$z", "Cz$flux_pulse_control_qubit0"),
    # (2, 3): ("qubit3$z", "Cz$flux_pulse_control_qubit2"),
    # (3, 2): ("qubit3$z", "Cz$flux_pulse_control_qubit2"),
}
```

I also commented out those which I think are unused, given that the default is `qubits = [0,2]`

deanpoulos:
**Cell 8**
Here you define some useful QUA macros. First, I'm concerned that in some of the macros, it implicitly assumes the template of some of the qubits, e.g., the following macro
```python
def align_qubit(qubit: int):
    base_qubit = f"qubit{qubit}$"
    elements = [base_qubit + el for el in ["xy", "rr", "z"]]
    align(*elements)
```
assumes that your elements have a dollar sign at the end. Instead, the macros should use a template similar to what you define in **Cell 6**, for example,
```python
qubit_element_template = "qubit{}"
...
def align_qubit(qubit: int, qubit_element_template: str):
    base_qubit = qubit_element_template.format(qubit)
    elements = [base_qubit + el for el in ["xy", "rr", "z"]]
    align(*elements)
...
align_qubit(qubit, qubit_element_template)
```

The same assumption is made in the `readout` macro.

Next, I think you should separate out all of the macros that _don't_ require user input into a separate file called `macros.py`. This way, you keep the entire top half of the notebook dedicated to "things users should change.

deanpoulos:
**Cell 9**
The program is very compact and makes use of macros which is great and helps with readability. One thing I wanted to mention is that you use a lot of in-line comments, which can make it a little hard to read. The official [python style-guide](https://peps.python.org/pep-0008/#inline-comments) and coding books alike discourage in-line comments, saying they should be used sparingly. I think it's easier to read when the comment is above. The style-guide is worth a skim if you haven't already seen it.

I also wanted to comment on the code that looks like this:
```python
with while_(g[q][d_] == g[q][d_ - 1])
```
With the comment you have, it makes it obvious what this code is doing, but in general, one-letter variables can be hard to track because you have to map in your head what the variable is referring to. Since in this case, each variable can be represented with a short word `g` --> `gates`, `q` --> `qubit`, `d_` --> `depth` or `depth_`, it doesn't increase the length too much to just use the full word and helps with readability.

Otherwise I haven't tried to hard to verify the logic of your code, but I trust that it's right if the results look good in a real experiment.

deanpoulos:
**General Comments**
The rest of the code seems to be post-processing, data loading, result verification, etc. On this, I want to add that:

1. (optional) Since you allow loading old data, you could make a separate notebook for analysis of results. This makes life easier when re-visiting old data for analysis, so you don't get confused/distracted by the parameters/code above which is dedicated to running a new experiment.
2. Try to put as many general functions for post-processing in separate file(s). This helps readability and testability.
3. Anything dedicated to your personal testing/verification should be removed from the final notebook(s).

Otherwise, it looks really good!

deanpoulos:
**Cell 3**
Now that we have a config, it is readable enough to add the parameters directly into the config definition, combining cells 3 and 4, also removing some redundant comments. In general, variable names should be expressive enough (which they currently are) so that you don't need such comments,

```python
step = 1
max_depth = 3
depths = np.arange(1, max_depth + 1, step),

# Define the configuration for the XEB sequence
xeb_config = XEBConfig(
    seqs=100,
    depths=depths,
    n_shots=101,
    impose_0_cycle=False,
    apply_two_qb_gate=False,
    gate_set_choice="sw",
    save_dir="",
    should_save_data=True,
    generate_new_data=True,
)
print(f"Depths array: {xeb_config.depths}")
```

Quantum-Control-Applications/Superconducting/Two-Flux-Tunable-Transmons/Use Case 3 - Two-Qubit Cross-Entropy Benchmarking/README.md
deanpoulos: @arthurostrauss

Quantum-Control-Applications/Superconducting/Two-Flux-Tunable-Transmons/Use Case 3 - Two-Qubit Cross-Entropy Benchmarking/gateset.py
deanpoulos, line 1: This file is formatted well as a standalone python module, however I think the style should follow a more software-library approach, in which you don't really need a docstring at the top of the file, and all the relevant info should be in the function docstring. This is especially necessary in preparing to move this code to py-qua-tools.

That way, when you hover over it from the notebook, you know exactly what the allowed values for `gate_set` are. I have asked ChatGPT to move the info at the top into the docstring, as well as make the error message more informative, and it did a pretty good job,

```python
from qiskit.circuit.library.standard_gates import get_standard_gate_name_mapping as gate_map
from qiskit.circuit.library import UnitaryGate, RYGate
from scipy.linalg import sqrtm
import numpy as np


def generate_gate_set(gate_set: str) -> dict:
    """
    Generate a gate set from a string for random single qubit gates for XEB.

    This function generates a dictionary of single-qubit gates for XEB (Cross-Entropy Benchmarking)
    based on the provided gate set string.

    The available gate sets are:
    - "sw": X/2, Y/2, and SW gates. SW gate is defined as the square root of the W gate.
    - "t": X/2, Y/2, and T gates. T gate is a pi/4 rotation around the Z axis.

    Each single qubit gate is assumed to be playable through the modulation of the amplitude
    matrix of a baseline calibrated X/2 (SX) pulse.

    Args:
        gate_set (str): String indicating the desired gate set. Allowed values are "sw" or "t".

    Returns:
        dict: Dictionary containing the generated single qubit gates.
    Raises:
        ValueError: If an invalid gate set string is provided.
    """

    SX, T, X, Y = [gate_map()[gate] for gate in ["sx", "t", "x", "y"]]
    SY = RYGate(np.pi / 2)

    # Define the relevant amplitude matrices for all single-qubit gates (all generated through initial X90 gate)
    X90_dict = {"gate": SX, "amp_matrix": np.array([1.0, 0.0, 0.0, 1.0])}
    Y90_dict = {"gate": SY, "amp_matrix": np.array([0.0, -1.0, 1.0, 0.0])}
    gate_dict = {0: X90_dict, 1: Y90_dict}
    if gate_set == "sw":
        W = UnitaryGate((X.to_matrix() + Y.to_matrix()) / np.sqrt(2), label="W")
        SW = UnitaryGate(sqrtm(W), label="sw")  # from Supremacy paper
        SW_dict = {"gate": SW, "amp_matrix": 0.70710678 * np.array([1.0, -1.0, 1.0, 1.0])}
        gate_dict[2] = SW_dict
    elif gate_set == "t":
        T_dict = {
            "gate": T,
            "amp_matrix": np.array([1.0, 0.0, 0.0, 1]),
        }  # No actual need for amp_matrix (done with frame rotation)
        gate_dict[2] = T_dict
    else:
        raise ValueError(f"Invalid gate set: {gate_set}. Allowed values are 'sw' or 't'.")

    return gate_dict
```

I replaced the `str` type hint with `Literal['sw', 't']`. A `Literal` is like a more specific version of a string, which only takes on certain values. In PyCharm, this will give you a warning if your input to the function is not 'sw' or 't'.

deanpoulos:
**Cell 6**
I like how you added the `qubit_element_template`. I have extended it's usage into the `xy` and `rr` element definitions, so you only have to change the template in one place:

```python
# Change this so that it follows the name of your qubit elements.
# This is how it would look if your elements start with "qubit0$",
#  since the qubit index will eventually fill in the {} curly braces
qubit_element_template = "qubit{}$"
# And this is how it would look if one of your elements was "qubit0$xy"
qubit_elements = [f"{qubit_element_template.format(i)}xy" for i in qubits]
readout_elements = [f"{qubit_element_template.format(i)}rr" for i in qubits]
```

deanpoulos:
> **Cell 3** Now that we have a config, it is readable enough to add the parameters directly into the config definition, combining cells 3 and 4, also removing some redundant comments. In general, variable names should be expressive enough (which they currently are) so that you don't need such comments,
> 
> ```python
> step = 1
> max_depth = 3
> depths = np.arange(1, max_depth + 1, step),
> 
> # Define the configuration for the XEB sequence
> xeb_config = XEBConfig(
>     seqs=100,
>     depths=depths,
>     n_shots=101,
>     impose_0_cycle=False,
>     apply_two_qb_gate=False,
>     gate_set_choice="sw",
>     save_dir="",
>     should_save_data=True,
>     generate_new_data=True,
> )
> print(f"Depths array: {xeb_config.depths}")
> ```

**Cell 7**
Now, anywhere else in the notebook, you should reger to e.g. `seqs` as `config.seqs`:

```python
if n_qubits < 2:
    config.apply_two_qb_gate = False

# For Data Saving:
qubits_involved = "_".join([str(x) for x in qubits])
comment = f"s({config.seqs})_d({max_depth})_g({config.gate_set_choice})_cz({int(config.apply_two_qb_gate)})_0cyc({int(config.impose_0_cycle)})"
filename = f"XEB_q{qubits_involved}_{comment}"
```

deanpoulos (approved):

PR #222, opened 2024-05-03
integration weights change

KevinAVR:
Previously the integration weight had to be taken at every clock cycle and for very long readout it can be problematic cuz of overhead but now in the integration weight program it returns the list of tuples to put directly in config dictionary

### DIFF
--- a/Quantum-Control-Applications/Superconducting/Single-Fixed-Transmon/configuration.py
+++ b/Quantum-Control-Applications/Superconducting/Single-Fixed-Transmon/configuration.py
@@ -155,13 +155,11 @@ depletion_time = 2 * u.us
 
 opt_weights = False
 if opt_weights:
-    from qualang_tools.config.integration_weights_tools import convert_integration_weights
-
     weights = np.load("optimal_weights.npz")
-    opt_weights_real = convert_integration_weights(weights["weights_real"])
-    opt_weights_minus_imag = convert_integration_weights(weights["weights_minus_imag"])
-    opt_weights_imag = convert_integration_weights(weights["weights_imag"])
-    opt_weights_minus_real = convert_integration_weights(weights["weights_minus_real"])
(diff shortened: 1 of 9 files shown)


Quantum-Control-Applications/Superconducting/Single-Fixed-Transmon/configuration.py
```diff
@@ -155,13 +155,11 @@ def IQ_imbalance(g, phi):
...
-    opt_weights_real = convert_integration_weights(weights["weights_real"])
-    opt_weights_minus_imag = convert_integration_weights(weights["weights_minus_imag"])
-    opt_weights_imag = convert_integration_weights(weights["weights_imag"])
-    opt_weights_minus_real = convert_integration_weights(weights["weights_minus_real"])
+    opt_weights_real = weights["weights_real"]
+    opt_weights_minus_imag = weights["weights_minus_imag"]
+    opt_weights_imag = weights["weights_imag"]
+    opt_weights_minus_real = weights["weights_minus_real"]
 else:
     opt_weights_real = [(1.0, readout_len)]
```
deanpoulos, line 164: Hey Kevin, I know this isn't in the scope of your solution, but are the weights in the un-optimized `else` section wrong? Shouldn't the first index (weight) of each tuple be 1, 0 or -1?

deanpoulos (approved):
Approving because it looks good and solves this weird `division_length` quirk/problem. Just see the comment on the "default" opt_weights in the `else` condition.

Quantum-Control-Applications/Superconducting/Single-Fixed-Transmon/configuration.py
KevinAVR: Hi Dean, yes its wrong an ill defined , I just fixed it

PR #224, opened 2024-05-14
Allow delta_clifford=1 in all single qubit RB

TheoLaudatQM:


deanpoulos (approved):

deanpoulos:
Hi @TheoLaudatQM, the logic of the program with `depth_target` and so on is a little hard to follow. However, I was able to validate both the old and new RB using `qua-qsim` and couldn't find any huge difference between the output RB curves of the programs

TheoLaudatQM:
Great, I am merging then!

PR #226, opened 2024-05-15
QDs class fixes

TheoLaudatQM:
Replace _Variable and _Expression by QuaVariable, QuaExpression
Cast level as float
Add error when level doesn't have the right format


KevinAVR (commented):
It runs in the new qm-qua version good -- I found a missing import np inside of macros and added a plt show for VS code

deanpoulos (approved):

PR #227, opened 2024-05-17
Proposed changes to quam_2transmons

nulinspiratie:
## Proposed changes 
- Move all components up one level
		- Into `Quantum_control_applications/Superconducting/quam/quam_components`
		- Added `./quam/Five-Flux-Tunable-Transmons/components.py` that points one level up
		- Importable as `from components import QuAM`
- Use nested dicts for data handler results
		- I showed this for 18_allXY, but this structure can be applied to all nodes
- Rename quam_component files
		- Instead of e.g. `FluxLine_component.py`, I named it `flux_line.py`
		- Using lowercase with underscore is standard so I suggest we stick to it.
		- We can discuss whether to keep `_component`
- Make `ReadoutResonator.f_01/12` proper class parameters
		- TODO Handle properly in quam structure
- Make `Transmon.f_01/12` proper class parameters
		- TODO Issue arises since f_12 is defined by anharmonicity
- Change `QuAM.get_depletion_time/get_thermalization time` to `QuAM.depletion_time, thermalization_time`
		- In the previous form, it had `@property` so it was behaving as an attribute
		- However, the name `get_depletion_time`suggests it is a function.
		- So we should either do `def get_depletion_time` (without `@property`), or `@property def depletion_time`
		- @theo preferences?
- Change `QuAM.calibrate_qubits` to `QuAM.calibrate_active_qubits_octave`
		- `calibrate_qubits` can be better used for full qubit calibration
## Open questions and comments
- Should we create dedicated `XyChannel`?
		- `Transmon.set_gate_shape()` arguably belongs there
- In the future I think we should decouple the `DataHandler` from QuAM. However, for now it's fine to combine the two


Quantum-Control-Applications/Superconducting/quam/Two-Flux-Tunable-Transmons/18_allxy.py
```diff
@@ -199,9 +199,7 @@ def get_prog(qubit, resonator):
             plt.tight_layout()
             plt.pause(0.1)
 
-            data[f"{qb.name}_I"] = I
-            data[f"{qb.name}_Q"] = Q
-            data[f"{qb.name}_figure"] = fig
+            data[qb.name] = {"I": I, "Q": Q, "figure": fig}
```
TheoLaudatQM, line 202: You cannot save a figure in a tree-like structure, it has to be in data = {"figure": fig}, otherwise you get 
FileNotFoundError: [Errno 2] No such file or directory: 'C:\\git\\qua-libs\\Quantum-Control-Applications\\Superconducting\\quam\\Two-Flux-Tunable-Transmons\\DATA\\2024-05-17\\#51_all_xy_125420\\q0\\figure.png'

nulinspiratie: Will try to fix it asap

nulinspiratie: Should be fixed by https://github.com/qua-platform/py-qua-tools/pull/215

Quantum-Control-Applications/Superconducting/quam/Five-Flux-Tunable-Transmons/99_test.py
```diff
@@ -0,0 +1,182 @@
+"""
```
TheoLaudatQM, line 1: What is this script doing?

Quantum-Control-Applications/Superconducting/quam/quam_components/quam.py
```diff
@@ -67,10 +67,12 @@ def connect(self) -> QuantumMachinesManager:
...
         """
         return QuantumMachinesManager(
-            host=self.network["host"], cluster_name=self.network["cluster_name"], octave=self.octave.get_octave_config()
+            host=self.network["host"],
+            cluster_name=self.network["cluster_name"],
+            octave=self.octave.get_octave_config(),
         )
 
-    def calibrate_active_qubits(self, QM: QuantumMachine) -> None:
+    def calibrate_active_qubits_octave(self, QM: QuantumMachine) -> None:
```
TheoLaudatQM, line 75: Maybe `calibrate_octave_ports` would make more sense, wdyt?

TheoLaudatQM (commented):
Just minor comments, but before disclosing it, we also need the ability to have a state for 5 qubits but only connect to two as we discussed

Quantum-Control-Applications/Superconducting/quam/Five-Flux-Tunable-Transmons/99_test.py
nulinspiratie: Err nothing, good catch... It wasn't supposed to be included. I will remove the file

Quantum-Control-Applications/Superconducting/quam/quam_components/quam.py
nulinspiratie: agreed

Quantum-Control-Applications/Superconducting/quam/Five-Flux-Tunable-Transmons/99_test.py
nulinspiratie: Err nothing, it wasn't supposed to be included, will remove

nulinspiratie:
> Just minor comments, but before disclosing it, we also need the ability to have a state for 5 qubits but only connect to two as we discussed

Agreed, we can add this as a later feature right?

TheoLaudatQM:
I still think it is important to have a way to do even if it is not perfect not perfect and I thought about a possible solution which shouldn't be to hard to implement, let's discuss it!

nulinspiratie:
Ok great, shall we merge this and discuss/implement in your branch?

deanpoulos (approved):

TheoLaudatQM (commented):
Now there is no components folder anymore, so the path where quam is imported `(from components import QuAM)` must be changed to point to the `quam_components` folder I guess.
I don't know if this must be done in this PR or in the other brancc...

nulinspiratie:
@TheoLaudatQM it still works, there's a components file. A later plan is to have a dedicated `quam_components` module, in a subsequent PR. Merging this PR as discussed

PR #228, opened 2024-05-28
Add schematic for octave introduction script.

deanpoulos:
Add a nice figure showing how to connect different components to perform different parts of the `octave_introduction.py` script.


MichalGQM (changes requested):
Hey, 
Nice idea!! 
I do have some comments: 
Step 1 - in the current QOP version, we need to configure the clock from the API. That means  that we need to set and internal or external clock. So it's not only for external clock. 
Step 2 -  In order to check up converters we have to connect the analog outputs to the IQ pairs. 
Step 3 -  In order to check triggers we have to connect the analog outputs to the IQ pairs,  and the digital outputs to the Trig ports. 
Step 4 -  In order to check down converters we have to connect the analog outputs to the IQ pairs, the RF out to the RF in and the IFOUT1, IFOUT2 to Analog inputs. 
Step 5 -  In order to check calibration we have to connect the analog outputs to the IQ pairs and the IFOUT1, IFOUT2 to Analog inputs.

deanpoulos:
Thanks for the feedback @MichalGQM.

I updated Step 1 to clarify that setting the clock is not optional.

As for your other comments, I believe all of that information is captured already, but I improved the schematic a little bit to make the wire colour to intro-step correspondence a little clearer.

Please confirm whether it is ready to merge

deanpoulos:
Hi @MichalGQM, please confirm if you're happy with this now :)

MichalGQM (approved):
Great! It looks really good!

PR #233, opened 2024-06-13
Quam/n tranmson support

deanpoulos:


Quantum-Control-Applications-QuAM/Superconducting/calibrations/00_hello_qua.py
```diff
@@ -1,20 +1,19 @@
...
-from components import QuAM
-from macros import node_save
+from quam_components import QuAM
 
 ###################################################
 #  Load QuAM and open Communication with the QOP  #
 ###################################################
 # Instantiate the QuAM class from the state file
-machine = QuAM.load("state.json")
+machine = QuAM.load(os.path.join('..', 'configuration', 'quam_state'))
```
nulinspiratie, line 13: We need a better solution for this, two issues I'm seeing:
- The code is difficult to read
- The actual path depends on the current working directory.

We should discuss alternatives

Quantum-Control-Applications-QuAM/Superconducting/calibrations/02_raw_adc_traces.py
```diff
@@ -44,15 +43,17 @@
...
 
     with for_(n, 0, n < n_avg, n + 1):  # QUA for_ loop for averaging
-        # Make sure that the readout pulse is sent with the same phase so that the acquired signal does not average out
-        reset_phase(rr1.name)
-        reset_phase(rr2.name)
-        # Measure the resonator (send a readout pulse and record the raw ADC trace)
-        rr1.measure("readout", stream=adc_st)
-        # Play the readout on rr2 as well for making sure that the ADC won't be saturated for multiplexed readout
-        rr2.measure("readout")
+        for i in range(len(resonators)):
```
nulinspiratie, line 46: What about `for i, resonator in enumerate(resonators)`?

Quantum-Control-Applications-QuAM/Superconducting/calibrations/06_resonator_spectroscopy_vs_amplitude.py
```diff
@@ -31,48 +30,47 @@
...
 ###################
 # The QUA program #
 ###################
 
 n_avg = 100  # The number of averages
 
-rr1.operations["readout"].amplitude = 0.01
-rr2.operations["readout"].amplitude = 0.01
+# Initial readout amplitude for all resonators
+for rr in resonators:
```
nulinspiratie, line 63: I wonder whether we should set the amplitude here. What if the user already calibrated it previously?

Quantum-Control-Applications-QuAM/Superconducting/calibrations/macros.py
nulinspiratie, line 1: I'm not sure what to do about this file. It doesn't belong in calibrations, but also not in `configurations`, nor in `quam_components`...

Quantum-Control-Applications-QuAM/Superconducting/configuration/calibration_db.json
nulinspiratie, line 1: Remove this file?

Quantum-Control-Applications-QuAM/Superconducting/configuration/generate_superconducting_quam.py
```diff
@@ -0,0 +1,312 @@
...
+    """
+    Create a wiring config tailored to the number of qubits.
+
+    """
+    wiring = {"qubits": {}}
+
+    # Generate example wiring by default
+    xy_lines, flux_lines, res_lines = _example_wiring(num_qubits)
+
+    # Uncomment this to manually set wiring for each qubit
```
nulinspiratie, line 25: I think it's better to try and avoid people modifying the functions themselves. We should think about if there's a good way to take this configurability out of the function

Quantum-Control-Applications-QuAM/Superconducting/configuration/generate_superconducting_quam.py
```diff
@@ -0,0 +1,312 @@
...
+
+def create_default_wiring(num_qubits: int) -> dict:
+    """
+    Create a wiring config tailored to the number of qubits.
+
+    """
+    wiring = {"qubits": {}}
+
+    # Generate example wiring by default
+    xy_lines, flux_lines, res_lines = _example_wiring(num_qubits)
```
nulinspiratie, line 23: Do we need another function `_example_wiring`? Doesn't this function create a default wiring?

Quantum-Control-Applications-QuAM/Superconducting/configuration/generate_superconducting_quam.py
```diff
@@ -0,0 +1,312 @@
...
+
+    # Uncomment this to manually set wiring for each qubit
+    # res_lines  = [("con1", 1, "octave1", 1)]*num_qubits
```
nulinspiratie, line 26: My own preference would be to group these lines per qubit, e.g. something like
```
qubit_lines = [
    {"xy": ("con1", 1), "readout": ...}
]
```
Any reason for grouping it the other way?

Quantum-Control-Applications-QuAM/Superconducting/configuration/generate_superconducting_quam.py
```diff
@@ -0,0 +1,312 @@
...
+                "opx_input_Q": (res_con, 2),
+                "digital_port": (res_con, res_I_ch_out),
+                "frequency_converter_up": "#/octaves/octave1/RF_outputs/1",
+                "frequency_converter_down": "#/octaves/octave1/RF_inputs/1",
+            },
+        }
+    return wiring
+
+
+def add_default_transmon_pulses(transmon):
```
nulinspiratie, line 57: We should discuss whether we want to have this function (see Jira issue)

Quantum-Control-Applications-QuAM/Superconducting/configuration/generate_superconducting_quam.py
```diff
@@ -0,0 +1,312 @@
...
+    if octaves is not None:
+        machine.octaves = octaves
+        print("If you haven't configured the octaves, please run: octave.initialize_frequency_converters()")
+    else:
+        # Add the Octave to the quam
+        for i in range(math.ceil(num_qubits / 4)):
+            # Assumes 1 Octave for every 4 qubits
+            octave = Octave(
+                name=f"octave{i+1}",
+                ip=octave_ip,
```
nulinspiratie, line 189: Octave ip and port should instead reference `machine.network`

Quantum-Control-Applications-QuAM/Superconducting/configuration/generate_superconducting_quam.py
```diff
@@ -0,0 +1,312 @@
...
+
+    machine = create_quam_superconducting(num_qubits=5)
+    machine.save(quam_folder, content_mapping={"wiring.json": {"wiring", "network"}})
+
+    qua_file = folder / "qua_config.json"
+    qua_config = machine.generate_config()
+    json.dump(qua_config, qua_file.open("w"), indent=4)
+
+    quam_loaded = QuAM.load(quam_folder)
+    qua_file_loaded = folder / "qua_config2.json"
```
nulinspiratie, line 310: I understand the need to see whether the loaded QuAM can also generate a config, but what about saving to the same file so we don't get people asking why there is also `qua_config2.json`?

nulinspiratie (changes requested):
Here are my comments, let me know what you think @deanpoulos. I also should have some time to work on it next week

Quantum-Control-Applications-QuAM/Superconducting/calibrations/00_hello_qua.py
deanpoulos: Good catch, I agree that the actual path should be relative to the script, not the CWD. I have changed this, but I don't know if the readability can be improved. Thoughts on:

```python
# Get the directory of the current script
current_dir = os.path.dirname(os.path.abspath(__file__))
# Define a path relative to this script, i.e., ../configuration/quam_state
config_path = os.path.join(current_dir, '..', 'configuration', 'quam_state')
# Load the machine state
machine = QuAM.load(config_path)
```
?

Quantum-Control-Applications-QuAM/Superconducting/calibrations/06_resonator_spectroscopy_vs_amplitude.py
deanpoulos: Since the amplitude sweep is performed w.r.t the current amplitude (via a scaling factor), even if a user has calibrated it previously, it is more convenient to:

1. Change the amplitude value to something "central" from within the script to provide a more meaningful range of amplitudes, even if it is not on resonance.
2. Change it from within the script rather than state.json for more efficient exploration, even if previously calibrated.

I think it's fine to keep for these reasons, but also to demonstrate how to change the amplitude of the machine from python, and to open a discussion about sweeping amplitudes using scaling factors.

Quantum-Control-Applications-QuAM/Superconducting/calibrations/macros.py
deanpoulos: The only other sensible suggestion I can think of is:
```
.
├── calibrations
├── configuration
├── macros.py <----- here
├── pyproject.toml
├── quam_components
└── README.md
```
but, this leads to really annoying imports. Making it installable seems complex and excessive. For now, since `calibrations` is not yet calibration nodes, can we just leave it as-is?

Quantum-Control-Applications-QuAM/Superconducting/configuration/calibration_db.json
deanpoulos: Done

Quantum-Control-Applications-QuAM/Superconducting/configuration/generate_superconducting_quam.py
deanpoulos: Are they not already doing this in the `create_quam_superconducting` function, to specify their network settings etc.?

We could always unwrap the `create_default_wiring` function into that one, but it then defeats the purpose. 

I agree that this semi-function-ized approach is a little clunky. Perhaps we lay everything out in the main file if that's how you feel.

deanpoulos: As a follow-up, as this file is used only once, I see it as a way to generate the boiler-plate code, or the QuAM foundation, which will be changed in more targeted ways as experiments go on. I don't think it's usability is a super important detail, although it would be nice to have something easy to work with.

Quantum-Control-Applications-QuAM/Superconducting/configuration/generate_superconducting_quam.py
deanpoulos: I guess it depends on what you mean by default wiring. 

The way I see it is that there is the default `wiring` generated by the `create_default_wiring` function, whose purpose is to more conveniently populate the fields of the Transmon, as a structured dictionary.

However, this did include an implicit choice of how to allocate wires for different qubit lines. For example, is the first connection a flux line or a resonator line? Do you group consecutive channels/fems by qubit or by line type (e.g. flux, resonator, then xy).

This choice should be configurable for a user if they need to some things around, but have a sensible default for many qubits so that the user doesn't have to manually assign a qubit's FEM and channel for it's multiple lines. This is the purpose of the `_example_wiring` which generates a more distilled "allocation" of wires.

Let me know if this context gives you an idea of how to improve its clarity/usability.

Quantum-Control-Applications-QuAM/Superconducting/configuration/generate_superconducting_quam.py
deanpoulos: The default allocation I made groups channels first by resonator lines, then by XY lines, then by flux lines. This provides the most convenient physical wiring between the Octave and OPX+ for example imagine this at 5-qubits
wiring-scheme

Since they are physically grouped by channel "type", they are grouped by channel type in the code. 

Let me know your thoughts- happy to change it.

Quantum-Control-Applications-QuAM/Superconducting/calibrations/00_hello_qua.py
nulinspiratie: I'm looking for a solution where we don't need to do any path manipulation in a script, but maybe we do this for a follow-up PR.
For now, I suggest we use pathlib which should look a bit nicer:
```
from pathlib import Path
config_path = Path(__file__).parent.parent / "configuration" / "quam_state"
```

Quantum-Control-Applications-QuAM/Superconducting/calibrations/06_resonator_spectroscopy_vs_amplitude.py
nulinspiratie: Your points make sense, but I do see it being overwriten as an issue because it's not clear from the script name that this would happen.

For example, let's say someone has calibrated their entire system including readout pulses. They then notice some inconsistent result, so decide to perform this measurement to see a punch-out. The results look as expected, so they continue their measurements, but now suddenly they can't do proper qubit readout anymore. After some debugging they find the reason to be that their readout amplitude has changed in this script.

wdyt? An alternative is to have this commented out by default and only overwrite it when uncommented

nulinspiratie: Or another option is that I was planning to implement `play(amplitude=...)` instead, which would automatically convert it to the right amplitude scale. But I won't have time until next week for this

Quantum-Control-Applications-QuAM/Superconducting/calibrations/macros.py
nulinspiratie: yes let's leave it as is for now. But let's keep thinking about it, I can also imagine other similar files will pop up later on, we need a proper place for it

Quantum-Control-Applications-QuAM/Superconducting/calibrations/06_resonator_spectroscopy_vs_amplitude.py
deanpoulos: Yep, you make a good point. For now I will make it commented.

Regarding your point about using something like: `play(amplitude=...)` - it is very convenient usability-wise, but it doesn't solve the fact that the `amp()` command used to input the scale factor has an input range of only -2..2, meaning you can _at most_ double your existing amplitude from the config.

So, unless you start with a high amplitude, you will not have access to a broad range of voltages.

Quantum-Control-Applications-QuAM/Superconducting/calibrations/00_hello_qua.py
deanpoulos: I agree.

I have implemented your suggestion for now.

deanpoulos:
From @KevinAVR 

Hello guys, what is the best way to drop a comment on a github repo about something I noticed?for example, I went to N_transmonsthen I went to quam_components/readout_resonator.py

@quam_dataclass
class ReadoutResonator(InOutIQChannel):
    """QuAM component for a readout resonator

    Args:
        depletion_time (int): the resonator depletion time in ns.
        frequency_bare (int, float): the bare resonator frequency in Hz.
    """

    depletion_time: int = 1000
    frequency_bare: float = None

    f_01: float = None
    f_12: float = None

I believe I understand that what you want to say is that f_01 and f_12 are the corresponding sensitive points of the resonator frequency to read out best ge or ef, but usually f_01 and f_12 are common jargon for the qubit transition, so I believe is wortwhile considering a new name in this quam component, but I did not know how to comment on it on git

deanpoulos:
Closing this one in favour of a direct merge from sub-branch to main

PR #237, opened 2024-07-05
Feat/auto quam state

nulinspiratie:
- Changes `pyproject.toml` to use Poetry
- Adds `configuration` and `macros` as importable python projects
- Auto save / load `quam` to `configuration/quam_state`
- Auto tries to save additional files, no longer needs to be hardcoded


Quantum-Control-Applications-QuAM/Superconducting/pyproject.toml
deanpoulos, line 1: Maybe since these are global packages, we call them e.g., `quam_configuration`, `quam_macros`? the existing names are a little too general

nulinspiratie: yeah I like it

PR #238, opened 2024-07-18
Feature/interleaved 2q rb validation

deanpoulos:
- Fixed likely bug in interleaved RB
- Validated the the sequences are correct
- Added fidelity calculation convenience function for interleaved gate fidelity
- Not yet tested on a device.


Quantum-Control-Applications/Superconducting/Two-Flux-Tunable-Transmons/Use Case 2 - Two-Qubit-Randomized-Benchmarking/two_qubit_rb/RBResult.py
```diff
@@ -32,6 +32,53 @@ def plot_hist(self, n_cols=3):
             self.data.state.sel(circuit_depth=circuit_depth).plot.hist(ax=ax, xticks=range(4))
         plt.tight_layout()
 
-    def plot_fidelity(self):
+    def plot_decay(self):
```
arthurostrauss, line 35: I would add a small docstring describing what it returns if this is a public method accessible to the user

Quantum-Control-Applications/Superconducting/Two-Flux-Tunable-Transmons/Use Case 2 - Two-Qubit-Randomized-Benchmarking/two_qubit_rb/RBResult.py
```diff
@@ -32,6 +32,53 @@ def plot_hist(self, n_cols=3):
...
         fidelity = (self.data.state == 0).sum(("repeat", "average")) / (self.num_repeats * self.num_averages)
         fidelity.rename("fidelity").plot.line()
+
+    def plot_fidelity(self):
```
arthurostrauss, line 39: Same comment here, just to clarify how plot_decay differentiates from plot_fidelity, would be nice to have a small docstring.

arthurostrauss (changes requested):
Looks good! I would just add a few docstrings on the public methods of the class that the user might want to use directly, so that one knows exactly what is being computed/plotted.

Quantum-Control-Applications/Superconducting/Two-Flux-Tunable-Transmons/Use Case 2 - Two-Qubit-Randomized-Benchmarking/two_qubit_rb/RBResult.py
deanpoulos: Hey Arthur, thanks for the suggestion.

I added docstrings everywhere and also renamed the methods to make them more clear. Please tell me what you think now.

Quantum-Control-Applications/Superconducting/Two-Flux-Tunable-Transmons/Use Case 2 - Two-Qubit-Randomized-Benchmarking/two_qubit_rb/RBResult.py
deanpoulos: Done.  Please tell me what you think

arthurostrauss (approved):

PR #245, opened 2024-09-02
Use Case 4 - Paraoanu Lab - Phase Modulated Pulses

michaelaeqm:
Adding example code for phase modulated pulses from Paraoanu Lab


deanpoulos:
Hi @michaelaeqm, thanks for making this!

I assume you meant to add a new folder called: `Use Case 4 - Paraoanu Lab - Phase Modulated Pulses`, but I think you put the code and `README.md` directly in the base `Single-Flux-Tunable-Transmon` folder. If that's the case, could you please restore the existing README in the base folder, then move your notebook+readme into a new folder?

TheoLaudatQM:
Close this PR since another one was reviewed and merged

PR #247, opened 2024-09-16
moved upconverter freq to the port

yomach:
These changes are the breaking changes introduced in QOP 3.2.0

PR #249, opened 2024-09-29
Update configs 3.2 mw fem

KevinAVR:
Updated LF-FEM comment that `offset` should be absolute values, and in MW-FEM now
upconverters is part of analog ouputs as well as in the elements the upconverters.

Tested in Beta_8, the mw- and lf-fem's programs


yomach:
Hi, 
See https://github.com/qua-platform/qua-libs/pull/247 which I added two weeks ago and addresses part of what is in this PR as well

PR #264, opened 2024-11-17
S cquam feature dataloading

TomDvirQM:
1. Added data loading feature which is fully tested for all node, except RB and time of flight
2. added a multiplexed run option for most nodes
3. changed the flux setting to a machine based macro on all nodes
4. qubit spec vs flux using pulses rather than DC


deanpoulos (approved):
Hi Tom, following our call I'm approving the PR. I haven't checked for minor bugs, but they should show up in testing if there are any.

PR #267, opened 2024-11-28
Single demod for Quantum Dots 

paulQM:
Spotted this issue while visiting Katsaros group. Reflectometry measurements use single demod but the demod2volts was in dual demod mode.


TheoLaudatQM (approved):
Well spotted, very goop point!

PR #269, opened 2024-12-13
cloud simulator username and password

KevinAVR:
for the demo


TheoLaudatQM:
I don't think that this makes sense to be honest, these are not used in the scripts... what is it for?

PR #272, opened 2025-01-08
Update configs 3.2 mw fem and import VoltageGateSequence

TheoLaudatQM:
Update the configurations for the MW fem with QOP32X
Remove the Voltage gate seqeunce for the QDs configs and import the VoltageGateSequence from qualang-tools instead


Quantum-Control-Applications/Quantum-Dots/Single_Spin_EDSR/configuration.py
deanpoulos, line 1: Is this a mistake? I see  that you renamed the class `OPX_virtual_gate_sequence` --> `VoltageGateSequence`, but probably you meant to import it, right?

deanpoulos (changes requested):

Quantum-Control-Applications/Quantum-Dots/Single_Spin_EDSR/configuration.py
TheoLaudatQM: Good catch, yes indeed it was a mistake!

deanpoulos (approved):

PR #274, opened 2025-01-18
Update two-qubit-process-tomography.py

bornman-nick:
Fix qubit1/qubit2 typo, add frame reset and phase reset note, and fix align qua function call typo


MichalGQM (approved):
Nice catch! Thanks for fixing those typos and adding the comment!

PR #275, opened 2025-01-22
Power Rabi conversion

omrieoqm:


deanpoulos (changes requested):
Can you please rename `Power_Rabi` and any other module names to be lowercase? https://peps.python.org/pep-0008/#package-and-module-names

Quantum-Control-Applications-QuAM/Superconducting/calibration_graph/04_Power_Rabi.py
```diff
@@ -18,65 +18,64 @@
...
+       simulation_duration_ns = 2500,
+       timeout = 100,
+       load_data_id = None,
+       multiplexed = True
+    ),
+)
 
 
 # %% {Initialize_QuAM_and_QOP}
 # Class containing tools to help handling units and conversions.
```
deanpoulos, line 65: Please remove redundant comments everywhere

Quantum-Control-Applications-QuAM/Superconducting/calibration_graph/04_Power_Rabi.py
```diff
@@ -18,65 +18,64 @@
...
-    qubits = machine.active_qubits
-else:
-    qubits = [machine.qubits[q] for q in node.parameters.qubits]
 num_qubits = len(qubits)
+# Open Communication with the QOP
+if node.parameters.load_data_id is None:
+    qmm = machine.connect()    
 
 
 # %% {QUA_program}
```
deanpoulos, line 81: Aside from removing redundant comments, inline comments should be used even more sparingly. https://peps.python.org/pep-0008/#inline-comments

The code itself here is "self-documenting".  It is obvious that 
```python
n_avg = node.parameters.num_averages  # The number of averages
                                     ^^^^^^^^^^^^^^^^^ unnecessary
```
is the number of averages.

deanpoulos (changes requested):
```python
amps = np.arange(
    node.parameters.min_amp_factor,
    node.parameters.max_amp_factor,
    node.parameters.amp_factor_step,
)
# Number of applied Rabi pulses sweep
if N_pi > 1:
    if operation == "x180":
        N_pi_vec = np.arange(1, N_pi, 2).astype("int")
    elif operation in ["x90", "-x90", "y90", "-y90"]:
        N_pi_vec = np.arange(2, N_pi, 4).astype("int")
    else:
        raise ValueError(f"Unrecognized operation {operation}.")
else:
    N_pi_vec = np.linspace(1, N_pi, N_pi).astype("int")[::2]
```
The code here fits into the category of defining sweep ranges, and I just like Ramsey, I want these to be put into functions in the `parameters.py` class for Power Rabi.

There are three main reasons for this:
1. It's not very insightful for the user to see all of the if-else statements which go into constructing the sweep range for different node parameters- they only usually care about one, and even then, don't need to see what it's doing
2. If you separate them into functions like [how it's done in Ramsey](https://github.com/qua-platform/qua-libs/blob/feature/SCquam/Quantum-Control-Applications-QuAM/Superconducting/quam_libs/experiments/ramsey/parameters.py), it makes it harder for bugs to hide. Each different branch of the if-else statement is put into it's own function (if the name doesn't match the behaviour, there is a bug) and the functions become testable if needed.
3. These functions can be called at any point, using only the `node.parameters` as input to reliably reproduce the sweep axes, instead of having to worry about passing it around to different functions. So, the sweep axes will be safer from a. being recreated with duplicated code somewhere and b. being modified accidentally somewhere and used again incorrectly.

Quantum-Control-Applications-QuAM/Superconducting/calibration_graph/04_Power_Rabi.py
```diff
@@ -113,34 +112,44 @@ class Parameters(NodeParameters):
     a = declare(fixed)  # QUA variable for the qubit drive amplitude pre-factor
     npi = declare(int)  # QUA variable for the number of qubit pulses
     count = declare(int)  # QUA variable for counting the qubit pulses
-
-    for i, qubit in enumerate(qubits):
-        # Bring the active qubits to the minimum frequency point
-        machine.set_all_fluxes(flux_point=flux_point, target=qubit)
+ 
+    for i, multiplexed_qubits in qubits.batch():
```
deanpoulos, line 116: There is a subtle but important bug here. Before, it was simple enough to do:
```python
    for i, qubit in enumerate(qubits):
```
because each qubit was played and processed one after the other, and its index in the list, `i`, was enough to identify it uniquely in the stream processing arrays, i.e., `state[i]` corresponds to the state of `i`th qubit in the list.

However, when moving to "batched" processing, qubits can organized arbitrarily into groups of multiplexed qubits which are played sequentially. To be concrete, imagine we had the following multiplexed groups:
Group 1: "q2", "q4"
Group 2: "q1", "q3"

So not only will:
```
    for i, multiplexed_qubits in qubits.batch():
```
not work, because `qubits.batch()` is a list of dictionaries, not a list of tuples, but it would give incorrect behaviour, even if you did 
```python
    for i, multiplexed_qubits in enumerate(qubits.batch()):
```
the variable `i` doesn't correspond with the qubit being measured, but the index of the multiplexed group.

This is the reason that `multiplexed_qubits` is actually a dictionary of qubits, each with a key corresponding to it's index in the _original_ qubits list. It In the example above, it would look like this:

```python
>>> qubits.batch()
[
   {1: q2, 3: q4},
   {0: q1, 2: q3}
]
```

So if you did:
```python
for multiplexed_qubits in qubits.batch()
   for i, qubit in multiplexed_qubits.items():
      ...
      save(state[i], ...)
```
then now `i` would once again be a unique identifier for that qubit, and correct stream processing is recovered

deanpoulos: Please refactor the outer- and inner-most for loops to be like the Ramsey node so that the behaviour is correct

Quantum-Control-Applications-QuAM/Superconducting/calibration_graph/04_Power_Rabi.py
```diff
@@ -169,19 +178,10 @@ class Parameters(NodeParameters):
...
-        plt.subplot(len(samples.keys()), 1, i + 1)
-        samples[con].plot()
-        plt.title(con)
-    plt.tight_layout()
+    fig, samples = simulate_and_plot(qmm, config, power_rabi, node.parameters)
+    node.results = {"figure": fig, "samples": samples}
     # Save the figure
-    node.results = {"figure": plt.gcf()}
-    node.machine = machine
+    node.results = {"figure": fig, "samples": samples}
```
deanpoulos, line 184: You added this line twice, please remove one.

And remove the redundant comments

deanpoulos, line 184: You added this line twice, please remove one.

And remove the redundant comments

Quantum-Control-Applications-QuAM/Superconducting/calibration_graph/04_Power_Rabi.py
```diff
@@ -197,125 +197,32 @@ class Parameters(NodeParameters):
...
-        # Add the qubit pulse absolute amplitude to the dataset
-        ds = ds.assign_coords(
-        {
-            "abs_amp": (
-                ["qubit", "amp"],
-                np.array([q.xy.operations[operation].amplitude * amps for q in qubits]),
-            )
-            }
-        )
+        ds = fetch_dataset(job, qubits, state_discrimination=state_discrimination,
```
deanpoulos, line 200: I prefer if this node just takes `node.parameters` as an input, and if `amps` and `N_pi_vec` are re-generated using the new functions you implemented, and `state_discrimination` is taken directly.

Quantum-Control-Applications-QuAM/Superconducting/calibration_graph/04_Power_Rabi.py
```diff
@@ -197,125 +197,32 @@ class Parameters(NodeParameters):
...
-            "abs_amp": (
-                ["qubit", "amp"],
-                np.array([q.xy.operations[operation].amplitude * amps for q in qubits]),
-            )
-            }
-        )
+        ds = fetch_dataset(job, qubits, state_discrimination=state_discrimination, 
+                           operation=operation, amps=amps, N_pi_vec=N_pi_vec)
     else:
         node = node.load_from_id(node.parameters.load_data_id)
```
deanpoulos, line 203: FYI I noticed a bug here:
```python
    if node.parameters.load_data_id is None:
        ds = fetch_dataset(job, qubits, node.parameters)
    else:
        node = node.load_from_id(load_data_id)
        ds = node.results["ds"]
...
if node.load_data_id is None:
   # do state updates
```
This format was meant to protect against having analyzed historical data updating your current state. But notice, that once you re-load the historical data, you also reload the `node.parameters.load_data_id` which was set to `None` when the node was executed the first time. So, analyzing historical data was always overwriting your current state.json in the current, incorrect format.

To fix it, we need to do something like this:
```
 load_data_id = node.parameters.load_data_id
 node = node.load_from_id(load_data_id)
 node.parameters.load_data_id = load_data_id
```

Quantum-Control-Applications-QuAM/Superconducting/calibration_graph/04_Power_Rabi.py
```diff
@@ -197,125 +197,32 @@ class Parameters(NodeParameters):
...
-            data_max_idx = I_n.argmin(dim="amp")
-        else:
-            data_max_idx = I_n.argmax(dim="amp")
-
-        # Save fitting results
+# %% {Data_analysis}
+fit_results = fit_pi_amplitude(ds, N_pi, state_discrimination, qubits, operation, N_pi_vec)
+node.results["fit_results"] = fit_results 
+# %% {Plotting}
+fig = plot(ds, qubits, fit_results, N_pi, state_discrimination)
```
deanpoulos, line 212: I would prefer if the plotting function wasn't just called `plot`, but instead something like:
`plot_rabi_oscillations` if that's what it's doing.

Quantum-Control-Applications-QuAM/Superconducting/quam_libs/experiments/Power_Rabi/analysis.py
```diff
@@ -0,0 +1,139 @@
...
+        )
+        }
+    )
+    
+    return ds
+
+def fit_pi_amplitude(ds: xr.Dataset, N_pi: int, state_discrimination: bool, qubits: List[Transmon], operation: str, N_pi_vec: np.ndarray) -> Dict:
+    
+    """
+    Fits the Pi pulse amplitude for a given dataset.
```
deanpoulos, line 61: The indentation here in the docstring is messed up

Quantum-Control-Applications-QuAM/Superconducting/quam_libs/experiments/Power_Rabi/analysis.py
```diff
@@ -0,0 +1,139 @@
...
+from quam_libs.lib.fit import fit_oscillation, oscillation
+from quam_libs.lib.instrument_limits import instrument_limits
+
+
+def fetch_dataset(job: QmJob, qubits: List[Transmon], state_discrimination: bool, operation: str, amps: np.ndarray, N_pi_vec: np.ndarray) -> xr.Dataset:
+    """
+    Fetches raw ADC data from the OPX and processes it into an xarray dataset with labeled coordinates and attributes.
+
+    arguments:
+    - job (QmJob): the job object containing the results.
```
deanpoulos, line 17: I don't think the arguments are as important to be documented as the coordinates of the Dataset that will be returned. Consider removing them from the documentation or moving the coordinates to the top of the docstring.

Quantum-Control-Applications-QuAM/Superconducting/quam_libs/experiments/Power_Rabi/parameters.py
```diff
@@ -0,0 +1,38 @@
+from qualibrate import NodeParameters
+from typing import List, Optional, Literal
+
+from quam_libs.experiments.node_parameters import DataLoadableNodeParameters, FluxControlledNodeParameters, MultiplexableNodeParameters, QmSessionNodeParameters, QubitsExperimentNodeParameters, SimulatableNodeParameters
+
+class Parameters(
```
deanpoulos, line 6: Take a look at the [Ramsey Node parameters class](https://github.com/qua-platform/qua-libs/blob/feature/SCquam/Quantum-Control-Applications-QuAM/Superconducting/quam_libs/experiments/ramsey/parameters.py) and notice the following

I created a separate class called `RamseyParameters` which gets inherited by the final `Parameters` class used in the node.

The order of inheritance is actually very important when using dataclasses like these, because it determines which order the arguments will be provided. For Ramsey, I put it here:
```python
class Parameters(
    NodeParameters,
    SimulatableNodeParameters,
    DataLoadableNodeParameters,
    QmSessionNodeParameters,
    RamseyParameters, <------------------
    FluxControlledNodeParameters,
    MultiplexableNodeParameters,
    QubitsExperimentNodeParameters,
):
```
Since when the order of the input arguments are built, they will be built from bottom-to-top, the `RamseyParameters` will be added **4th**, leading to something that looks like this:
image
This means qubits are first, then whether you want it multiplexed, then flux control, _then_ `RamseyParameters`. I like this order, especially since the lesser-used parameters (simulation, data-loading, etc.) are at the bottom, and think it should be kept consistent for all refactored nodes.

Can you do something like this for `PowerRabi`?

Quantum-Control-Applications-QuAM/Superconducting/quam_libs/experiments/Power_Rabi/plotting.py
```diff
@@ -0,0 +1,50 @@
+from matplotlib import pyplot as plt
+from matplotlib.figure import Figure
+from quam_libs.lib.plot_utils import QubitGrid, grid_iter
+
+
+def plot(ds, qubits, fit_results, N_pi, state_discrimination) -> Figure:
+    
+    """
+    Plots the Rabi oscillation results and fit for a given dataset and qubits.
+    Parameters:
```
deanpoulos, line 10: Again, fix indenting

deanpoulos (changes requested):

omrieoqm:
> ```python
> amps = np.arange(
>     node.parameters.min_amp_factor,
>     node.parameters.max_amp_factor,
>     node.parameters.amp_factor_step,
> )
> ```

omrieoqm:
I put this into a method of the parameters class

```
# Number of applied Rabi pulses sweep
if N_pi > 1:
    if operation == "x180":
        N_pi_vec = np.arange(1, N_pi, 2).astype("int")
    elif operation in ["x90", "-x90", "y90", "-y90"]:
        N_pi_vec = np.arange(2, N_pi, 4).astype("int")
    else:
        raise ValueError(f"Unrecognized operation {operation}.")
else:
    N_pi_vec = np.linspace(1, N_pi, N_pi).astype("int")[::2]
    
```
   Not sure what you meant by separating the ifs into different functions. Would be happy to discuss this.

Quantum-Control-Applications-QuAM/Superconducting/calibration_graph/04_Power_Rabi.py
omrieoqm: Done

Quantum-Control-Applications-QuAM/Superconducting/calibration_graph/04_Power_Rabi.py
omrieoqm: Done

Quantum-Control-Applications-QuAM/Superconducting/calibration_graph/04_Power_Rabi.py
omrieoqm: I think it is alright now. I would be happy if you could double check

Quantum-Control-Applications-QuAM/Superconducting/calibration_graph/04_Power_Rabi.py
omrieoqm: Done

omrieoqm: Done

Quantum-Control-Applications-QuAM/Superconducting/calibration_graph/04_Power_Rabi.py
omrieoqm: Done. I incorporated amps and N_pi_vec as attributes of the class Parameters to avoid code duplications

Quantum-Control-Applications-QuAM/Superconducting/calibration_graph/04_Power_Rabi.py
omrieoqm: right. Done.

Quantum-Control-Applications-QuAM/Superconducting/calibration_graph/04_Power_Rabi.py
omrieoqm: Done

Quantum-Control-Applications-QuAM/Superconducting/quam_libs/experiments/Power_Rabi/analysis.py
omrieoqm: Done

Quantum-Control-Applications-QuAM/Superconducting/quam_libs/experiments/Power_Rabi/analysis.py
omrieoqm: Moved them to the top

Quantum-Control-Applications-QuAM/Superconducting/quam_libs/experiments/Power_Rabi/parameters.py
omrieoqm: Done. Please have a look to verify that this is what you intended.

deanpoulos:
> I put this into a method of the parameters class
> 
> ```
> # Number of applied Rabi pulses sweep
> if N_pi > 1:
>     if operation == "x180":
>         N_pi_vec = np.arange(1, N_pi, 2).astype("int")
>     elif operation in ["x90", "-x90", "y90", "-y90"]:
>         N_pi_vec = np.arange(2, N_pi, 4).astype("int")
>     else:
>         raise ValueError(f"Unrecognized operation {operation}.")
> else:
>     N_pi_vec = np.linspace(1, N_pi, N_pi).astype("int")[::2]
>     
> ```
> 
> Not sure what you meant by separating the ifs into different functions. Would be happy to discuss this.

I have a few comments here.

1. The `Parameters` class **is not allowed to have methods** due to a Qualibrate issue, so it needs to go from this:
```python
class RabiParameters(RunnableParameters):
    """
    Parameters for the Power Rabi experiment.
   """
    def get_n_pi_vec(self):
        ...
```
to this:
```python
class RabiParameters(RunnableParameters):
    """
    Parameters for the Power Rabi experiment.
   """
   ...

def get_n_pi_vec(parameters: RabiParameters):
   ...
```

2. The name of the function should be explanatory. `n_pi_vec` is a really bad name. The `vec` part is redundant, and at the very least, the function should be called `get_number_of_pi_pulses`.
3. By "separate" functions I meant something like this

```python
def get_number_of_pi_pulses(parameters):
    
    N_pi = self.max_number_rabi_pulses_per_sweep
    operation = self.operation_x180_or_any_90
    if N_pi > 1:
        if operation == "x180":
            N_pi_vec = _get_number_of_pi_pulses_for_x180_rabi(parameters)
        elif operation in ["x90", "-x90", "y90", "-y90"]:
            N_pi_vec = _get_number_of_pi_pulses_for_x90_rabi(parameters)
        else:
            raise ValueError(f"Unrecognized operation {operation}.")
    else:
        N_pi_vec = np.linspace(1, N_pi, N_pi).astype("int")[::2]
        
    return N_pi_vec

# separate function
def _get_number_of_pi_pulses_for_x180_rabi(parameters: Rabi parameters)
   """
   Generates the sweep axis for the number of pi-pulsese to be applied in a Rabi experiment.

   The axis starts at 1 and steps by 2 each time so that on-resonance, the state-population is always
   high (or low? not sure)
   """
   return N_pi_vec = np.arange(1, N_pi, 2).astype("int")


def _get_number_of_pi_pulses_for_x90_rabi(...):
   ...
```
It might seem redundant or more complex, but like I said it has many advantages:
1. It can be clearly documented with a docstring
2. It can be imported and tested by an individual **unit** test
3. It can be more quickly scanned by a user who is interested in how the axis is generated for their particular experiment, e.g., x180 rabi than when it's in an if-statement like before

PR #285, opened 2025-02-18
Sketch/SCquam/better node structure with working monorepo

deanpoulos:
# Overview
The [sketch/SCquam/better_node_structure](https://github.com/qua-platform/qua-libs/tree/sketch/SCquam/better_node_structure) branch has the new structure for the `quam_builder`, but broke some of the stable functionality, e.g., multiplexable nodes, front-end access. This PR fixes the broken functionality and makes some structural changes towards a nice, intermediate, stable branch for installations.

# Changes
 - Converted `Superconducting` folder into a "monorepo" structure as a placeholder for moving _some_ of the packages here into their own repos. In this form, doing `pip install -e .` in the `Superconducting` folder installs:
   - `quam_builder` (intended to move to it's own repo)
      - uses the base-level pyproject.toml, since you can't have a monorepo that only installs subpackages
   - `quam_config` (intended to _stay_ in the user-namespace)
      - uses its own pyproject.toml with separated dependencies
   - `quam_experiments` (intended to _stay_ in the user-namespace)
      - uses its own pyproject.toml with separated dependencies
   - `quam_libs` (intended to move to it's own repo)
      - uses its own pyproject.toml with separated dependencies
 - `lib.fit`, `lib.fit_utils` moved to `quam_experiments.analysis`
 - `lib.cryoscope_tools` moved to `quam_experiments.experiments.cryoscope`
 - `lib` --> `quam_libs`
 - New refactored experiments moved to `quam_experiments.experiments`
 - `BatchableList` behaviour restored and moved to `quam_experiments.experiments.parameters`
    - Moved `get_qubits_used_in_node` from `machine` to `quam_experiments.experiments.parameters`
    - I would like this to be a method of the node parameters, e.g., `node.parameter.get_qubits_...()`, but I'm waiting for https://github.com/qua-platform/qualibrate-core/pull/89
 - Removed `QUAM_STATE_PATH` environment variable support in favour of QUAlibrate config
    - My justification is that we always defined it in the QUAlibrate config for `machine.save` to work in the front-end anyway, so it is always _as available_ as the environment variable.
 - Changed `quam_libs` --> `quam_builder`
 - Changed `quam_builder.components.superconducting.architectural_elements` to `quam_builder.architecture.superconducting.components`

# Known Issues
 - Somewhere along the way we removed the `Octave` IP and port as arguments to the quam builder front-end. But, `quam`-proper lists them as **required**, even though in QOP>2.2.2 they are not required @nulinspiratie. For now, the `quam_builder` is broken for setups with Octave


nulinspiratie:
@deanpoulos will review PR, just wanted to let you know that QUAlibrate Parameters now support custom classes

Quantum-Control-Applications-QuAM/Superconducting/quam_builder/architecture/superconducting/qpu/flux_tunable_qpu.py
```diff
@@ -22,7 +22,8 @@ class QuAM(BaseQuAM):
     qubits: Dict[str, FluxTunableTransmon] = field(default_factory=dict)
     qubit_pairs: Dict[str, TransmonPair] = field(default_factory=dict)
 
-    def load(self, *args, **kwargs) -> "QuAM":
+    @classmethod
```
TheoLaudatQM, line 25: Out of curiosity, what does this change?

Quantum-Control-Applications-QuAM/Superconducting/quam_builder/architecture/superconducting/qpu/base_quam.py
```diff
@@ -64,36 +58,29 @@ def load(cls, *args, **kwargs) -> "QuAM":
             if quam_state_path is None:
                 raise ValueError(
                     "No path argument provided to load the QuAM state. "
-                    "Please provide a path or set the 'QUAM_STATE_PATH' environment variable. "
+                    "Please set the quam_state_path in the QUAlibrate config. "
```
TheoLaudatQM, line 61: This means that one can't create a QUAM without using Qualibrate, right?

Quantum-Control-Applications-QuAM/Superconducting/quam_builder/architecture/superconducting/qubit/base_transmon.py
```diff
@@ -1,7 +1,7 @@
 from quam.core import quam_dataclass
 from quam.components.channels import IQChannel, MWChannel, Pulse
 from quam import QuamComponent
-from ..architectural_elements.readout_resonator import ReadoutResonatorIQ, ReadoutResonatorMW
+from ..components.readout_resonator import ReadoutResonatorIQ, ReadoutResonatorMW
```
TheoLaudatQM, line 4: isn't it better to write the absolute path: `from quam_builder.architecture.superconducting.components` instead ?

Quantum-Control-Applications-QuAM/Superconducting/quam_builder/architecture/superconducting/qubit_pair/fixed_frequency_transmons.py
```diff
@@ -4,8 +4,8 @@
 
 from quam.core import QuamComponent, quam_dataclass
 from ..qubit.fixed_frequency_transmon import FixedFrequencyTransmon
-from ..architectural_elements.cross_resonance import CrossResonanceIQ, CrossResonanceMW
-from ..architectural_elements.zz_drive import ZZDriveIQ, ZZDriveMW
+from ..components.cross_resonance import CrossResonanceIQ, CrossResonanceMW
+from ..components.zz_drive import ZZDriveIQ, ZZDriveMW
```
TheoLaudatQM, line 8: Here again, isn't it better to write the absolute path as it is done elsewhere?

Quantum-Control-Applications-QuAM/Superconducting/quam_builder/architecture/superconducting/qubit_pair/flux_tunable_transmons.py
```diff
@@ -4,7 +4,7 @@
 
 from quam.core import QuamComponent, quam_dataclass
 from ..qubit.flux_tunable_transmon import FluxTunableTransmon
-from ..architectural_elements.tunable_coupler import TunableCoupler
+from ..components.tunable_coupler import TunableCoupler
```
TheoLaudatQM, line 7: Here again, isn't it better to write the absolute path as it is done elsewhere?

Quantum-Control-Applications-QuAM/Superconducting/quam_builder/builder/qop_connectivity/build_quam_wiring.py
```diff
@@ -1,26 +1,28 @@
...
-from quam_libs.quam_builder.qop_connectivity.create_wiring import create_wiring
+from quam_builder.architecture.superconducting.qpu import BaseQuAM, FixedFrequencyQuAM, FluxTunableQuAM
+from quam_builder.builder.qop_connectivity.create_wiring import create_wiring
 from quam.components.ports import (
     FEMPortsContainer,
     OPXPlusPortsContainer,
 )
 
+
+QuAMType = Union[BaseQuAM, FixedFrequencyQuAM, FluxTunableQuAM]
```
TheoLaudatQM, line 11: Shouldn't we use `AnyQuAM` here instead which is already defined under `architecture/superconducting/qpu/__init__.py`?

Quantum-Control-Applications-QuAM/Superconducting/quam_builder/builder/superconducting/build_quam.py
```diff
@@ -101,10 +101,13 @@ def add_pulses(machine: Union[AnyQuAM]):
             add_default_transmon_pair_pulses(qubit_pair)
 
 
-def add_octaves(machine: Union[AnyQuAM], quam_state_path: Union[Path, str]):
-    if isinstance(quam_state_path, str):
-        quam_state_path = Path(quam_state_path)
-    quam_state_path = str(quam_state_path.parent.resolve())
+def add_octaves(machine: Union[AnyQuAM], calibration_db_path: Optional[Union[Path, str]] = None):
```
TheoLaudatQM, line 104: Shouldn't the `calibration_db_path` be exposed to the user as part of the input arguments of `make_quam()`?
Otherwise it will always be None since users don't have access to this function.

Quantum-Control-Applications-QuAM/Superconducting/quam_builder/builder/superconducting/pulses.py
```diff
@@ -1,6 +1,6 @@
 from quam.components import pulses
```
TheoLaudatQM, line 1: I am in favor of modifying this file so that it exposes functions for the user to decide which pulses/gate_shapes to add to the QuAM, possibly in `initialize_quam.py`. The reason is that now we enforce the gate shape to be a DRAG_Gaussian, but maybe some users want a DRAG_Cosine instead and don't want to see that DRAG_Gaussian in the state at all.
I just write it here as a not, but I plan to make another PR for it, if you also think that it is relevant

Quantum-Control-Applications-QuAM/Superconducting/quam_experiments/analysis/fit.py
```diff
@@ -1,10 +1,9 @@
 from matplotlib import pyplot as plt
```
TheoLaudatQM, line 1: Why not putting this file under `quam_libs`?

Quantum-Control-Applications-QuAM/Superconducting/quam_experiments/analysis/fit_utils.py
```diff
@@ -2,7 +2,7 @@
 import xarray as xr
```
TheoLaudatQM, line 2: Why not putting this file under `quam_libs`?

Quantum-Control-Applications-QuAM/Superconducting/quam_experiments/pyproject.toml
```diff
@@ -0,0 +1,22 @@
...
+[tool.poetry.dependencies]
+python = ">=3.9, <3.13"
+plotly = "*"
+xarray = "*"
+qm-qua = ">=1.2.1"
+lmfit = "*"
+quam_config = "*"
+quam_builder = "*"
+scipy = "*"
+qiskit-experiments = "*"
```
TheoLaudatQM, line 19: Why do we need `qiskit-experiments`?

TheoLaudatQM (commented):
It looks very good Dean, well done!
I just added a few comments, but only minor things for this PR I believe

Quantum-Control-Applications-QuAM/Superconducting/quam_builder/architecture/superconducting/qpu/flux_tunable_qpu.py
deanpoulos: So before, `my_quam` instantiated the `QuAM` class so that `QuAM` was an object.

Because we were always doing `QuAM().load`, it was not a problem. But, I refactored it so that `my_quam` just defines the class, and it is instantiated in `make_quam.py` which I think makes more sense.

Now that we're doing something to the effect of `QuAM.load()`, the method needs to be defined as a `@classmethod`, since there is no concept of `self` at the time of calling the method.

This follows how it is implemented in the QuAMRoot (as a class method)

deanpoulos: The reason it is useful that the `load` method is overrided with the correct type hint in all of the subclasses of `BaseQuAM` is so that if someone ever instantiates it directly using ` QuAM.load()`, the IDE knows that the return type of the load call  is a `QuAM` and you get all the attribute/method hints.

Quantum-Control-Applications-QuAM/Superconducting/quam_builder/architecture/superconducting/qpu/base_quam.py
deanpoulos: Not exactly, because they can always use the hardcoded path if they really need to. 

But as it stands, we list `qualibrate` as a requirement in the same place `quam` is listed, so if they have `quam`, they should also have `qualibrate`. 

We should completely deprecate this environment variable stuff, because it's confusing, and doesn't work properly on Mac anyway,

Quantum-Control-Applications-QuAM/Superconducting/quam_builder/architecture/superconducting/qubit/base_transmon.py
deanpoulos: Yep, I just did a search & replace, I didn't change whether it was relative or absolute. Will change it though

deanpoulos: Actually, it is relative in many places still. I will leave it for now, we can address it in another PR

Quantum-Control-Applications-QuAM/Superconducting/quam_builder/architecture/superconducting/qubit_pair/fixed_frequency_transmons.py
deanpoulos: Done

Quantum-Control-Applications-QuAM/Superconducting/quam_builder/architecture/superconducting/qubit_pair/flux_tunable_transmons.py
deanpoulos: Done

Quantum-Control-Applications-QuAM/Superconducting/quam_builder/builder/qop_connectivity/build_quam_wiring.py
deanpoulos: Done

Quantum-Control-Applications-QuAM/Superconducting/quam_builder/builder/superconducting/build_quam.py
deanpoulos: If it is `None`, it correctly defaults to the parent directory of the `quam_state` folder. This is the standard way we have been doing it installations (it ends up in `quam_config`).

But, I added it as an optional argument here to be one-step closer to making it fully configurable in `make_quam`. Since you mentioned it, I will wire it all the way up to `make_quam`

deanpoulos: Done

Quantum-Control-Applications-QuAM/Superconducting/quam_builder/builder/superconducting/pulses.py
deanpoulos: It don't think it's good practice for a module in one package to expose some classes in another package by just importing it. 

Instead, we could import them all into `initialize_quam` clearly so the user can see what options they have. 

I gave it a go, see `initialize_quam.py`

Quantum-Control-Applications-QuAM/Superconducting/quam_experiments/analysis/fit_utils.py
deanpoulos: https://github.com/qua-platform/qua-libs/pull/285#discussion_r1962795724

Quantum-Control-Applications-QuAM/Superconducting/quam_experiments/pyproject.toml
deanpoulos: From searching the code, it is used in `quam_experiments.analysis.fit` in Ramsey fitting routines. 

`qiskit-experiments` is a huge requirement for something so basic, and we need to re-think that.

Quantum-Control-Applications-QuAM/Superconducting/pyproject.toml
```diff
@@ -13,17 +12,21 @@ classifiers = [
...
-tqdm = "*"
-qualibrate = "*"
-qm-qua = "1.2.1"
-colorama = "*"
+qualibrate = ">=0.1.6"
+qm-qua = ">=1.2.1"
 tomli = { version = "^2.0.1", python = "<3.11" }
+quam_libs = { path = "quam_libs", develop=true }
+quam_config = { path = "quam_config", develop=true }
+quam_experiments = { path = "quam_experiments", develop=true }
```
nulinspiratie, line 23: Why do we have quam_config and quam_experiments as dependencies rather than modules?
My original thinking was that they don't need separate pyproject.toml files, but can instead be written as two source directories

Quantum-Control-Applications-QuAM/Superconducting/quam_builder/architecture/superconducting/qpu/base_quam.py
nulinspiratie: @TheoLaudatQM pretty soon I'll add a dedicated config to quam, but it will use the same config file in `~/.qualibrate/config.toml`.  So even though it's the "qualibrate" config, it's really meant for multiple things (qualibrate, quam, qua-dashboards, etc)

Quantum-Control-Applications-QuAM/Superconducting/quam_builder/architecture/superconducting/qpu/base_quam.py
```diff
@@ -64,36 +58,29 @@ def load(cls, *args, **kwargs) -> "QuAM":
...
-            if "QUAM_STATE_PATH" in os.environ:
-                args = (os.environ["QUAM_STATE_PATH"],)
-            else:
-                raise ValueError(
-                    "No path argument provided to load the QuAM state. "
-                    "Please provide a path or set the 'QUAM_STATE_PATH' environment variable. "
-                    "See the README for instructions."
-                )
-        return super().load(*args, **kwargs)
-
```
nulinspiratie, line 87: what's happening to the load method? Don't we still need it because otherwise it doesn't know where to get the quam state path from?

Quantum-Control-Applications-QuAM/Superconducting/quam_builder/architecture/superconducting/qpu/flux_tunable_qpu.py
nulinspiratie: Ahh gotcha, yes that makes sense. It might no longer be necessary once we drop support for python 3.9 or 3.10, but I can't remember for sure

Quantum-Control-Applications-QuAM/Superconducting/quam_builder/builder/qop_connectivity/build_quam_wiring.py
```diff
@@ -1,26 +1,26 @@
...
+    connectivity: Connectivity, host_ip: str, cluster_name: str,
+    quam_instance: AnyQuAM, port: int = None
 ):
 
     machine = quam_instance
     add_ports_container(connectivity, machine)
     add_name_and_ip(machine, host_ip, cluster_name, port)
     machine.wiring = create_wiring(connectivity)
-    save_machine(machine, quam_state_path)
+    save_machine(machine)
```
nulinspiratie, line 20: How about we remove `save_machine`? If we want special save behaviour, how about we override `QUAM.save()`?

Quantum-Control-Applications-QuAM/Superconducting/quam_config/pyproject.toml
nulinspiratie, line 1: I don't think this file is needed. the parent pyproject.toml should allow installing multiple modules. Any reason you want multiple pyproject files?

Quantum-Control-Applications-QuAM/Superconducting/pyproject.toml
```diff
@@ -1,9 +1,8 @@
 [tool.poetry]
```
nulinspiratie, line 1: I recall some issues with poetry combined with an editable installation. In this case, static type checkers can't always resolve imports. We should verify that it doesn't cause issues in this case

Quantum-Control-Applications-QuAM/Superconducting/quam_experiments/analysis/fit.py
nulinspiratie: Agreed with @deanpoulos, to me `quam_builder/quam_libs` would only contain things directly relevant to quam, so the components and the tools to build a quam.
A second distinction we could make is that `quam_builder` is versioned, whereas this is not

Quantum-Control-Applications-QuAM/Superconducting/quam_experiments/experiments/qubit_spectroscopy/node.py
```diff
@@ -1,7 +1,7 @@
 from typing import List, Union
 
-from quam_libs.components.superconducting.qubit import AnyTransmon
-from experiments.time_of_flight.parameters import Parameters
+from quam_builder.architecture.superconducting.qubit import AnyTransmon
+from quam_experiments.experiments.time_of_flight.parameters import Parameters
 
 
 def get_optional_pulse_duration(qubits: List[AnyTransmon], node_parameters: Parameters):
```
nulinspiratie, line 7: This seems like a common function, does it make sense to have it within qubit_spectroscopy? Wouldn't other nodes need it?

Quantum-Control-Applications-QuAM/Superconducting/quam_experiments/pyproject.toml
nulinspiratie: Agreed, it's community maintained so there's the additional risk of breaking changes

nulinspiratie (commented):
Great work Dean! This structure makes a lot of sense.
My main comment on the monorepo is that it's not clear to me that separate pyproject.toml files are needed. Given that quam_config is very lightweight, it seems easier to me to add both `quam_config` and `quam_experiments` into the top-level pyproject.toml, something like:

```
[tool.poetry]
...
packages = [
    { include = "quam_config"},
    { include = "quam_experiments"}
]
```

My other comments are very minor, and they don't need to be handled in this PR

Quantum-Control-Applications-QuAM/Superconducting/quam_experiments/analysis/fit.py
TheoLaudatQM: @nulinspiratie Just to clarify the comment: the idea is to have two different repos:
1. `quam_builder` for the quam components and builder functions
2. `quam_libs` for utility functions used in the nodes (qua_dataset, save_utils, plotting_utils...)
I 100% agree that these shouldn't belong to `quam_builder`, but maybe` quam_libs` makes sense.

@deanpoulos exactly, if we don't include those into `quam_libs`, I am also questioning its utility. To me it is a way of ensuring proper testing and versioning of these general (not node-specific) functions which doesn't seem easy/possible with `quam_experiments`. Also these functions are also most likely to be used across different quam types and qubit platforms which justifies being part of a separate repo. What do you think?

Quantum-Control-Applications-QuAM/Superconducting/quam_builder/builder/superconducting/pulses.py
TheoLaudatQM: Yes it is exactly what I had in mind, perfect!

Quantum-Control-Applications-QuAM/Superconducting/quam_builder/builder/superconducting/build_quam.py
TheoLaudatQM: Great thanks!
This way it gives more flexibility to the user

Quantum-Control-Applications-QuAM/Superconducting/quam_builder/architecture/superconducting/qpu/flux_tunable_qpu.py
TheoLaudatQM: It is very clear, thanks Dean!

Quantum-Control-Applications-QuAM/Superconducting/quam_experiments/pyproject.toml
TheoLaudatQM: Great, let's remove it then and perform the analysis ourselves

Quantum-Control-Applications-QuAM/Superconducting/quam_builder/architecture/superconducting/qpu/base_quam.py
```diff
@@ -48,9 +44,7 @@ class BaseQuAM(QuamRoot):
     def get_quam_state_path(cls) -> Optional[Path]:
         qualibrate_config_path = Path.home() / ".qualibrate" / "config.toml"
 
-        if "QUAM_STATE_PATH" in os.environ:
-            return Path(os.environ["QUAM_STATE_PATH"])
-        elif qualibrate_config_path.exists():
+        if qualibrate_config_path.exists():
             config = tomllib.loads(qualibrate_config_path.read_text())
             quam_state_path = config.get("quam", {}).get("state_path", None)
             return Path(quam_state_path)
```
nulinspiratie, line 50: @deanpoulos just found a bug here. It should say
```
return Path(quam_state_path) if quam_state_path is not None else None
```

nulinspiratie: I changed this

Quantum-Control-Applications-QuAM/Superconducting/quam_builder/builder/qop_connectivity/build_quam_wiring.py
TheoLaudatQM: yes let's do this in another PR

Quantum-Control-Applications-QuAM/Superconducting/pyproject.toml
TheoLaudatQM: Let's sort the `pyproject.toml` discussion outside of this PR so that we can move on

Quantum-Control-Applications-QuAM/Superconducting/pyproject.toml
TheoLaudatQM: Let's sort the `pyproject.toml` discussion outside of this PR so that we can move on

PR #286, opened 2025-02-19
Latest and gratest/waveform report

MichalGQM:
Hello, 
I added the waveform report to all the scripts, besides the use cases.


Quantum-Control-Applications/Superconducting/Single-Fixed-Transmon/02_raw_adc_traces_mw_fem.py
```diff
@@ -46,15 +46,21 @@
 ###########################
 # Run or Simulate Program #
 ###########################
-simulate = False
+simulate = True
```
TheoLaudatQM, line 50: Need to revert this

TheoLaudatQM (commented):
Great, a few comments:

- Maybe we should also keep the old plotting which is just `samples.con1.plot()` at least to indicate how to do it if the customers don't have plotly or want to keep the old way, what do you think?
- For the figure to appear, I need to have this in my scripts:
```
import plotly.io as pio
pio.renderers.default = 'browser'
```
Shall we add it as well in all script, only in the configs, or is there any workaround to it?

Quantum-Control-Applications/Superconducting/Single-Fixed-Transmon/02_raw_adc_traces_mw_fem.py
MichalGQM: Weird - I see `simulate = True`.

TheoLaudatQM: Yes and we want False by default

MichalGQM:
Hey 
* I was wondering about that as well. I don't want to code to be messy...
I kept it in some of the Quantum Dots files. Do you think this will work and not be confusing? 

    # Simulates the QUA program for the specified duration
    simulation_config = SimulationConfig(duration=10_000)  # In clock cycles = 4ns
    # Simulate blocks python until the simulation is done
    job = qmm.simulate(config, hello_qua, simulation_config)
    # Get the simulated samples
    samples = job.get_simulated_samples()
    # Plot the saimulated samples
    samples.con1.plot()
    # Get the waveform report object
    waveform_report = job.get_simulated_waveform_report()
    # Cast the waveform report to a python dictionary
    waveform_dict = waveform_report.to_dict()
    # Visualize and save the waveform report
    waveform_report.create_plot(samples, plot=True)

* I didn't have to do anything for the figure to appear on the web.. maybe it depend on the plotly version? which one you are using? Do you think your solution is robust? If so, maybe it's better to add it in the configurations.

Quantum-Control-Applications/Superconducting/Single-Fixed-Transmon/02_raw_adc_traces_mw_fem.py
MichalGQM: Oh I thought you mean the other way around. Ok yes I will fix it.

TheoLaudatQM:
- I think it looks okay personally and worst case we can keep it commented. I just think that it makes sense for users to still get the syntax easily because it isn't straightforward to guess.
- I am not sure where it comes from to be honest, but it works 100% of the time with my configuration. maybe there is a way around it...

MichalGQM:
* Ok. I do agree.. I will add it. 
* So I think we can add it in the configurations.

paulQM:
Good job! One small modification I propose is to replace 

```python
waveform_report.create_plot(samples, plot=True, save_path='./')
```

by:

```python
from pathlib import Path
waveform_report.create_plot(samples, plot=True, save_path=Path(__file__).resolve())
```

This ensures you are taking the path from the file itself otherwise you take the path of the terminal that runs the script thet might be different when you hit the play button on vscode or other editors.

deanpoulos:
> Hey @paulQM I am not sure if we need to force it to be saved in the same directory as the script. save_path='./' saves it in the current working directory which might be different than where the script it stored.. Maybe I can add a comment that it saves it in the current working directory and then the user can change it if needed?

I like Paul's suggestion, because it will guarantee the waveform report html is always next to the script. Otherwise, depending on whether you use VSCode or PyCharm, it will end up either in the root project directory (which is not well-defined) or in next to the script.

The only downside I see to Paul's suggestion is that it will probably break when running in VSCode in interactive mode- which some users like to do, at least in the QuAM-based nodes, since the interactive console has no concept of the current "file".

paulQM:
> The only downside I see to Paul's suggestion is that it will probably break when running in VSCode in interactive mode- which some users like to do, at least in the QuAM-based nodes, since the interactive console has no concept of the current "file".
>

I does work for me in VSCode interactive mode. Interactive mode can access the __file__ property aswell:

image

MichalGQM:
Ok, thanks for checking @paulQM ! I will change it :)

MichalGQM:
I had an issue in my environment to get it work with 
```
waveform_report.create_plot(samples, plot=True, save_path=Path(__file__).resolve())
```
so for now, let's keep it with 
```waveform_report.create_plot(samples, plot=True, save_path='./'```

paulQM (approved):
Good job! 🚀

PR #288, opened 2025-02-24
QL-104/remove assign variable to element

paulQM:
- Remove `assign_variable_to_element()` trick from scripts.
- Closes: [QL-104](https://quantum-machines.atlassian.net/browse/QL-104)


MichalGQM:
Did we check with Support team that this is 100% not needed with QOP242X?

TheoLaudatQM:
Hey @paulQM, this one should rather be merged to the [branch related to QOP24X](https://github.com/qua-platform/qua-libs/tree/fix/qop24x) than the main since it will only be applicable when the version will be released

PR #293, opened 2025-03-04
delete communication to CS cluster

MichalGQM:
Hey, I added just a small fix but I wanted to fix it quickly so I am pushing it.


paulQM (approved):
Nice catch!

PR #302, opened 2025-04-25
Ql 158/qop33x new filters

paulQM:
Need to discuss if this is actually needed


paulQM:
Modified with the suggestions from @TheoLaudatQM :) 

Just cosmetic changes in comments

MichalGQM (approved):
Great!

PR #303, opened 2025-04-28
Feature/s cquam run action quam config

TheoLaudatQM:
Refactoring of the quam_config folder according to QL-371 and QL-372


nulinspiratie (commented):
Very nice! Minor change requests (I can also handle if needed)
- I propose to have generate_quam.py be almost empty. The only contents should be comments: a description of its purpose, and an instruction to copy the contents from a relevant file in wiring examples
- Each file should have a short description at the top about what it does + what the user can/should modify.

Also happy to handle this in a follow up, so I'll register this as a "Comment" rather than "Request Changes"

TheoLaudatQM:
Great, thanks!
I followed your comments and stand ready to merge if you apporve :)

nulinspiratie (approved):

PR #307, opened 2025-04-30
Add comments to the configs with lf and mw fems

paulQM:
Adresses:
- Delays between MW fem outputs and LF FEM outputs (band dependant)
- MW FEM full scale power new range
- Modify comment about coupled port bands 1 & 3 not possible


MichalGQM (approved):
Great!

PR #315, opened 2025-06-02
QL-405: Rabi ef node

paulQM:
Adds ef Rabi node to the qualibration graph


qualibration_graphs/superconducting/quam_config/cloud_infrastructure.py
```diff
@@ -0,0 +1,88 @@
+import importlib
```
TheoLaudatQM: We probably don't want to merge it with this file. You can add it to .gitignore to avoid having it in the PRs

qualibration_graphs/superconducting/quam_config/my_quam.py
```diff
@@ -1,7 +1,30 @@
-from quam_builder.architecture.superconducting.qpu import FixedFrequencyQuam, FluxTunableQuam
```
TheoLaudatQM: Same comment here as for the cloud_infrastructure.py

qualibration_graphs/superconducting/calibrations/04b_power_rabi.py
```diff
@@ -53,6 +53,7 @@
     parameters=Parameters(),  # Node parameters defined under quam_experiment/experiments/node_name
 )
 
+node.namespace["Rabi_ef"] = None
```
TheoLaudatQM: What is this?

qualibration_graphs/superconducting/calibrations/12_power_rabi_ef.py
```diff
@@ -0,0 +1,261 @@
...
+                            # Loop for error amplification (perform many qubit pulses)
+                            with for_(count, 0, count < npi, count + 1):
+                                qubit.xy.play("EF_x180", amplitude_scale=a)
+                        align()
+
+                        # Qubit readout
+                        for i, qubit in multiplexed_qubits.items():
+                            if node.parameters.use_state_discrimination:
+                                qubit.readout_state(state[i])
+                                save(state[i], state_st[i])
```
TheoLaudatQM: Since the readout frequency is updated at the beginning, it is very unlikely that you can do state discrimination here, no?

qualibration_graphs/superconducting/calibrations/12_power_rabi_ef.py
```diff
@@ -0,0 +1,261 @@
...
+                qubit.resonator.update_frequency(qubit.resonator.intermediate_frequency + qubit.chi)
+            align()
+
+            with for_(n, 0, n < n_avg, n + 1):
+                save(n, n_st)
+                with for_(*from_array(npi, N_pi_vec)):
+                    with for_(*from_array(a, amps)):
+                        # Qubit initialization
+                        for i, qubit in multiplexed_qubits.items():
+                            qubit.reset(node.parameters.reset_type, node.parameters.simulate)
```
TheoLaudatQM: Since the readout frequency is updated, active reset might not work, it is probably better to update the readout frequency afterwards instead, no?

qualibration_graphs/superconducting/calibrations/12_power_rabi_ef.py
```diff
@@ -0,0 +1,261 @@
...
+                            qubit.xy.update_frequency(qubit.xy.intermediate_frequency)
```
TheoLaudatQM: I would update the qubit frequency at the end of the sequence so that active reset can work

qualibration_graphs/superconducting/calibrations/12_power_rabi_ef.py
```diff
@@ -0,0 +1,261 @@
...
+                        align()
+                        for i, qubit in multiplexed_qubits.items():
+                            qubit.xy.play("x180")
+                            qubit.xy.update_frequency(qubit.xy.intermediate_frequency - qubit.anharmonicity)
+                        align()
+                        for i, qubit in multiplexed_qubits.items():
+                            # Loop for error amplification (perform many qubit pulses)
+                            with for_(count, 0, count < npi, count + 1):
+                                qubit.xy.play("EF_x180", amplitude_scale=a)
```
TheoLaudatQM: Why not doing everything within the same for loop?

qualibration_graphs/superconducting/calibration_utils/power_rabi/analysis.py
```diff
@@ -47,9 +47,11 @@ def log_fitted_results(fit_results: Dict, log_callable=None):
...
     if not node.parameters.use_state_discrimination:
         ds = convert_IQ_to_V(ds, node.namespace["qubits"])
-    full_amp = np.array(
-        [ds.amp_prefactor * q.xy.operations[node.parameters.operation].amplitude for q in node.namespace["qubits"]]
-    )
+
+    if node.namespace["Rabi_ef"] is not None:
+        full_amp = np.array([ds.amp_prefactor * q.xy.operations["EF_x180"].amplitude for q in node.namespace["qubits"]])
+    else:
+        full_amp = np.array([ds.amp_prefactor * q.xy.operations[node.parameters.operation].amplitude for q in node.namespace["qubits"]])
```
TheoLaudatQM: I don't like too much having this parameter, because if complexifies the logic unnecessarily and forces to have it in power_rabi.py. Why not just setting node.parameters.operation to EF_x180?

qualibration_graphs/superconducting/calibrations/12_power_rabi_ef.py
```diff
@@ -0,0 +1,261 @@
...
+# Be sure to include [Parameters, Quam] so the node has proper type hinting
+node = QualibrationNode[Parameters, Quam](
+    name="12_power_rabi_ef",  # Name should be unique
+    description=description,  # Describe what the node is doing, which is also reflected in the QUAlibrate GUI
+    parameters=Parameters(),  # Node parameters defined under quam_experiment/experiments/node_name
+)
+
+node.namespace["Rabi_ef"] = True
+node.parameters.operation = "x180"
+node.parameters.max_number_pulses_per_sweep = 1
```
TheoLaudatQM: These parameters shouldn't be hardcoded here, either the default from power rabi are good, or we should create specific parameters from EF

qualibration_graphs/superconducting/calibration_utils/power_rabi/analysis.py
```diff
@@ -111,10 +113,16 @@ def _extract_relevant_fit_parameters(fit: xr.Dataset, node: QualibrationNode):
...
+        if node.namespace["Rabi_ef"] is not None:
+            current_amps = xr.DataArray(
+                [q.xy.operations["EF_x180"].amplitude for q in node.namespace["qubits"]],
+                coords=dict(qubit=fit.qubit.data),
+            )
+        else:
+            current_amps = xr.DataArray(
+                [q.xy.operations[node.parameters.operation].amplitude for q in node.namespace["qubits"]],
+                coords=dict(qubit=fit.qubit.data),
+            )
```
TheoLaudatQM: Same here

qualibration_graphs/superconducting/calibrations/12_power_rabi_ef.py
```diff
@@ -0,0 +1,261 @@
...
+        "nb_of_pulses": xr.DataArray(N_pi_vec, attrs={"long_name": "number of pulses"}),
+        "amp_prefactor": xr.DataArray(amps, attrs={"long_name": "pulse amplitude prefactor"}),
+    }
+    for qubit in qubits:
+        # Check if the qubit has the required operations
+        if hasattr(qubit.xy.operations, "EF_x180"):
+            continue
+        else:
+            x180 = qubit.xy.operations["x180"]
+            qubit.xy.operations["EF_x180"] = dataclasses.replace(x180, alpha=0.0)
```
TheoLaudatQM: Nice, I didn't test it, but I like it :)

TheoLaudatQM (changes requested):
Nice, I have a few general comments for which we should find a solution, but we are on the right track!

qualibration_graphs/superconducting/calibrations/12_power_rabi_ef.py
paulQM: good point, i think we should try to make it run with ef state discrimination.

qualibration_graphs/superconducting/calibrations/12_power_rabi_ef.py
paulQM: it could run on active_gef reset once it is calibrated

qualibration_graphs/superconducting/quam_config/cloud_infrastructure.py
paulQM: agreed, i modified to merge into `QL-203-node-conversion` branch so we can delete it there before merging to `main`

qualibration_graphs/superconducting/calibrations/12_power_rabi_ef.py
paulQM: good point, i'll fix it

qualibration_graphs/superconducting/calibration_utils/power_rabi/analysis.py
paulQM: makes sense, i will refactor it

qualibration_graphs/superconducting/calibrations/12b_power_rabi_ef.py
```diff
@@ -92,7 +90,6 @@ def create_qua_program(node: QualibrationNode[EfParameters, Quam]):
     # Register the sweep axes to be added to the dataset when fetching data
     node.namespace["sweep_axes"] = {
         "qubit": xr.DataArray(qubits.get_names()),
-        "nb_of_pulses": xr.DataArray(N_pi_vec, attrs={"long_name": "number of pulses"}),
         "amp_prefactor": xr.DataArray(amps, attrs={"long_name": "pulse amplitude prefactor"}),
     }
     for qubit in qubits:
```
TheoLaudatQM, line 95: Would this line work if you don't use a DRAG pulse that doesn't have a parameter called alpha?
`qubit.xy.operations["EF_x180"] = dataclasses.replace(x180, alpha=0.0)`

qualibration_graphs/superconducting/calibrations/12b_power_rabi_ef.py
```diff
@@ -124,44 +121,43 @@ def create_qua_program(node: QualibrationNode[EfParameters, Quam]):
...
-                                qubit.resonator.measure("readout", qua_vars=(I[i], Q[i]))
-                                save(I[i], I_st[i])
-                                save(Q[i], Q_st[i])
-                        align()
+                with for_(*from_array(a, amps)):
+                    # Qubit initialization
+                    for i, qubit in multiplexed_qubits.items(): # Reset the qubit to ground state
+                        qubit.reset(reset_type=node.parameters.reset_type, simulate=node.parameters.simulate)
+                        if node.parameters.reset_type == "thermal": # Wait twice the regular thermal time
+                            qubit.wait(2 * qubit.thermalization_time // 2)
```
TheoLaudatQM: Why reset + 2*thermalization time?

TheoLaudatQM (changes requested):

qualibration_graphs/superconducting/calibrations/12b_power_rabi_ef.py
paulQM: good point, I checked it and it won't work. I added a fix

qualibration_graphs/superconducting/calibrations/12b_power_rabi_ef.py
```diff
@@ -0,0 +1,264 @@
...
+            with for_(n, 0, n < n_avg, n + 1):
+                save(n, n_st)
+                with for_(*from_array(a, amps)):
+                    # Qubit initialization
+                    for i, qubit in multiplexed_qubits.items():  # Reset the qubit to ground state
+                        qubit.reset(reset_type=node.parameters.reset_type, simulate=node.parameters.simulate)
+                        if (
+                            node.parameters.reset_type == "thermal"
+                        ):  # Wait twice the regular thermal time for proper |f> state reset
+                            qubit.wait(qubit.thermalization_time * u.ns)
```
TheoLaudatQM:
```suggestion
                        # Wait twice the regular thermal time for proper |f> state reset
                        if node.parameters.reset_type == "thermal":
                            qubit.wait(qubit.thermalization_time * u.ns)
```

TheoLaudatQM (approved):

PR #316, opened 2025-06-03
Additional default gates and transmon methods

HanaGiatQM:


Quantum-Control-Applications-QuAM/Superconducting/quam_libs/components/gate_macros.py
```diff
@@ -184,4 +183,63 @@ class IdMacro(QubitMacro):
...
+
+    """
+
+    def apply(self):
+        self.qubit.xy.play('y90')
+        self.qubit.xy.play('x180')
+
+
+@quam_dataclass
+class XGate(QubitMacro):
```
arthurostrauss, line 198: The convention that we chose for the macros is that the positional arguments are reserved for the specification of the necessary gate parameters at the "logical" level. For instance, if you were to create a macro for RX(theta), the definition should look like:
def apply(self, theta:float, **kwargs).
For gates which do not contain any logical parameters, as it is the case for X, Y, Hadamard and SX, no positional arguments should be allowed in the call of apply.

In contrast the keyword arguments are reserved for specific pulse level specifications that a hardware aware user might want to tweak dynamically at the physical level.

Quantum-Control-Applications-QuAM/Superconducting/quam_libs/components/gate_macros.py
```diff
@@ -184,4 +183,63 @@ class IdMacro(QubitMacro):
...
+@quam_dataclass
+class UGate(QubitMacro):
+    def _rz(self, angle):
+        self.qubit.xy.frame_rotation(angle)
+
+    def _x90(self):
+        self.qubit.xy.play('x90')
+
+    # implementation based on https://docs.quantum.ibm.com/api/qiskit/qiskit.circuit.library.U3Gate
+    def apply(self, *args, **kwargs):
```
arthurostrauss: Following the previous logic, *args should not be variadic but instead contain explicitly the three required arguments theta, phi, lambda. One thing we could even enforce is that those arguments have to be passed positionally to avoid any mistake.

Quantum-Control-Applications-QuAM/Superconducting/quam_libs/components/gate_macros.py
```diff
@@ -184,4 +183,63 @@ class IdMacro(QubitMacro):
...
+        lambda_ = args[2]
+        self._rz(lambda_)
+        self._x90()
+        self._rz(theta + np.pi)
+        self._x90()
+        self._rz(phi + np.pi)
+
+
+@quam_dataclass
+class U1Gate(UGate):
```
arthurostrauss: Same thing as previous comments for this gate, specifically address the theta angle, not variadic.

Quantum-Control-Applications-QuAM/Superconducting/quam_libs/components/transmon.py
```diff
@@ -92,7 +92,7 @@ def inferred_anharmonicity(self) -> float:
             raise AttributeError(f"Error inferring anharmonicity for channel {name}: {self.f_12=} is not a number")
         return self.f_12 - self.f_01
 
-    @property
+    #@property
```
arthurostrauss: This should not just be commented but actually removed as it is clearly an error.

Quantum-Control-Applications-QuAM/Superconducting/quam_libs/components/gate_macros.py
```diff
@@ -55,31 +63,8 @@ def __post_init__(self) -> None:
...
-                         max_attempts=self.max_attempts)
-            # self.qubit.reset_qubit_active(
-            #     pi_pulse_name=get_pulse_name(pi_pulse),
-            #     readout_pulse_name=get_pulse_name(readout_pulse),
-            #     max_attempts=self.max_attempts,
-            # )
-        else:
-            # Thermalize the qubit
-            self.qubit.wait(self.thermalize_time // 4)
+    def apply(self) -> None:
```
arthurostrauss: The kwargs should still be part of the definition here for ensuring flexibility.

arthurostrauss (changes requested):
There is a bit of inconsistencies between gates to handle some keyword arguments. In principle, kwargs only arguments should be enabled for all gates, even those not containing virtual parameters. One way to office the specification of those arguments by keywords (hence avoiding the confusion between the two types) is to define apply as:
def apply(args, *, **kwargs)

And of course if the gate doesn't have args then simply def apply(*, **kwargs).
This is specifically important for the reset but could also be true for any gate, even if you do the decomposition yourself into sub gates. 

On a side note, it is often better when fidelities are not great force the high level circuits to be first decomposed manually into the set of native gates through a high level transpiler. The main drawback I see by doing it a the QUAM level as suggested here is that we lose the ability to perform circuit simplifications that would have been obvious to do at a higher level of the stack and will now be straightforwardly mapped to low level, non optimized definitions. But I reckon it is a request from a user so I won't expand on that. I would just advise against putting those macros as part of the native QUAM state, because we'll have an over complete set of basis gates that could prevent the automated high level optimization when using default transpiler settings.

arthurostrauss (approved):
LGTM! It is therefore expected that the users willing to adjust pulse level arguments with the kwargs redefine their macros with custom implementations. We should provide one example in the documentation but this is probably Quam level. @nulinspiratie , FYI. There is today no explanation in the QuAM documentation about the difference between the positional and keyword only arguments of the Macros. This should be explained somewhere with a concrete example like RX for instance.

PR #317, opened 2025-06-06
update qua version requirement to 1.2.2

paulQM:


MichalGQM:
@yomach, I changed the requirements to be qm-qua==1.2.1, since for example 
`octave_config.set_calibration_db` is moved to the QuantumMachinesManager class, which will be used in the scripts (as can be seen for example in the `.../Superconducting/Single-Fixed-Transmon/configuration_with_octave.py` in this branch.

yomach (approved):

PR #322, opened 2025-06-11
Latest and greatest/optically adressable spin qubits configuratios

MichalGQM:
I added all the missing configuration to optically addressable spin qubits folder.


paulQM (approved):
Looks good! 🚀

Issue #320, opened 2025-06-10
Two qubit RB unable to load pickled file


SatyaBade12:
Hi,

I'm trying to run [two qubit RB example code](https://github.com/qua-platform/qua-libs/blob/main/Quantum-Control-Applications/Superconducting/Two-Flux-Tunable-Coupled-Transmons/Use%20Case%202%20-%20Two-Qubit-Randomized-Benchmarking/two_qubit_rb_example.py), but the code fails when attempting to load the file `symplectic_compilation_XZ.pkl`, resulting in an error (likely due to a mismatch between the Python environment/versions used for pickling and unpickling)

```
Traceback (most recent call last):
  File "/home/jovyan/work/personal/two_qubit_rb_example.py", line 10, in <module>
    from two_qubit_rb import TwoQubitRb
  File "/home/jovyan/work/personal/two_qubit_rb/__init__.py", line 1, in <module>
    from .TwoQubitRB import TwoQubitRb
  File "/home/jovyan/work/personal/two_qubit_rb/TwoQubitRB.py", line 11, in <module>
    from .RBBaker import RBBaker
  File "/home/jovyan/work/personal/two_qubit_rb/RBBaker.py", line 11, in <module>
    from .gates import GateGenerator, gate_db
  File "/home/jovyan/work/personal/two_qubit_rb/gates.py", line 282, in <module>
    gate_db = _GateDatabase()
              ^^^^^^^^^^^^^^^
  File "/home/jovyan/work/personal/two_qubit_rb/gates.py", line 210, in __init__
    self._commands, self._tableaus, self._symplectic_range, self._pauli_range = self._gen_commands_and_tableaus()
                                                                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jovyan/work/personal/two_qubit_rb/gates.py", line 216, in _gen_commands_and_tableaus
    compilation = pickle.load(f)
                  ^^^^^^^^^^^^^^
TypeError: LineQubit.__new__() missing 1 required positional argument: 'x'
```
Would it be possible to share the code used to generate the pickle?
Thanks!

deanpoulos:
Hi @SatyaBade12 ,

Even though the issue seems to happen when the pickle is loaded, I actually think it's occurring because of an incorrect `cirq` version.

Can you try again after reinstalling with version `cirq==1.2.0`?

[...]

PR #318, opened 2025-06-06
Latest and greatest/qu alibrate

MichalGQM:
Hey, 
This is the PR for creating a tutorial on QUAlibrate. I am using the time of flight script for this tutorial.

PR #251, opened 2024-10-08
SC quam with Qualibrate nodes

TheoLaudatQM:
SC calibration nodes with QuAM and Qualibrate.
Reviewing steps:
- Comprehensive readme with explanations about the folder structure and the installation guide
- Test the generation of quam and ensure that it is user friendly (make wiring, make_quam and modify_quam)
- Have a clean and working requirements.txt for easing the installation process --> wirer in the main of py-qua-tools + official qualibrate
- Figure out the structure of quam-libs: separate repo? In qualang-tools? Unversioned reference code? (required even if we only merge the generation of quam)

PR #325, opened 2025-06-24
feat: Robust Analysis (nodes 02a, 02b, 02c and 04b) and Standardized Plotting Framework

shanto268:
### Summary

This contribution introduces two enhancements to the Open Source Calibration Graph (OSCG): a more robust analysis framework for calibration nodes - 02a, 02b, 02c and 04b - and a new standardized, interactive plotting system (implementation example for node 02a). 

### Robust Analysis for Calibration Nodes

The core analysis logic for the calibration nodes mentioned above has been optimized to better handle common failure modes, such as poor data quality or fit convergence issues. The new approach ensures that analysis routines complete gracefully, provide clearer outcomes, and are easier to debug.

**Key Improvements:**

-   **Graceful Error Handling**: Prevents crashes from fit failures and preserves raw data for later inspection.
-   **False Positive Detection**: Implements layered quality assessments to validate data and fit results, marking nodes as `failed` when a reliable fit is not found.
-   **Clear Outcome Reporting**: Provides detailed logs with specific reasons for failure.

**Affected Nodes:**

-   Power Rabi Analysis
-   Resonator Spectroscopy Analysis
-   Resonator Spectroscopy vs Amplitude Analysis
-   Resonator Spectroscopy vs Flux Analysis

A detailed breakdown of the new logic, including flowcharts and outcome definitions for each node, can be found in `analysis.md`.

### Standardized and Interactive Plotting Framework

A new framework for generating plots has been implemented to reduce code duplication and enhance maintainability. The central feature is the introduction of interactive plotting via Plotly.

**New Architecture:**

The framework decouples plot generation into three distinct components:

1.  **`PlotConfig` Model**: A declarative, backend-agnostic configuration for a plot's appearance.
2.  **Experiment-Specific Preparators**: Functions dedicated to transforming raw analysis data into a plot-ready format.
3.  **Generic Rendering Engine**: A dual-backend plotter that generates both interactive **Plotly** figures and static **Matplotlib** figures from a single configuration.

This new architecture simplifies the process of adding or modifying plots and ensures a consistent look and feel across the project. The design philosophy and an implementation guide are available in `plotting_strategy.md` and an example implementation is provided for node 02a.

### DIFF
--- a/qualibration_graphs/superconducting/calibration_utils/power_rabi/analysis.py
+++ b/qualibration_graphs/superconducting/calibration_utils/power_rabi/analysis.py
@@ -6,12 +6,15 @@ from typing import Dict, Optional, Tuple, Union
 
 import numpy as np
 import xarray as xr
-from qualibration_libs.analysis import fit_oscillation
+from qualibration_libs.analysis.helpers.power_rabi import (
+    assess_parameter_validity, calculate_multi_pulse_parameters,
+    calculate_signal_quality_metrics, calculate_single_pulse_parameters,
+    create_fit_results_dictionary, determine_fit_outcomes,
+    evaluate_amplitude_constraints, fit_multi_pulse_experiment,
+    fit_single_pulse_experiment)
 from qualibration_libs.analysis.parameters import analysis_config_manager
-from qualibration_libs.analysis.signal import compute_signal_to_noise_ratio
 from qualibration_libs.data import convert_IQ_to_V
 from quam_config.instrument_limits import instrument_limits
-from scipy.signal import correlate
 
 from qualibrate import QualibrationNode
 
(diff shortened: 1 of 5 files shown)


qualibration_graphs/superconducting/calibration_utils/resonator_spectroscopy/analysis.py
```diff
@@ -1,105 +1,413 @@
...
+    # Calculate resonator frequencies and FWHM
+    fit = _calculate_resonator_parameters(fit, node)
+    
+    # Evaluate fit quality for each qubit and determine outcomes
+    fit_results = {}
+    outcomes = []
+    
+    for qubit_name in fit.qubit.values:
+        # Extract fit metrics for this qubit
+        fit_metrics = _extract_qubit_fit_metrics(fit, qubit_name)
```
TheoLaudatQM: The fit_metrics dict contains very valuable information which should probably be added to `fit_results`, so that users can access it for debugging purposes.

qualibration_graphs/superconducting/calibration_utils/resonator_spectroscopy/analysis.py
```diff
@@ -1,105 +1,413 @@
...
+        True if peak shape is distorted
+    """
+    asymmetry_bad = (asymmetry is not None and 
+                    (asymmetry < asymmetry_min or asymmetry > asymmetry_max))
+    skewness_bad = (skewness is not None and abs(skewness) > skewness_max)
+    
+    return asymmetry_bad or skewness_bad
+
+
+def _is_peak_too_wide(
```
TheoLaudatQM: Why isn't this function part of qualibration-libs in analysis/feature_detection.py?

qualibration_graphs/superconducting/calibration_utils/resonator_spectroscopy/analysis.py
```diff
@@ -1,105 +1,413 @@
...
+                        fwhm_absolute_threshold):
+        if snr < snr_distorted:
+            return "The SNR isn't large enough and the peak shape is distorted"
+        else:
+            return "Distorted peak detected"
+    
+    return "successful"
+
+
+def _is_peak_shape_distorted(
```
TheoLaudatQM: Why isn't this function part of qualibration-libs in analysis/feature_detection.py?

qualibration_graphs/superconducting/calibration_utils/resonator_spectroscopy/analysis.py
```diff
@@ -1,105 +1,413 @@
...
+    skewness = metrics["skewness"]
+    opx_bandwidth_artifact = metrics["opx_bandwidth_artifact"]
+    
+    # Check SNR first
+    if snr < snr_min:
+        return "The SNR isn't large enough, consider increasing the number of shots"
+    
+    # Check for OPX bandwidth artifacts
+    if opx_bandwidth_artifact:
+        return "OPX bandwidth artifact detected, check your experiment bandwidth settings"
```
TheoLaudatQM:
```suggestion
        return "OPX bandwidth artifact detected, consider shifting the LO frequency by ~20MHz"
```

qualibration_graphs/superconducting/calibration_utils/resonator_spectroscopy/analysis.py
```diff
@@ -1,105 +1,413 @@
...
+        return "The peak shape is distorted"
+    
+    # Check for peak width issues
+    if _is_peak_too_wide(fwhm, sweep_span, snr, snr_distorted, 
+                        distorted_fraction_low_snr, distorted_fraction_high_snr, 
+                        fwhm_absolute_threshold):
+        if snr < snr_distorted:
+            return "The SNR isn't large enough and the peak shape is distorted"
+        else:
+            return "Distorted peak detected"
```
TheoLaudatQM: What is the purpose of this check?
I mean, why is it necessary to make the analysis fail if the peak is wide or a bit distorted?
For instance, qC3 and qC2 from dataset #544 are failing because of it, even though the frequency and FWHM are correctly estimated.

qualibration_graphs/superconducting/calibration_utils/resonator_spectroscopy/plot_configs.py
```diff
@@ -0,0 +1,58 @@
+from qualibration_libs.plotting.standard_plotter import (LayoutConfig,
```
TheoLaudatQM: Can you please explain what this is doing, especially since this framework seems to only be used for resonator_spectroscopy and not the other nodes?

qualibration_graphs/superconducting/calibration_utils/resonator_spectroscopy/preparators.py
```diff
@@ -0,0 +1,67 @@
+from typing import Optional
+
+import numpy as np
+import xarray as xr
+from qualibration_libs.analysis import lorentzian_dip
+
+
+def prepare_resonator_spectroscopy_data(
```
TheoLaudatQM: It looks like this is there in case there are inconsistencies between the plotting section and the analysis one, shouldn't this be handled by the analysis section directly?
Also this framework isn't implemented for the other nodes, is it intended?

qualibration_graphs/superconducting/calibrations/02a_resonator_spectroscopy.py
```diff
@@ -182,24 +181,57 @@ def analyse_data(node: QualibrationNode[Parameters, Quam]):
...
+        ds_fit=ds_fit,
+    )
+    plotly_amplitude.show()
+    plotly_phase.show()
+
+    # 3. Generate static matplotlib figures
+    static_phase_fig = create_matplotlib_figure(
+        ds_raw=ds_raw,
+        qubits=node.namespace["qubits"],
+        plot_configs=[phase_vs_freq_config],
```
TheoLaudatQM: I didn't check this carefully, but the second x-axis on the figures (detuning) disappeared, and I am suspecting that this is due to these new plot_configs

qualibration_graphs/superconducting/calibration_utils/resonator_spectroscopy/plotting.py
```diff
@@ -1,16 +1,32 @@
...
 
+# --- Constants ---
+MATPLOTLIB_FIG_WIDTH = 15
+MATPLOTLIB_FIG_HEIGHT = 9
+PLOTLY_FIG_HEIGHT = 900
+PLOTLY_FIG_WIDTH = 1500
+RAW_DATA_COLOR = "#1f77b4"
+FIT_COLOR = "#FF0000"  # Red
+FIT_LINE_STYLE = "dash"
+PLOTLY_AXIS_OFFSET = 100
```
TheoLaudatQM: Why not having a general dataclass in qualibration-libs/plotting to handle these parameters and have them uniform across all nodes?

qualibration_graphs/superconducting/calibration_utils/resonator_spectroscopy/analysis.py
```diff
@@ -1,105 +1,428 @@
...
+SNR_MIN: float = 2.5
+SNR_DISTORTED: float = 5.0
+ASYMMETRY_MIN: float = 0.4
+ASYMMETRY_MAX: float = 2.5
+SKEWNESS_MAX: float = 1.5
+DISTORTED_FRACTION_LOW_SNR: float = 0.2
+DISTORTED_FRACTION_HIGH_SNR: float = 0.3
+FWHM_ABSOLUTE_THRESHOLD_HZ: float = 1e6
+NRMSE_THRESHOLD: float = 0.14
+R_SQUARED_THRESHOLD: float = 0.90
```
TheoLaudatQM: How users are supposed to tweak these parameters?
What do they represent and how they connect to the analysis?

qualibration_graphs/superconducting/calibration_utils/resonator_spectroscopy_vs_amplitude/analysis.py
```diff
@@ -1,150 +1,882 @@
...
+GS_DETECTION_MIN_COUNT: int = 5
+GS_DETECTION_WINDOW: int = 2
+OPTIMAL_POWER_N_CLUSTERS: int = 2
+OPTIMAL_POWER_MARGIN_DB: float = 3.0
+SAVGOL_BASELINE_WINDOW: int = 51
+SAVGOL_SMOOTH_WINDOW: int = 21
+SAVGOL_POLYORDER: int = 3
+PEAK_PROMINENCE_STD_FACTOR: float = 3.0
+PEAK_DISTANCE: int = 10
+PEAK_WIDTH: int = 3
```
TheoLaudatQM: How users are supposed to tweak these parameters?
What do they represent and how they connect to the analysis?

qualibration_graphs/superconducting/calibration_utils/resonator_spectroscopy_vs_amplitude/analysis.py
```diff
@@ -1,150 +1,882 @@
...
-        s_power = f"Optimal readout power: {fit_results[q]['optimal_power']:.2f} dBm | "
-        s_freq = f"Resonator frequency: {1e-9 * fit_results[q]['resonator_frequency']:.3f} GHz | "
-        s_shift = f"(shift of {1e-6 * fit_results[q]['frequency_shift']:.0f} MHz)\n"
-        if fit_results[q]["success"]:
-            s_qubit += " SUCCESS!\n"
+    
+    for qubit_name, results in fit_results.items():
+        status_line = f"Results for qubit {qubit_name}: "
+        
+        # Handle both FitParameters objects and dictionaries
```
TheoLaudatQM, line 52: Why is this needed?
I am asking because vlaues are being printed even when the analysis is marked as failed.

qualibration_graphs/superconducting/calibration_utils/resonator_spectroscopy_vs_amplitude/analysis.py
```diff
@@ -1,150 +1,882 @@
...
+    freq_shifts = _calculate_frequency_shifts_at_optimal_power(ds_fit, node)
+    ds_fit = ds_fit.assign_coords({"freq_shift": (["qubit"], freq_shifts)})
+    
+    # Extract and validate fit parameters
+    fit_dataset, fit_results = _extract_and_validate_amplitude_parameters(ds_fit, node, ds)
+    
+    return fit_dataset, fit_results
+
+
+def _track_resonator_response_across_power(ds: xr.Dataset, qubits: list) -> Dict[str, np.ndarray]:
```
TheoLaudatQM: The analysis here is pretty complicated and I am wondering if it wouldn't make sense to transfer all these private functions to the qualibration-libs repo so that customers can just import what is necessary to fit the data, What do you think?
Are users supposed to manually tweak some parameters here?

qualibration_graphs/superconducting/calibration_utils/resonator_spectroscopy_vs_amplitude/analysis.py
```diff
@@ -1,150 +1,882 @@
...
     ds = ds.assign({"IQ_abs_norm": ds["IQ_abs"] / ds.IQ_abs.mean(dim=["detuning"])})
+    
     return ds
 
 
-def fit_raw_data(ds: xr.Dataset, node: QualibrationNode) -> Tuple[xr.Dataset, dict[str, FitParameters]]:
+def fit_raw_data(ds: xr.Dataset, node: QualibrationNode) -> Tuple[xr.Dataset, Dict[str, FitParameters]]:
     """
-    Fit the T1 relaxation time for each qubit according to ``a * np.exp(t * decay) + offset``.
+    Fit resonator spectroscopy vs amplitude data for each qubit in the dataset.
```
TheoLaudatQM, line 113: Could you maybe explain the analysis method here? How is the optimal power calculated?
I am asking to understand the method and why the optimal power is different between qubits. For instance on dataset #320 or #318, the distance between the optimal power and the non linearity is different for the three qubits and I don't really understand why.
Similarly, on dataset #669, why is the analysis converging on qC4 and not the others?

qualibration_graphs/superconducting/calibration_utils/resonator_spectroscopy_vs_amplitude/analysis.py
```diff
@@ -1,150 +1,882 @@
...
+    Returns
+    -------
+    Dict[str, np.ndarray]
+        Dictionary containing tracking results for each qubit
+    """
+    tracking_results = {
+        "rr_tracked": [],
+        "rr_tracked_filtered": [], 
+        "fit_quality_masks": []
+    }
```
TheoLaudatQM: I think that explaining the meaning of these keys would be useful for debugging purposes

qualibration_graphs/superconducting/calibration_utils/resonator_spectroscopy_vs_amplitude/analysis.py
```diff
@@ -1,150 +1,882 @@
...
+    
+    # Smooth the filtered centers if we have enough good points
+    smoothed_centers = _smooth_resonance_centers(filtered_centers, fit_mask)
+    
+    return {
+        "raw_centers": np.array(centers),
+        "filtered_centers": np.array(filtered_centers),
+        "smoothed_centers": smoothed_centers,
+        "fit_mask": np.array(fit_mask)
+    }
```
TheoLaudatQM: I think that explaining the meaning of these keys would be useful for debugging purposes

qualibration_graphs/superconducting/calibration_utils/resonator_spectroscopy_vs_flux/analysis.py
```diff
@@ -1,161 +1,632 @@
...
+AMP_REL_THRESHOLD: float = 0.01
+SNR_MIN: float = 2.0
+RESONATOR_TRACE_SMOOTH_SIGMA: float = 1.5
+RESONATOR_TRACE_DIP_THRESHOLD: float = 0.01
+RESONATOR_TRACE_GRADIENT_THRESHOLD: float = 0.001
+MIN_FLUX_MODULATION_HZ: float = 1e6
+FLUX_MODULATION_SMOOTH_SIGMA: float = 1.5
+RESONANCE_CORRECTION_CONFIDENCE_THRESHOLD: float = -0.6
+RESONANCE_CORRECTION_WINDOW: int = 5
+EPSILON: float = 1e-9
```
TheoLaudatQM: How users are supposed to tweak these parameters?
What do they represent and how they connect to the analysis?

qualibration_graphs/superconducting/calibration_utils/resonator_spectroscopy_vs_flux/analysis.py
```diff
@@ -1,161 +1,632 @@
...
+        freq_str = f"Resonator frequency: {resonator_frequency * 1e-9:.3f} GHz | "
+        shift_str = f"(shift of {frequency_shift * 1e-6:.0f} MHz)\n"
+        
+        log_callable(status_line + idle_str + min_str + freq_str + shift_str)
 
-def process_raw_dataset(ds: xr.Dataset, node: QualibrationNode):
-    """Processes the raw dataset by converting the 'I' and 'Q' quadratures to V, or adding the RF_frequency as a coordinate for instance."""
 
-    # Convert the 'I' and 'Q' quadratures from demodulation units to V.
+def process_raw_dataset(ds: xr.Dataset, node: QualibrationNode) -> xr.Dataset:
```
TheoLaudatQM, line 82: The analysis seems to work well on the provided dataset, except for #681, where the min offset of qC3 doesn't right, do you know how come?

qualibration_graphs/superconducting/calibration_utils/power_rabi/analysis.py
```diff
@@ -1,148 +1,681 @@
...
+from qualibrate import QualibrationNode
+
+# Analysis Constants
+MIN_FIT_QUALITY: float = 0.5
+MIN_AMPLITUDE: float = 0.01
+MAX_AMP_PREFACTOR: float = 10.0
+SNR_MIN: float = 2.5
+AUTOCORRELATION_R1_THRESHOLD: float = 0.8
+AUTOCORRELATION_R2_THRESHOLD: float = 0.5
+CHEVRON_MODULATION_THRESHOLD: float = 0.4
```
nulinspiratie: We need to think about how we store these constants. I'm not sure if the best approach is to have them as file-level variables. 

Some alternatives would be to have a global config or to encapsulate everything as arguments in the functions and classes. I'm not sure about the best approach, we should discuss

qualibration_graphs/superconducting/calibration_utils/power_rabi/analysis.py
```diff
@@ -1,148 +1,681 @@
...
-    fit_results : dict
+    Parameters
+    ----------
+    fit_results : Dict[str, FitParameters]
         Dictionary containing the fitted results for all qubits.
-    logger : logging.Logger, optional
+    log_callable : callable, optional
         Logger for logging the fitted results. If None, a default logger is used.
-
     """
```
nulinspiratie: Let's stick to Google style docstrings.

nulinspiratie: We should avoid adding parameter types in the docstring since they are already captured by the function signature.

qualibration_graphs/superconducting/calibration_utils/power_rabi/analysis.py
```diff
@@ -1,148 +1,681 @@
...
+            operation = results.operation
+            fit_quality = results.fit_quality
         else:
-            s_qubit += " FAIL!\n"
-        log_callable(s_qubit + s_amp)
+            outcome = results.get('outcome', 'unknown')
+            opt_amp = results.get('opt_amp', 0.0)
+            opt_amp_prefactor = results.get('opt_amp_prefactor', 1.0)
+            operation = results.get('operation', 'x180')
+            fit_quality = results.get('fit_quality', None)
```
nulinspiratie: Why are we handling both fit parameters objects and dictionaries? Ideally, we should use just one class to reduce code complexity and duplication. I suggest we have a standardized class, such as FitParameters, that can manage this.

qualibration_graphs/superconducting/calibration_utils/power_rabi/analysis.py
```diff
@@ -1,148 +1,681 @@
...
+MIN_AMPLITUDE: float = 0.01
+MAX_AMP_PREFACTOR: float = 10.0
+SNR_MIN: float = 2.5
+AUTOCORRELATION_R1_THRESHOLD: float = 0.8
+AUTOCORRELATION_R2_THRESHOLD: float = 0.5
+CHEVRON_MODULATION_THRESHOLD: float = 0.4
 
 
 @dataclass
 class FitParameters:
```
nulinspiratie, line 26: Given that we expect to use this throughout, it might make sense to have a base class for it. I'm even wondering whether that base class should belong in QUA-Libs or if it actually belongs in QUAlibration-Libs.

qualibration_graphs/superconducting/calibration_utils/power_rabi/analysis.py
```diff
@@ -1,148 +1,681 @@
...
+    """Calculate signal-to-noise ratio for each qubit."""
+    snrs = []
+    for q in fit.qubit.values:
+        # Choose appropriate signal data
+        if hasattr(fit, "I") and hasattr(fit.I, "sel"):
+            signal = fit.I.sel(qubit=q).values
+        elif hasattr(fit, "state") and hasattr(fit.state, "sel"):
+            signal = fit.state.sel(qubit=q).values
+        else:
+            signal = np.zeros(2)
```
nulinspiratie: In which cases would we reach this? Can fit not contain either? In this case, would a warning be appropriate?

qualibration_graphs/superconducting/calibration_utils/power_rabi/plotting.py
```diff
@@ -1,16 +1,37 @@
...
+MATPLOTLIB_FIG_WIDTH = 15
+MATPLOTLIB_FIG_HEIGHT = 9
+PLOTLY_FIG_HEIGHT = 900
+PLOTLY_FIG_WIDTH = 1500
+PLOTLY_COLORSACLE = "Viridis"
+# For overlaying axes in plotly
+PLOTLY_AXIS_OFFSET = 100
+
+
 def plot_raw_data_with_fit(ds: xr.Dataset, qubits: List[AnyTransmon], fits: xr.Dataset):
```
nulinspiratie: I vote we avoid having two plotting routines (matplotlib & plotly) as it's more code to maintain

qualibration_graphs/superconducting/calibration_utils/power_rabi/plotting.py
```diff
@@ -133,9 +160,214 @@ def plot_individual_data_with_fit_2D(ax: Axes, ds: xr.Dataset, qubit: dict[str,
...
         ax2.axvline(
             x=fit.opt_amp_prefactor,
-            color="g",
+            color=FIT_COLOR,
             linestyle="-",
+            linewidth=FIT_LINE_WIDTH,
+        )
+
+
+def plotly_plot_raw_data_with_fit(ds: xr.Dataset, qubits: List[AnyTransmon], fits: xr.Dataset) -> go.Figure:
```
nulinspiratie: This function seems very generic, why not move it to qualibration-libs?

qualibration_graphs/superconducting/calibration_utils/power_rabi/analysis.py
```diff
@@ -1,148 +1,681 @@
...
+        
+    except (AttributeError, KeyError, ValueError):
+        pass
+    
+    return ""
+
+
+# Main Outcome Determination Function
+
+def _determine_qubit_outcome(
```
nulinspiratie: The return type is a bit odd here. It seems there are two pieces of information: whether it is a success or not, and additional details. For example, if we currently want to say "it was a success but the fit had more noise than ideal", that cannot be captured properly. I think it makes more sense to separate this into a dict with outcome, and details

qualibration_graphs/superconducting/calibration_utils/resonator_spectroscopy_vs_flux/plotting.py
```diff
@@ -1,15 +1,56 @@
...
+PLOTLY_COLORBAR_THICKNESS = 14
+PLOTLY_FIT_LINE_WIDTH = 2.5
+PLOTLY_FIT_LINE_STYLE = "dash"
+PLOTLY_SWEET_SPOT_MARKER_SYMBOL = "x"
+PLOTLY_SWEET_SPOT_MARKER_SIZE = 15
+PLOTLY_ANNOTATION_FONT_SIZE = 16
+PLOTLY_SUBPLOT_WIDTH = 400
+PLOTLY_SUBPLOT_HEIGHT = 400
+PLOTLY_MIN_WIDTH = 1000
+PLOTLY_MARGIN = {"l": 60, "r": 60, "t": 80, "b": 60}
```
nulinspiratie: The amount of parameters should be reduced as much as possible per plotting file. Common parameters (marker symbol, colorscale etc) should be moved away into a common config

qualibration_graphs/superconducting/calibration_utils/resonator_spectroscopy_vs_flux/plotting.py
```diff
@@ -79,30 +120,555 @@ def plot_individual_raw_data_with_fit(ax: Axes, ds: xr.Dataset, qubit: dict[str,
...
+            marker=SWEET_SPOT_MARKER,
+            color=SWEET_SPOT_COLOR,
+            markersize=SWEET_SPOT_MARKERSIZE,
+            linestyle="None",
         )
     ax.set_title(qubit["qubit"])
     ax.set_xlabel("Flux (V)")
+
+
+def plotly_plot_raw_data_with_fit(
```
nulinspiratie: It's worth rethinking these plotting functions. Currently it's not reusing any common functions, and as a result, the functions are very long with significant code duplication. Ideally all the common plotting functions should move to qualibration-libs, and this calls those functions.
Many properties (e.g. hovertemplate) should also not be hardcoded IMHO, but there should be a generalized function that can extract it

qualibration_graphs/superconducting/calibration_utils/resonator_spectroscopy_vs_amplitude/analysis.py
```diff
@@ -1,150 +1,882 @@
...
+            else:
+                # Fallback: try to extract any numeric characters
+                numeric_part = ''.join(c for c in str(p) if c.isdigit() or c in '.-')
+                if numeric_part and numeric_part != '.' and numeric_part != '-':
+                    numeric_powers.append(float(numeric_part))
+                else:
+                    # Last resort: use index
+                    numeric_powers.append(float(len(numeric_powers)))
+        return np.array(numeric_powers)
+    except (ValueError, TypeError, ImportError):
```
nulinspiratie: should avoid imports in a function. Also why are we adding an importerror exception? re is part of the python core packages

qualibration_graphs/superconducting/calibration_utils/resonator_spectroscopy_vs_amplitude/analysis.py
```diff
@@ -1,150 +1,882 @@
...
+            window_length = polyorder + 1 if (polyorder + 1) % 2 != 0 else polyorder + 2
+            if window_length > data_length:
+                return -1  # Not possible to find a valid window
+        else:
+            return -1  # Not possible to find a valid window
+    
+    return window_length
+
+
+def _determine_amplitude_outcome(
```
nulinspiratie: Also here I propose we avoid using strings to capture all the return information. The main reason is that it becomes tricky to discriminate. For example, how would we discriminate "successful, but here are some additional details"?

qualibration_graphs/superconducting/calibration_utils/resonator_spectroscopy_vs_amplitude/analysis.py
```diff
@@ -1,150 +1,882 @@
...
+    optimal_power = float(fit.optimal_power.sel(qubit=qubit_name).data)
+    snr = fit.snr.sel(qubit=qubit_name).data if "snr" in fit.data_vars else None
+    num_lines = fit.num_lines.sel(qubit=qubit_name).data if "num_lines" in fit.data_vars else 1
+    low_power_center_std = fit.low_power_center_std.sel(qubit=qubit_name).data if "low_power_center_std" in fit.data_vars else None
+    
+    # Get experiment parameters
+    min_power_dbm = getattr(node.parameters, "min_power_dbm", None)
+    max_power_dbm = getattr(node.parameters, "max_power_dbm", None)
+    power_values = fit.power.values if "power" in fit.dims or "power" in fit.coords else None
+
```
nulinspiratie: We should be more assertive in these statements. If the node contains the parameter `min_power_dbm`, then we can assume that it exists here, rather than use getattr with a default.
I would say the same for `fit`, unless there's a clear reason why `power` wouldn't exist

qualibration_graphs/superconducting/calibration_utils/power_rabi/analysis.py
```diff
@@ -1,148 +1,681 @@
...
+    Returns
+    -------
+    xr.Dataset
+        Dataset with oscillation fit results
+    """
+    ds_fit = ds.sel(nb_of_pulses=1)
+    
+    # Choose signal based on discrimination method
+    signal_data = ds_fit.state if node.parameters.use_state_discrimination else ds_fit.I
+    fit_vals = fit_oscillation(signal_data, "amp_prefactor")
```
TheoLaudatQM: This isn't working for dataset #251 as you can see on qD3 and qC1.
A better initial guess for the oscillation frequency has been investigated in [this PR](https://github.com/qua-platform/qualibration-libs/pull/8/files).

qualibration_graphs/superconducting/calibration_utils/power_rabi/analysis.py
```diff
@@ -1,148 +1,681 @@
...
+    # Check for missing structure in 2D datasets
+    if not has_structure and not is_1d_dataset:
+        return "No chevron modulation detected. Please check the drive frequency and amplitude sweep range"
+
+    # Check signal-to-noise ratio
+    if snr is not None and snr < snr_min:
+        return f"SNR too low (SNR = {snr:.2f} < {snr_min})"
+
+    # Check amplitude hardware limits
+    if not amp_success:
```
TheoLaudatQM: How does this relate to a large detuning?

TheoLaudatQM (commented):
Thanks for having put so much effort into this, however reviewing these 5000 lines of code at onces was tedious and having started with a smaller PR on one node only as suggested in the kick-off meeting would have probably eased this part and allows for clarifying the direction and some design choices.
Here are a few general comments:

1. The complexity of the analysis and plotting parts increases tremendously. It isn't a bad thing in itself, but it basically makes it impossible for a lambda user to debug the analysis and understand what is happening there.
2. There seem to have many free parameters, supposedly for the user to define and adjust to make the fits converge, but it is quite difficult to understand the meaning of these and how the user is supposed to tweak them, if he supposed to.
3. Considering the complexity, defining base classes for the analysis and plotting section seem mandatory to avoid code duplication and to avoid exposing the complexity to the user. This would allow to abstract away the complicated analysis logic and allow the user to easily replace it by his own fitting routines if needed.
4. The analysis on some of the provided datasets aren't converging, and it is unclear why.
5. I also understood that there are two strategies for the plotting section, one that is abstracted away using plot_configs.py and preparators.py and one that is more not standardized using individual plotting.py files, is it correct?

Let's discuss these point during the review meeting

qualibration_graphs/superconducting/calibration_utils/resonator_spectroscopy/analysis.py
shanto268: It has been addressed in the latest PR and added to `FitParameters`

qualibration_graphs/superconducting/calibration_utils/resonator_spectroscopy/analysis.py
shanto268: I thought it was best to have node specific small helper analysis methods like this in their file but I can of course move it to `analysis/feature_detection.py`

qualibration_graphs/superconducting/calibration_utils/resonator_spectroscopy/analysis.py
shanto268: I thought it was best to have node specific small helper analysis methods like this in their file but I can of course move it to `analysis/feature_detection.py`

qualibration_graphs/superconducting/calibration_utils/resonator_spectroscopy/analysis.py
shanto268: This has been addressed post our discussions from last week but the idea was to keep a check on such geometric flags because that can inform the users about potential physics issues e.g. power broadening, thermal effects, etc

qualibration_graphs/superconducting/calibration_utils/resonator_spectroscopy/plot_configs.py
shanto268: As discussed during our meeting last week, this was an implementation example as per the application note

qualibration_graphs/superconducting/calibration_utils/resonator_spectroscopy/preparators.py
shanto268: As discussed during our meeting last week, this was done intentionally since this was an implementation example following the idea explained in the application note.

The new PR simplifies this a bit more and has a slightly different framework and has been implemented for all the nodes.

qualibration_graphs/superconducting/calibrations/02a_resonator_spectroscopy.py
shanto268: unfortunately as dumb as it sounds but `plotly` does not allow for this feature trivially. Ref [https://github.com/plotly/plotly.py/issues/3948](https://github.com/plotly/plotly.py/issues/3948). I did add the detuning information in the hover over labels and if the detuning axes info is really needed then we can either try to solve it by including a toggle button to change the x-axes from RF frequency to detuning.

qualibration_graphs/superconducting/calibration_utils/resonator_spectroscopy/plotting.py
shanto268: Has been implemented in the latest PR in `qualibration-libs/plotting/config.py`

qualibration_graphs/superconducting/calibration_utils/resonator_spectroscopy/analysis.py
shanto268: As per our discussion last week, we have implemented a pydantic BaseModel (`qualibration-libs/plotting/parameters.py`) for storage, reference and ease of change for analysis needs.

qualibration_graphs/superconducting/calibration_utils/resonator_spectroscopy_vs_amplitude/analysis.py
shanto268: As per our discussion last week, we have implemented a pydantic BaseModel (`qualibration-libs/analysis/parameters.py`) for storage, reference and ease of change for analysis needs.

qualibration_graphs/superconducting/calibration_utils/resonator_spectroscopy_vs_amplitude/analysis.py
shanto268: It was intentionally kept with the idea that this information will still be useful to the user for any debugging reasons.

qualibration_graphs/superconducting/calibration_utils/resonator_spectroscopy_vs_amplitude/analysis.py
shanto268: As discussed in the previous meeting, the parameters are now stable to cover all the test datasets but of course for edge cases some tweaking would be needed which now can be done with the pydantic model introduced in the new PR.

There is no strong reason to keep these methods in the node specific analysis files except for that it can be easily accessed. We can move these to qualibration-libs repo.

qualibration_graphs/superconducting/calibration_utils/resonator_spectroscopy_vs_amplitude/analysis.py
shanto268: The optimal power is chosen by analyzing how the resonator dip depth changes with power. For each qubit, we collapse the 2D sweep to a 1D curve by taking the minimum |IQ| across detuning at each power. We then apply KMeans clustering to this curve, and identify the cluster with the lowest average dip as the ground-state region. From that cluster, we select the median power as a robust point that avoids edge effects, and subtract a fixed safety margin (`optimal_power_margin_db`) to stay safely below the onset of non-linearity.

The reason why the optimal power looks different is because it only takes into account the ground-state cluster which in #318 for example for qC3 and qC2 has many more datapoints than qC1 but in all cases the median value is chosen. Also, if clustering fails for a given qubit (e.g. due to low SNR), the code falls back to the midpoint of the sweep, which can create additional variation - which is not the case for any for #320 or #318.

Could you rerun on #669 it should really converge on any of the qubits in #669 given the obvious SNR issues.

qualibration_graphs/superconducting/calibration_utils/resonator_spectroscopy_vs_amplitude/analysis.py
shanto268: woops missed this but will add to the doc string.

```
        - "rr_tracked": Smoothed and interpolated resonance centers. Provides a clean trace.
        - "rr_tracked_filtered": Resonance centers with poor quality fits replaced by NaN.
        - "fit_quality_masks": Boolean mask indicating if fit for each power level was good (True).
```

qualibration_graphs/superconducting/calibration_utils/resonator_spectroscopy_vs_amplitude/analysis.py
shanto268: added in the last PR

qualibration_graphs/superconducting/calibration_utils/resonator_spectroscopy_vs_flux/analysis.py
shanto268: As per our discussion last week, we have implemented a pydantic BaseModel (`qualibration-libs/analysis/parameters.py`) for storage, reference and ease of change for analysis needs.

qualibration_graphs/superconducting/calibration_utils/resonator_spectroscopy_vs_flux/analysis.py
shanto268: I have addressed this in the latest PR

qualibration_graphs/superconducting/calibration_utils/power_rabi/analysis.py
shanto268: I have implemented the code from that PR and now it works on 251

qualibration_graphs/superconducting/calibration_utils/power_rabi/analysis.py
shanto268: In setting that variable, I check for drive frequency and qubit frequency thats how I keep track of detuning. You are right it is a bit misleading since there is no obvious connection for determining a large detuning. The latest PR with the new osciallation fit makes this a mute point

qualibration_graphs/superconducting/calibration_utils/resonator_spectroscopy_vs_amplitude/analysis.py
shanto268: addressed in the latest PR

qualibration_graphs/superconducting/calibration_utils/resonator_spectroscopy_vs_amplitude/analysis.py
shanto268: I agree that dict would be the best choice but in the SOW document provided it was mentioned to keep outcome as a str and so I respected that choice

qualibration_graphs/superconducting/calibration_utils/resonator_spectroscopy_vs_amplitude/analysis.py
shanto268: addressed in the latest PR

qualibration_graphs/superconducting/calibration_utils/resonator_spectroscopy_vs_flux/plotting.py
shanto268: yes you are totally right. 

I think the confusion stems from the fact that I didn't upload the application note to the github (it was shared over slack). I only implemented the "proper" structure to node02a and left the other 3 nodes in a mostly backwards compatible flow but still kept those plotly methods that I wrote while developing.

nevertheless, the latest PR addresses this issue

qualibration_graphs/superconducting/calibration_utils/resonator_spectroscopy_vs_flux/plotting.py
shanto268: addressed in the latest PR

qualibration_graphs/superconducting/calibration_utils/power_rabi/analysis.py
shanto268: I agree that dict would be the best choice but in the SOW document provided it was mentioned to keep outcome as a str and so I respected that choice

qualibration_graphs/superconducting/calibration_utils/power_rabi/plotting.py
shanto268: The latest PR removes all of this.

The reason I had this was just as a quick way to get these plotly plots as I was working on this codebase

qualibration_graphs/superconducting/calibration_utils/power_rabi/plotting.py
shanto268: You are absolutely right about the added complexity. I have kept the matplotlib engine for backwards compatibility reasons and also because the `qualibrate` webapp renders the static images nicely

qualibration_graphs/superconducting/calibration_utils/power_rabi/analysis.py
shanto268: great catch. I have added a warning in the latest PR.

qualibration_graphs/superconducting/calibration_utils/power_rabi/analysis.py
shanto268: Addressed in the latest PR

qualibration_graphs/superconducting/calibration_utils/power_rabi/analysis.py
shanto268: Addressed in the latest PR

qualibration_graphs/superconducting/calibration_utils/power_rabi/analysis.py
shanto268: Noted. I have not addressed this comment yet but once all the more pressing things are ironed out - I will make sure all the documentation follows the guidelines y'all have :)

qualibration_graphs/superconducting/calibration_utils/power_rabi/analysis.py
shanto268: Hi @nulinspiratie, thanks a bunch for all your comments - they were very useful and helpful. We had a discussion regarding this issue and in the latest PR we have implemented a pydantic BaseModel (qualibration-libs/analysis/parameters.py) for storage, reference and ease of change for analysis needs.

qualibration_graphs/superconducting/calibration_utils/power_rabi/analysis.py
TheoLaudatQM: I think that Serwan is right and aligns very well with the example of dataset 86, where multiple peaks are detected, but the fit should still converge on the most prominent one with the indication to the user that multiple peaks were detected. If you don't think that it belongs to the SOW, then it's okay and we'll implement it ourselves.

qualibration_graphs/superconducting/calibration_utils/resonator_spectroscopy/analysis.py
TheoLaudatQM: I understand, but since the goal of this function is very close to the one of `_opx_bandwidth_artifact` (if I understand correctly how they work), I don't see why they don't belong in the same location. This also applies to `_is_peak_shape_distorted`.
The differentiation between what is in qua-libs and what is in qualibration-libs, and @nulinspiratie correct me if I am wrong, is the following:
- Everything that users will have to modify or debug themselves should belong in qua-libs.
- Everything that we provide to the users, but don't expect them to modify or debug, or can be re-used across multiple nodes, should belong in qualibration-libs.

Since these private functions are only used in this specific fitting routine that is, in my opinion, too complex for our users to dive into and debug, then they should be together.

More generally, I believe that the analysis.py file in qua-libs must contain:

- `log_fitted_results`: to expose relevant information to the user through the Terminal
- `process_raw_dataset`: to convert the data into V, get the amplitude and phase, redefine the axes...
- `fit_raw_data`: to perform the fitting routine, extract the fitted parameters and fit metrics
- `_extract_and_validate_fit_parameters`: to extract the fit parameters and outcome

as these functions provide data that is required in all the nodes.
My concern with the current implementation is that many  private functions are dependent on a specific analysis implementation and may confuse users when trying to debug or add their own methods, but this is something we can take on our side unless you have suggestions.

qualibration_graphs/superconducting/calibration_utils/resonator_spectroscopy/analysis.py
TheoLaudatQM: This I agree, but this check only makes sense to me if the user first measure the resonator response, and then define in `qc_params` which width is considered too large, how what distortion actually means. Is it how you intended the framework to be?

qualibration_graphs/superconducting/calibration_utils/resonator_spectroscopy/analysis.py
shanto268: Great points, Theo. I have addressed them in the latest PR

qualibration_graphs/superconducting/calibration_utils/resonator_spectroscopy/analysis.py
shanto268: One can certainly follow the workflow you explained - my reasoning for this check was to be conservative in our automated fits for the customers and to explain the features represented in the datasets provided

PR #328, opened 2025-07-02
load -> load_historical / init -> node init

KevinAVR:
In the qualibration nodes,

Initialization -> Node Initialization
load_data -> load_historical_data


nulinspiratie:
Agreed

PR #331, opened 2025-07-14
Pulsed ODMR and XY8 sequences for NV centers

NicolasStaudenmaier:
This PR implements the sequences of pulsed ODMR and XY8 for NV centers. Pulsed ODMR is a basic and widely used measurement to detect the spin resonance frequency. When calibrated well, it usually gives much better and precise results than cw ODMR. 
Furthermore, the dynamical decoupling sequence XY8-N of order N is added. The sequence can be performed in two different modes. One is sweeping the interpulse spacing (tau) between the pi-pulses and the other is sweeping the order N (number of repetitions) of the XY8 block. 
The `README.md` file is updated accordingly and some file names are changed to ensure consistent naming with the general structure in this repo. 

The XY8 tau sweep sequence shows no problems with unintentional gaps within the XY8 sequence during runtime. However, for the XY8 order sweep a gap may occur because a dynamic QUA `for_()` loop is required to cycle through the orders. I found that with the implementation in `09c_xy8_order.py` a single gap is evaluated before entering the loop. Importantly, the command structure as it is reduces the number of gaps to one. With other structures, gaps inside the loop were determined. 
The exact duration of the gap depends on the sequence parameters and I determined something between no gap and 10 clock cycles. It can be compensated by adding the gap time between the end of the XY8-block and the final pi/2-pulse. When calibrated properly, this operation does not affect the physics of the sequence. 

Proper implementation is verified with the waveform report in simulated mode on the OPX+ in the CS_1 rack.


Quantum-Control-Applications/Optically addressable spin qubits/NV center in a confocal setup/05b_pulsed_odmr.py
```diff
@@ -0,0 +1,145 @@
...
+
+#####################################
+#  Open Communication with the QOP  #
+#####################################
+qmm = QuantumMachinesManager(host=qop_ip, cluster_name=cluster_name, octave=octave_config)
+
+#######################
+# Simulate or execute #
+#######################
+simulate = True
```
TheoLaudatQM: I would set the default to False so that you don't have to always toggle it when running the script during an installation

Quantum-Control-Applications/Optically addressable spin qubits/NV center in a confocal setup/05b_pulsed_odmr.py
```diff
@@ -0,0 +1,145 @@
...
+counts received by the SPCM across varying the mw frequency.
+The sequence is repeated without playing the mw pulses to measure the dark counts on the SPCM.
+
+The data is then post-processed to determine the spin resonance frequency.
+This frequency can be used to update the NV intermediate frequency in the configuration under "NV_IF_freq".
+
+Prerequisites:
+    - Ensure calibration of the different delays in the system (calibrate_delays).
+    - Update the different delays in the configuration
+    - Setting the pi pulse amplitude and duration in the configuration
```
TheoLaudatQM: How do you get the pi-pulse amplitude before running power Rabi, or getting a Rabi chevron?

Quantum-Control-Applications/Optically addressable spin qubits/NV center in a confocal setup/09b_xy8_tau.py
```diff
@@ -0,0 +1,217 @@
...
+       XY8 MEASUREMENT (tau sweep)
+The program consists in playing two XY8 sequences successively (first ending with x90 and then with -x90)
+and measure the photon counts received by the SPCM across varying idle times between pi-pulses.
+Note that the pulse spacing tau is the interpulse spacing, i.e. the time between consecutive pi-pulses and
+does not take into account the finite length of the pi-pulses.
+
+The data is post-processed to determine the coherence time T2 associated with the XY8 tau sweep measurement.
+
+The sequence is defined in the following way:
+pix/2 - [t - pix - 2t - piy - 2t - pix - 2t - piy - 2t - piy - 2t - pix - 2t - piy - 2t - pix - t] ^ xy8_order - pix/2
```
TheoLaudatQM: Why not sticky with the operation names from the config here (pix/2 -> x90, piy -> y180...)?

Quantum-Control-Applications/Optically addressable spin qubits/NV center in a confocal setup/09b_xy8_tau.py
```diff
@@ -0,0 +1,217 @@
...
+    "n_avg": n_avg,
+    "t_vec": tau_vec,
+    "config": config,
+}
+
+
+###########################
+#  XY8 sequence wrappers  #
+###########################
+def xy8_n(order):
```
TheoLaudatQM: Shouldn't tau and tau_half be also input parameters of the macro for better readability?

Quantum-Control-Applications/Optically addressable spin qubits/NV center in a confocal setup/09b_xy8_tau.py
```diff
@@ -0,0 +1,217 @@
...
+    Assumes it starts frame at x, if not, need to reset_frame before"""
+    wait(tau_half, "NV")
+    xy8_block()
+    for _ in range(order - 1):
+        wait(tau, "NV")
+        xy8_block()
+    wait(tau_half, "NV")
+
+
+def xy8_block():  # A single XY8 block, ends at x frame.
```
TheoLaudatQM: Same here with tau?

Quantum-Control-Applications/Optically addressable spin qubits/NV center in a confocal setup/09b_xy8_tau.py
```diff
@@ -0,0 +1,217 @@
...
+
+
+###########################
+#  XY8 sequence wrappers  #
+###########################
+def xy8_n(order):
+    """Performs the full xy8_n sequence.
+    The first block is outside loop to avoid delays caused from either the
+    loop or from two consecutive wait commands.
+    Assumes it starts frame at x, if not, need to reset_frame before"""
```
TheoLaudatQM: What does this sentence mean in this context?
You don't seem to update the frame in the sequence, so why would `reset_frame()` do here?

Quantum-Control-Applications/Optically addressable spin qubits/NV center in a confocal setup/09b_xy8_tau.py
```diff
@@ -0,0 +1,217 @@
...
+
+    # Spin initialization
+    play("laser_ON", "AOM1")
+    wait(wait_for_initialization * u.ns, "AOM1")
+
+    # XY8-n sequence
+    with for_(n, 0, n < n_avg, n + 1):
+        with for_(*from_array(tau, tau_vec)):
+            assign(tau_half, tau / 2)
+            with strict_timing_():
```
TheoLaudatQM, line 140: I think that it is the first time that the `strict_timing` block appears in this set of scripts, so I would briefly explain what it does for new users. Something like "# Strict_timing ensures that the sequence will be played without gaps, if gaps are detected then an error will be raised"

Quantum-Control-Applications/Optically addressable spin qubits/NV center in a confocal setup/09b_xy8_tau.py
```diff
@@ -0,0 +1,217 @@
...
+    n_st = declare_stream()  # stream to save iterations
+
+    # Spin initialization
+    play("laser_ON", "AOM1")
+    wait(wait_for_initialization * u.ns, "AOM1")
+
+    # XY8-n sequence
+    with for_(n, 0, n < n_avg, n + 1):
+        with for_(*from_array(tau, tau_vec)):
+            assign(tau_half, tau / 2)
```
TheoLaudatQM: For best practice, I would use bit shifting instead of division here as the calculation can be done in 1 clock cycle instead of something like 60: `tau / 2 -> tau >> 1`

Quantum-Control-Applications/Optically addressable spin qubits/NV center in a confocal setup/09b_xy8_tau.py
```diff
@@ -0,0 +1,217 @@
...
+                xy8_n(xy8_order)
+                play("-x90", "NV")
+            align()  # Play the laser pulse after the Echo sequence
+            # Measure and detect the photons on SPCM1
+            play("laser_ON", "AOM1")
+            measure("readout", "SPCM1", None, time_tagging.analog(times2, meas_len_1, counts2))
+            save(counts2, counts_2_st)  # save counts
+            wait(wait_between_runs * u.ns, "AOM1")
+
+            wait(wait_between_runs * u.ns, "AOM1")
```
TheoLaudatQM, line 181: This line seems to be duplicated, no?

Quantum-Control-Applications/Optically addressable spin qubits/NV center in a confocal setup/09b_xy8_tau.py
```diff
@@ -0,0 +1,217 @@
...
+    interrupt_on_close(fig, job)  # Interrupts the job when closing the figure
+
+    while results.is_processing():
+        # Fetch results
+        counts1, counts2, iteration = results.fetch_all()
+        # Progress bar
+        progress_counter(iteration, n_avg, start_time=results.get_start_time())
+        # Plot data
+        plt.cla()
+        plt.plot(2 * tau_vec, counts1 / 1000 / (meas_len_1 / u.s), label="x90_XY8_x90")
```
TheoLaudatQM: In order to convert `tau` in ns you need to multiply `tau_vec` by 4, no?

Quantum-Control-Applications/Optically addressable spin qubits/NV center in a confocal setup/09b_xy8_tau.py
```diff
@@ -0,0 +1,217 @@
...
+        plt.ylabel("Contrast [kcps]")
+        plt.title("XY8-{} tau sweep".format(xy8_order))
+        plt.legend()
+        plt.pause(0.1)
+    # Save results
+    script_name = Path(__file__).name
+    data_handler = DataHandler(root_data_folder=save_dir)
+    save_data_dict.update({"counts1_data": counts1})
+    save_data_dict.update({"counts2_data": counts2})
+    save_data_dict.update({"fig_live": fig})
```
TheoLaudatQM, line 266: I would also save the processed data (contrast, counts in kcps) as it can be relevant if the user wants to re-generate the figures or work on the data analysis, what do you think?

Quantum-Control-Applications/Optically addressable spin qubits/NV center in a confocal setup/09b_xy8_tau.py
```diff
@@ -0,0 +1,217 @@
...
+from qualang_tools.loops import from_array
+from qualang_tools.results.data_handler import DataHandler
+
+import matplotlib.pyplot as plt
+
+
+##################
+#   Parameters   #
+##################
+tau_vec = 2 * np.arange(4, 500, 20)  # The time vector for varying the time between pulses in clock cycles (4ns)
```
TheoLaudatQM: I would specify that each tau must be a multiple of 8 ns (2 clock cycle) as tau/2 must be an integer multiple of 4ns (1 clock cycle)

Quantum-Control-Applications/Optically addressable spin qubits/NV center in a confocal setup/09c_xy8_order.py
```diff
@@ -0,0 +1,211 @@
...
+from qualang_tools.loops import from_array
+from qualang_tools.results.data_handler import DataHandler
+
+import matplotlib.pyplot as plt
+
+
+##################
+#   Parameters   #
+##################
+tau = 2 * 50  # The time between pi-pulses in clock cycles (4ns)
```
TheoLaudatQM: I would specify that tau must be a multiple of 8 ns (2 clock cycle) as tau/2 must be an integer multiple of 4ns (1 clock cycle)

Quantum-Control-Applications/Optically addressable spin qubits/NV center in a confocal setup/09c_xy8_order.py
```diff
@@ -0,0 +1,211 @@
...
+order_vec = np.arange(1, 21, 1, dtype=int)  # order vector for varying the order n of the XY8-n measurement
+n_avg = 1_000_000
+t_gap = 10  # possible timing gap in clock cycles that is to be compensated
+tau_half = tau // 2  # time between pi/2-pulse and XY8 block
+
+# Data to save
+save_data_dict = {
```
TheoLaudatQM, line 64: I think that saving tau could be useful if one wants to analyze the data later on

Quantum-Control-Applications/Optically addressable spin qubits/NV center in a confocal setup/09b_xy8_tau.py
```diff
@@ -0,0 +1,217 @@
...
+xy8_order = 4  # order n of the XY8-n measurement
+n_avg = 1_000_000
+
+# Data to save
+save_data_dict = {
```
TheoLaudatQM, line 68: I think that saving the order could be useful if one wants to analyze the data later on

Quantum-Control-Applications/Optically addressable spin qubits/NV center in a confocal setup/09c_xy8_order.py
```diff
@@ -0,0 +1,211 @@
...
+
+    # Spin initialization
+    play("laser_ON", "AOM1")
+    wait(wait_for_initialization * u.ns, "AOM1")
+
+    # XY8-n sequence
+    with for_(n, 0, n < n_avg, n + 1):
+        with for_(*from_array(xy8_order, order_vec)):
+            play("x90", "NV")
+            with for_(i, 0, i < xy8_order, i + 1):
```
TheoLaudatQM: Isn't the `strict_timing` block needed here for the user to estimate the gap, as instructed in the docstring?

Quantum-Control-Applications/Optically addressable spin qubits/NV center in a confocal setup/09c_xy8_order.py
```diff
@@ -0,0 +1,211 @@
...
+        plt.plot(2 * order_vec, contrast, label="XY8 contrast")
+        plt.xlabel("XY8 order")
+        plt.ylabel("Contrast [kcps]")
+        plt.title("XY8 order sweep")
+        plt.legend()
+        plt.pause(0.1)
+    # Save results
+    script_name = Path(__file__).name
+    data_handler = DataHandler(root_data_folder=save_dir)
+    save_data_dict.update({"counts1_data": counts1})
```
TheoLaudatQM, line 254: Here again, I would personally save the processed data to make it easier for the user to regenerate the figures

Quantum-Control-Applications/Optically addressable spin qubits/NV center in a confocal setup/09b_xy8_tau.py
```diff
@@ -0,0 +1,217 @@
...
+
+#####################################
+#  Open Communication with the QOP  #
+#####################################
+qmm = QuantumMachinesManager(host=qop_ip, cluster_name=cluster_name, octave=octave_config)
+
+#######################
+# Simulate or execute #
+#######################
+simulate = True
```
TheoLaudatQM: I would set the default to False so that you don't have to always toggle it when running the script during an installation

Quantum-Control-Applications/Optically addressable spin qubits/NV center in a confocal setup/09c_xy8_order.py
```diff
@@ -0,0 +1,211 @@
...
+
+#####################################
+#  Open Communication with the QOP  #
+#####################################
+qmm = QuantumMachinesManager(host=qop_ip, cluster_name=cluster_name, octave=octave_config)
+
+#######################
+# Simulate or execute #
+#######################
+simulate = True
```
TheoLaudatQM: I would set the default to False so that you don't have to always toggle it when running the script during an installation

Quantum-Control-Applications/Optically addressable spin qubits/NV center in a confocal setup/05b_pulsed_odmr.py
NicolasStaudenmaier: Totally. Also to be consistent.

Quantum-Control-Applications/Optically addressable spin qubits/NV center in a confocal setup/05b_pulsed_odmr.py
NicolasStaudenmaier: In my experience you do both Rabi and pulsed ODMR initially with some good guess as you know the order of the params in your setup. For Rabi you have to "guess" the frequency, for pulsed ODMR it is the duration/amplitude. Then you update them one another. 
But I can also remove that line as in the list of measurements Rabi follows after this.

Quantum-Control-Applications/Optically addressable spin qubits/NV center in a confocal setup/09b_xy8_tau.py
NicolasStaudenmaier: In some cases you might want to shift the frame to rotate around different axes. I remove the sentence to avoid confusion.

Quantum-Control-Applications/Optically addressable spin qubits/NV center in a confocal setup/09b_xy8_tau.py
NicolasStaudenmaier: Good idea. I replace the count data with the one in kcps and add the contrast.

NicolasStaudenmaier:
Thank you for review, @TheoLaudatQM. I have addressed your comments. 
I will further work on the XY8 wrapper after discussing with Daniel. I will notify here, when it is ready for review again.

Quantum-Control-Applications/Optically addressable spin qubits/NV center in a confocal setup/05b_pulsed_odmr.py
TheoLaudatQM: I see, yes that makes sense and then I would just specify that it is an initial guess, like: "Setting the rough expected pi pulse amplitude and duration in the configuration"

Quantum-Control-Applications/Optically addressable spin qubits/NV center in a confocal setup/09b_xy8_tau.py
```diff
@@ -130,6 +130,7 @@ def xy8_block(tau):  # A single XY8 block
 
     with for_(n, 0, n < n_avg, n + 1):
         with for_each_((tau_spacing, tau_half_spacing), (tau_vec_spacing, tau_half_vec_spacing)):
+            wait(4)
```
TheoLaudatQM: What is this for?

Quantum-Control-Applications/Optically addressable spin qubits/NV center in a confocal setup/09c_xy8_order.py
```diff
@@ -43,7 +43,7 @@
 # Must be a multiple of 2 clock cycles to ensure that tau_half is a multiple of a single clock cycle
 tau = 2 * 50
 order_vec = np.arange(1, 21, 1, dtype=int)  # order vector for varying the order n of the XY8-n measurement
-n_avg = 1_000
+n_avg = 1_000_000
```
TheoLaudatQM, line 46: Is 1000000 averages really needed as the default value?
I usually prefer to have a smaller number of averages to ensure that the data is correct quickly, before increasing it to improve the SNR, but I leave it to you.

TheoLaudatQM (changes requested):
Nice, this is an interesting way of performing the readout!
I didn't test it, so I am trusting you that it compiles and simulates properly.
The only comment is that the docstrings at the beginning of the scripts should be updated to reflect the changes.

Quantum-Control-Applications/Optically addressable spin qubits/NV center in a confocal setup/09b_xy8_tau.py
NicolasStaudenmaier: The `for_each_` loop with "zipped" values needs 4 clock cycles to properly assign the values. Else there is a strict timing error. I can add a comment to it.

Quantum-Control-Applications/Optically addressable spin qubits/NV center in a confocal setup/09c_xy8_order.py
NicolasStaudenmaier: That is the default value for all measurements. I would keep it consistent. 
In order to see any meaningful signal one would have to use at least 10_000 rather 100_000 iterations.

Quantum-Control-Applications/Optically addressable spin qubits/NV center in a confocal setup/09b_xy8_tau.py
TheoLaudatQM: Yes please :)

PR #332, opened 2025-07-21
qm-qua>=1.2.2 API update

TheoLaudatQM:
API updates corresponding to qm-qua >= 1.2.2:

- `reset_phase()` --> `reset_if_phase()`
- Remove the `None` corresponding to the adc stream in the `measure` commands


yomach (changes requested):
Looks good, but there is one more change to remove deprecation messages
When there is an adc stream, we need to rename it:
`measure("readout", "tank_circuit", adc_st)`
should change to
`measure("readout", "tank_circuit", adc_stream=adc_st)`
And if there is both this and demod/other stuff, then the adc stream is at the end

PR #333, opened 2025-07-21
baud rate should be applied to both usb and network

soonqm:
The current qdac driver does not work on usb because of the missing baud rate


soonqm:
@TheoLaudatQM

PR #334, opened 2025-07-22
Ql 411 cz conditional phase

paulQM:
Ready for review

### DIFF
--- a/qualibration_graphs/superconducting/calibration_utils/cz_conditional_phase/analysis.py
+++ b/qualibration_graphs/superconducting/calibration_utils/cz_conditional_phase/analysis.py
@@ -1,0 +1,160 @@
+import logging
+from dataclasses import dataclass
+from typing import Dict, Tuple
+
+import numpy as np
+import xarray as xr
+from qualibrate import QualibrationNode
+from qualibration_libs.analysis import fit_oscillation, oscillation
+from scipy.optimize import curve_fit
+
+
+@dataclass
+class CzConditionalPhaseFit:
+    """Stores the relevant CZ conditional phase experiment fit parameters for a single qubit pair"""
+
+    optimal_amplitude: float
+    phase_diff: xr.DataArray
+    fitted_curve: np.ndarray
+    leakage: xr.DataArray
+    success: bool
+
+
+def fix_oscillation_phi_2pi(fit_data):
+    """Extract and fix the phase parameter from oscillation fit data."""
+    # Extract the phase parameter from the fit results
+    phase = fit_data.sel(fit_vals="phi")
+    # Normalize phase to [0, 1] range (representing 0 to 2π)
+    phase = (phase / (2 * np.pi)) % 1
+    return phase
+
+
+def tanh_fit(x, a, b, c, d):
+    """Tanh fitting function for phase difference vs amplitude."""
+    return a * np.tanh(b * x + c) + d
+
+
+def log_fitted_results(optimal_amps: Dict[str, float], log_callable=None):
+    """
+    Logs the node-specific fitted results for all qubit pairs.
+
+    Parameters:
+    -----------
+    optimal_amps : Dict[str, float]
+        Dictionary containing optimal amplitudes for each qubit pair.
+    log_callable : callable, optional
+        Logger for logging the fitted results. If None, a default logger is used.
+    """
+    if log_callable is None:
+        log_callable = logging.getLogger(__name__).info
+
+    for qp_name, amp in optimal_amps.items():
+        log_callable(f"Optimal CZ amplitude for {qp_name}: {amp:.6f} V --> SUCCESS!")
+
+
+def process_raw_dataset(ds: xr.Dataset, node: QualibrationNode):
+    """
+    Process the raw dataset by adding amplitude and detuning coordinates.
+
+    Parameters:
+    -----------
+    ds : xr.Dataset
+        Raw dataset from the experiment
+    node : QualibrationNode
+        The calibration node containing qubit pairs information
+
+    Returns:
+    --------
+    xr.Dataset
+        Processed dataset with additional coordinates
+    """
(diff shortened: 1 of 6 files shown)


qualibration_graphs/superconducting/calibrations/14_cz_conditional_phase.py
```diff
@@ -126,19 +122,26 @@ def create_qua_program(node: QualibrationNode[Parameters, Quam]):
...
             state_t_st = [declare_stream() for _ in range(num_qubit_pairs)]
 
-        for qubit in node.machine.active_qubits:
-            node.machine.initialize_qpu(target=qubit)
-
         for multiplexed_qubit_pairs in qubit_pairs.batch():
+            # Initialize the qubits
+            for qp in multiplexed_qubit_pairs.values():
+                node.machine.initialize_qpu(target=qp.qubit_control)
+                node.machine.initialize_qpu(target=qp.qubit_target)
```
TheoLaudatQM, line 127: Shouldn't the initialize_qpu method take the qubit pair as input to place them at the mutual flux point?
If you do this, then the control qubit will be moved when initializing the target qubit

qualibration_graphs/superconducting/calibrations/14_cz_conditional_phase.py
```diff
@@ -0,0 +1,301 @@
...
+        control_initial = declare(int)
+        n = declare(int)
+        n_st = declare_stream()
+        I_c, I_c_st, Q_c, Q_c_st, n, n_st = node.machine.declare_qua_variables()
+        I_t, I_t_st, Q_t, Q_t_st, _, _ = node.machine.declare_qua_variables()
+        if node.parameters.use_state_discrimination:
+            state_c = [declare(int) for _ in range(num_qubit_pairs)]
+            state_t = [declare(int) for _ in range(num_qubit_pairs)]
+            state_c_st = [declare_stream() for _ in range(num_qubit_pairs)]
+            state_t_st = [declare_stream() for _ in range(num_qubit_pairs)]
```
TheoLaudatQM, line 121:
```suggestion
        I_c, I_c_st, Q_c, Q_c_st, n, n_st = node.machine.declare_qua_variables()
        I_t, I_t_st, Q_t, Q_t_st, _, _ = node.machine.declare_qua_variables()
        amp = declare(fixed)
        frame = declare(fixed)
        control_initial = declare(int)
        if node.parameters.use_state_discrimination:
            state_c = [declare(int) for _ in range(num_qubit_pairs)]
            state_t = [declare(int) for _ in range(num_qubit_pairs)]
            state_c_st = [declare_stream() for _ in range(num_qubit_pairs)]
            state_t_st = [declare_stream() for _ in range(num_qubit_pairs)]
```

qualibration_graphs/superconducting/calibration_utils/cz_conditional_phase/analysis.py
```diff
@@ -0,0 +1,213 @@
...
+    xr.Dataset
+        Processed dataset with additional coordinates
+    """
+    qubit_pairs = node.namespace["qubit_pairs"]
+
+    def abs_amp(qp, amp):
+        return amp * qp.macros["cz_unipolar"].flux_pulse_control.amplitude
+
+    def detuning(qp, amp):
+        amplitude_squared = (amp * qp.macros["cz_unipolar"].flux_pulse_control.amplitude) ** 2
```
TheoLaudatQM: In the script you let the user chose the operation, but here it is hardcoded to cz_unipolar

qualibration_graphs/superconducting/calibration_utils/cz_conditional_phase/parameters.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class NodeSpecificParameters(RunnableParameters):
+    num_averages: int = 100
+    """Number of averages to perform. Default is 50."""
+    amp_range: float = 0.030
+    """Range of amplitude variation around the nominal value. Default is 0.030."""
+    amp_step: float = 0.001
+    """Step size for amplitude scanning. Default is 0.001."""
+    num_frames: int = 10
```
TheoLaudatQM: I would rename it `num_phases` or something a bit more explicit

qualibration_graphs/superconducting/calibrations/14_cz_conditional_phase.py
paulQM: We are not using pairwise mutual flux point, initialize_qpu takes a `flux_point` argument which for now we are leaving blank and letting it take the default value `"joint"`. I think for scaling you want to calibrate everything in joint mode always.

qualibration_graphs/superconducting/calibration_utils/cz_conditional_phase/analysis.py
paulQM: Nice catch!

qualibration_graphs/superconducting/calibration_utils/cz_conditional_phase/parameters.py
paulQM: How about `num_frame_rotations`?

TheoLaudatQM: Yes I am fine with it as long as it is explicit

qualibration_graphs/superconducting/calibrations/14_cz_conditional_phase.py
TheoLaudatQM: I think that this is a point worth double-checking with how Omrie and the team is doing, since it significantly impacts the program and the quam-builder methods (if so, then pair wise should be removed)

paulQM: I checked and they also use "joint"

PR #336, opened 2025-07-28
Fix: Add @quam_dataclass to root Quam component

nulinspiratie:


TheoLaudatQM (approved):
Sounds good, but what does it do exactly?

PR #345, opened 2025-08-21
Update python requirement

TheoLaudatQM:


TheoLaudatQM:
Hey @yomach since I had to update the poetry lock file, can you please take a quick look if everything is fine before I merge it?

yomach:
1. There are better ways, we can ask R&D for help if needed.
2. I'm not even sure we need it here, the Python (and numpy) requirements will come from qm-qua and qualang-tools... We can have much leaner requirements here
3. In any case, you can copy numpy (and others) from here: https://github.com/qm-labs/qm-qua-sdk/blob/main/pyproject.toml

PR #352, opened 2025-09-16
Feature/qualibration graph 2 q gates cr rand cz

KevinAVR:
This is the initial point of the Pull Request of CR and CZ (ZZ-driven) nodes into the main branch


qualibration_graphs/superconducting/CR_CZ_calibrations/30_CR_time_rabi_QST.py
```diff
@@ -0,0 +1,382 @@
...
+                    continue
+
+                # # cr drive
+                # operation = qp.cross_resonance.operations[node.parameters.wf_type]
+                # operation.amplitude = node.parameters.cr_drive_amp_scaling[i] * operation.amplitude
+                # operation.axis_angle = node.parameters.cr_drive_phase[i] * 2 * np.pi
+                # # cr cancel
+                # operation = qp.qubit_target.xy.operations[f"cr_{node.parameters.wf_type}_{qp.name}"]
+                # operation.amplitude = node.parameters.cr_cancel_amp_scaling[i] * operation.amplitude
+                # operation.axis_angle = node.parameters.cr_cancel_phase[i] * 2 * np.pi
```
soonqm, line 373: This is not necessary for QST. This code does not scan any pulse parameter.

qualibration_graphs/superconducting/CR_CZ_calibrations/30_CR_time_rabi_QST.py
```diff
@@ -0,0 +1,375 @@
...
+                    continue
+
+                # cr drive
+                operation_c = qp.cross_resonance.operations[node.parameters.wf_type]
+                operation_c.amplitude = node.parameters.cr_drive_amp_scaling[i] * operation_c.amplitude
+                operation_c.axis_angle = node.parameters.cr_drive_phase[i] * 2 * np.pi
+                # cr cancel
+                operation_t = qp.qubit_target.xy.operations[f"cr_{node.parameters.wf_type}_{qp.name}"]
+                operation_t.amplitude = node.parameters.cr_cancel_amp_scaling[i] * operation_t.amplitude
+                operation_t.axis_angle = node.parameters.cr_cancel_phase[i] * 2 * np.pi
```
soonqm: Not sure if this makes much sense because there is no fitting, but running this each time with amp_scaling != 1 will lead to runaway in the amplitude.

qualibration_graphs/superconducting/CR_CZ_calibrations/31a_CR_hamiltonian_tomography_vs_cr_drive_amp.py
```diff
@@ -0,0 +1,352 @@
...
+                    continue
+
+                # cr drive
+                operation_c = qp.cross_resonance.operations[node.parameters.wf_type]
+                operation_c.amplitude = node.parameters.cr_drive_amp_scaling[i] * operation_c.amplitude
+                operation_c.axis_angle = node.parameters.cr_drive_phase[i] * 2 * np.pi
+                # cr cancel
+                operation_t = qp.qubit_target.xy.operations[f"cr_{node.parameters.wf_type}_{qp.name}"]
+                operation_t.amplitude = node.parameters.cr_cancel_amp_scaling[i] * operation_t.amplitude
+                operation_t.axis_angle = node.parameters.cr_cancel_phase[i] * 2 * np.pi
```
soonqm: Same with 30_CR_time_rabi_QST comment on the amplitude runaway

soonqm: I think the way to go here is to choose a specific ZX interaction strength in the parameters and fit to the value. Then, the amplitude scan makes sense to why we are running it. Or we can just skip 30 and 31a completely without affecting anything.

Below is what we should see from 31a
image

qualibration_graphs/superconducting/CR_CZ_calibrations/31b_CR_hamiltonian_tomography_vs_cr_drive_phase.py
```diff
@@ -0,0 +1,339 @@
...
+                continue
+
+            # cr drive
+            operation_c = qp.cross_resonance.operations[node.parameters.wf_type]
+            operation_c.amplitude = node.parameters.cr_drive_amp_scaling[i] * operation_c.amplitude
+            operation_c.axis_angle = node.parameters.cr_drive_phase[i] * 2 * np.pi
+            # cr cancel
+            operation_t = qp.qubit_target.xy.operations[f"cr_{node.parameters.wf_type}_{qp.name}"]
+            operation_t.amplitude = node.parameters.cr_cancel_amp_scaling[i] * operation_t.amplitude
+            operation_t.axis_angle = node.parameters.cr_cancel_phase[i] * 2 * np.pi
```
soonqm: I would assume that the other parameters to be fixed after choosing the appropriate value in the previous step. I am removing all the unneccessary reference to the other parameters therof, in this case leaving only the cr_drive_phase. Any objection from @KevinAVR?

qualibration_graphs/superconducting/calibrations/31e_CR_phase_correction.py
```diff
@@ -0,0 +1,313 @@
...
+    # Get the active qubits from the loaded node parameters
+    node.namespace["qubit_pairs"] = get_qubit_pairs(node)
+
+
+# %% {Analyse_data}
+@node.run_action(skip_if=node.parameters.simulate)
+def analyse_data(node: QualibrationNode[Parameters, Quam]):
+    """Analyse the raw data and store the fitted data in another xarray dataset "ds_fit" and the fitted results in the "fit_results" dictionary."""
+    node.results["ds_raw"] = process_raw_dataset(node.results["ds_raw"], node)
+    node.results["ds_fit"], fit_results = fit_raw_data(node.results["ds_raw"], node)
```
soonqm, line 262: This is not implemented at the moment. Should be added later. I will raised an error at the end of this node.

soonqm:
I don't think 31c is needed, the cancellation phase is known analytically. If there is any reason why it deviates, we should instead do a 2D scan of cancellation amplitude and phase, since changing only one does not lead to minimization.

image

soonqm:
Currently, the protocol seems fine, some analysis is there but no fitting and state update. This needs to be added, preferably from testing on actual device.

qualibration_graphs/superconducting/CR_CZ_calibrations/31a_CR_hamiltonian_tomography_vs_cr_drive_amp.py
HiroQM: Indeed, the amplitude scan seems somewhat unnecessary. However, again, confirming the expected linear dependence can serve as a sanity check. In experiments, this can turn out to be non-linear.

PR #366, opened 2025-10-17
CZ error amplification node

paulQM:
Implements the error amplification version of the CZ calibration


qualibration_graphs/superconducting/calibrations/CZ_calibration_fixed_couplers/20b_cz_conditional_phase_error_amp.py
```diff
@@ -0,0 +1,302 @@
...
+            np.arange(1, num_operations + 1),
+            attrs={"long_name": "number of operations"},
+        ),
+        "amp": xr.DataArray(amplitudes, attrs={"long_name": "amplitude scale", "units": "a.u."}),
+        "frame": xr.DataArray(frames, attrs={"long_name": "frame rotation", "units": "2π"}),
+        "control_axis": xr.DataArray([0, 1], attrs={"long_name": "control qubit state"}),
+    }
+
+    # The QUA program stored in the node namespace to be transfer to the simulation and execution run_actions
+    with program() as node.namespace["qua_program"]:
```
TheoLaudatQM, line 123: I think that it would be nice to have the same comments and parameter names as for the single CZ (node 20a) for consistency. For instance, here the operation is called `operation_name` instead of `operation` in node 20a. I think that everything should be identical except the loop over applying CZ and the frame rotation trick.

qualibration_graphs/superconducting/calibrations/CZ_calibration_fixed_couplers/20b_cz_conditional_phase_error_amp.py
```diff
@@ -0,0 +1,302 @@
...
+                                    # Loop over the number of CZ operations
+                                    with for_(count, 0, count < n_op, count + 1):
+                                        # play the CZ gate
+                                        qp.macros[operation_name].apply(amplitude_scale_control=amp)
+                                    # rotate the frame by 𝜋/2 for odd number of operations
+                                    with if_(((n_op & 1) == 0) & (control_initial == 1)):
+                                        assign(frame_odd, frame - 0.5)
+                                        qp.qubit_target.xy.frame_rotation_2pi(frame_odd)
+                                    with else_():
+                                        qp.qubit_target.xy.frame_rotation_2pi(frame)
```
TheoLaudatQM, line 179: Wouldn't it be simpler to just do an even number of rotations, as it is done in power Rabi?

qualibration_graphs/superconducting/calibration_utils/cz_conditional_phase_error_amp/parameters.py
```diff
@@ -0,0 +1,29 @@
...
+
+from qualibrate import NodeParameters
+from qualibrate.parameters import RunnableParameters
+from qualibration_libs.parameters import CommonNodeParameters, QubitPairExperimentNodeParameters
+
+
+class NodeSpecificParameters(RunnableParameters):
+    num_averages: int = 100
+    """Number of averages to perform. Default is 100."""
+    amp_range: float = 0.010
```
TheoLaudatQM, line 11: 1. Is this in V or as a prefactor?
2. is it optimal value +/- amp_range or +/- amp_range/2?

qualibration_graphs/superconducting/calibration_utils/cz_conditional_phase_error_amp/parameters.py
```diff
@@ -0,0 +1,29 @@
...
+    """Range of amplitude variation around the nominal value. Default is 0.030."""
```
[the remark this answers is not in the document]
TheoLaudatQM: Should be `num_frame_rotations `as in the previous node

qualibration_graphs/superconducting/calibration_utils/cz_conditional_phase_error_amp/data_models.py
```diff
@@ -0,0 +1,15 @@
+from dataclasses import dataclass
```
TheoLaudatQM: Is this file being used at all?

qualibration_graphs/superconducting/calibrations/CZ_calibration_fixed_couplers/20b_cz_conditional_phase_error_amp.py
```diff
@@ -0,0 +1,302 @@
...
+
+The calibration process involves:
+1. Applying a CZ gate with varying amplitudes
+2. Measuring the phase of the target qubit for both control qubit states
+3. Calculating the phase difference
+4. Measuring the population in the |f> state to quantify leakage
+
+The optimal CZ gate amplitude is determined by finding the point where:
+1. The phase difference is closest to π (0.5 in normalized units)
+2. The leakage to the |f> state is minimized
```
TheoLaudatQM: I think that this need to be adjusted for the error amplification version

qualibration_graphs/superconducting/calibration_utils/cz_conditional_phase_error_amp/parameters.py
paulQM: - it's a prefactor
- Optimal +-range

paulQM: Should i change it with the /2 ? @TheoLaudatQM

qualibration_graphs/superconducting/calibrations/CZ_calibration_fixed_couplers/20b_cz_conditional_phase_error_amp.py
paulQM: Yes, but this is already working great and allows us to measure more data points. IQCC people only measure odd but i think this is better since you get more information and unless you already have a great fidelity you cannot apply many CZs before the data gets blurred out

TheoLaudatQM: Cool, I am convinced!
It would be worth explaining it in a comment ;)

qualibration_graphs/superconducting/calibration_utils/cz_conditional_phase_error_amp/parameters.py
TheoLaudatQM: Then I would write it in the docstring and I would keep the same convention as with the version without error amplification. Just write it explicitely.

qualibration_graphs/superconducting/calibrations/CZ_calibration_fixed_couplers/20a_cz_conditional_phase.py
```diff
@@ -1,9 +1,9 @@
 # %% {Imports}
 from dataclasses import asdict
 
+from iqcc_cloud_client import state
```
paulQM:
```suggestion

```

qualibration_graphs/superconducting/calibrations/CZ_calibration_fixed_couplers/19_chevron_11_02.py
```diff
@@ -394,12 +393,37 @@ def update_state(node: QualibrationNode[Parameters, Quam]):
...
+            )
+            pulse_amp = (
+                node.machine.qubit_pairs[qp.name].macros["cz_bipolar"].flux_pulse_control.get_reference() + "/amplitude"
+            )
+            pulse_name = node.machine.qubit_pairs[qp.name].macros["cz_bipolar"].flux_pulse_control_label
+            control_qb = node.machine.qubit_pairs[qp.name].qubit_control
+            control_qb.z.operations[pulse_name] = CosineBipolarPulse(length=100, amplitude=0.5, flat_length=10)
+            control_qb.z.operations[pulse_name].length = pulse_length
+            control_qb.z.operations[pulse_name].amplitude = pulse_amp
+            control_qb.z.operations[pulse_name].flat_length = flat_length
```
TheoLaudatQM: I am not sure I understand the logic here, why do you need to define the pulse twice (lines 400 and 415)?

qualibration_graphs/superconducting/calibrations/CZ_calibration_fixed_couplers/19_chevron_11_02.py
```diff
@@ -408,9 +432,15 @@ def update_state(node: QualibrationNode[Parameters, Quam]):
...
                     np.ceil(node.results["fit_results"][qp.name]["cz_len"] / 4) * 4
                 )
+                qp.macros["cz_bipolar"].flux_pulse_control.flat_length = int(
+                    np.ceil(node.results["fit_results"][qp.name]["cz_len"] / 2) * 2
+                )
                 qp.macros["cz_flattop"].flux_pulse_control.length = (
                     int(np.ceil(node.results["fit_results"][qp.name]["cz_len"] / 4) * 4) + 20
                 )
+                qp.macros["cz_bipolar"].flux_pulse_control.length = (
+                    int(np.ceil(node.results["fit_results"][qp.name]["cz_len"] / 4) * 4) + 12
```
TheoLaudatQM: Why is this value of 12ns hardcoded here?

qualibration_graphs/superconducting/calibrations/CZ_calibration_fixed_couplers/19_chevron_11_02.py
```diff
@@ -29,7 +29,7 @@
 from quam_config import Quam
 from scipy.optimize import curve_fit
 
-from quam.components.pulses import FlatTopGaussianPulse, SquarePulse
+from quam.components.pulses import CosineBipolarPulse, FlatTopGaussianPulse, SquarePulse
```
TheoLaudatQM: Because of this, we can merge this PR only after a new quam version is released

qualibration_graphs/superconducting/calibrations/CZ_calibration_fixed_couplers/19_chevron_11_02.py
```diff
@@ -29,7 +29,7 @@
 from quam_config import Quam
 from scipy.optimize import curve_fit
 
-from quam.components.pulses import FlatTopGaussianPulse, SquarePulse
```
TheoLaudatQM, line 32: Need to remove the packages that are not used from the imports

qualibration_graphs/superconducting/calibrations/CZ_calibration_fixed_couplers/20a_cz_conditional_phase.py
```diff
@@ -292,6 +326,12 @@ def update_state(node: QualibrationNode[Parameters, Quam]):
 # %% {Save_results}
 @node.run_action()
 def save_results(node: QualibrationNode[Parameters, Quam]):
+
+    for qp in node.namespace.get("tracked_qubit_pairs", []):
+        try:
+            qp.revert_changes()
+        except Exception:
+            pass
     node.save()
```
TheoLaudatQM: Why do you need this? In which case do you get an error?

qualibration_graphs/superconducting/calibrations/CZ_calibration_fixed_couplers/20b_cz_conditional_phase_error_amp.py
```diff
@@ -0,0 +1,347 @@
...
+                                    reset_frame(qp.qubit_control.xy.name)
+                                    # setting both qubits to the initial state
+                                    qp.qubit_control.xy.play("x180", condition=control_initial == 1)
+                                    qp.qubit_target.xy.play("x90")
+                                    qp.align()
+                                    # Loop over the number of CZ operations
+                                    with for_(count, 0, count < n_op, count + 1):
+                                        # play the CZ gate
+                                        qp.macros[operation_name].apply(amplitude_scale_control=amp)
+                                        qp.wait(4)  # wait for flux to settle
```
TheoLaudatQM: why do you need this here and not in the version without error amplification?

qualibration_graphs/superconducting/calibrations/CZ_calibration_fixed_couplers/20b_cz_conditional_phase_error_amp.py
```diff
@@ -0,0 +1,347 @@
...
+
+# %% {Save_results}
+@node.run_action()
+def save_results(node: QualibrationNode[Parameters, Quam]):
+
+    for qp in node.namespace.get("tracked_qubit_pairs", []):
+        try:
+            qp.revert_changes()
+        except Exception:
+            pass
```
TheoLaudatQM: Same here, why is this needed?

qualibration_graphs/superconducting/calibrations/1Q_calibrations/17_pi_vs_flux_long_distortions.py
```diff
@@ -173,6 +173,7 @@ def create_qua_program(node: QualibrationNode[Parameters, Quam]):
                         # Reset the qubits
                         for i, qubit in multiplexed_qubits.items():
                             qubit.reset(node.parameters.reset_type, node.parameters.simulate)
+                            qubit.wait(times.max())
```
TheoLaudatQM, line 177: What is this for?

qualibration_graphs/superconducting/calibrations/CZ_calibration_fixed_couplers/19_chevron_11_02.py
paulQM: we need to define twice as qubit.z pulse and macro pulse, they are referenced to each other later anyway

qualibration_graphs/superconducting/calibrations/CZ_calibration_fixed_couplers/19_chevron_11_02.py
paulQM: this will add it to the state so we need to give some value, the user can change it later. I don't know any criteria to decide the value

qualibration_graphs/superconducting/calibrations/CZ_calibration_fixed_couplers/19_chevron_11_02.py
paulQM: it's merged already :)

qualibration_graphs/superconducting/calibrations/CZ_calibration_fixed_couplers/20a_cz_conditional_phase.py
paulQM: i have seen it like this in all nodes that use revert changes. I guess it's to allow the save to get executed in case of an error

qualibration_graphs/superconducting/calibrations/CZ_calibration_fixed_couplers/20b_cz_conditional_phase_error_amp.py
paulQM: this is to make sure pulses don't play back to back and create a flat effect (for square and snz it could be a problem)

qualibration_graphs/superconducting/calibrations/1Q_calibrations/17_pi_vs_flux_long_distortions.py
paulQM: we realised that on some devices that have very long distortions you can have tail effects between different iterations of the same experiment. We have seen cases where the thermal reset is not long enough for the distortion to fully settle

TheoLaudatQM: Then, as always, a comment is needed to explain it to customers.

qualibration_graphs/superconducting/calibrations/CZ_calibration_fixed_couplers/20b_cz_conditional_phase_error_amp.py
TheoLaudatQM: First, if so then the comment is wrong.
Second, I don't get it, why would it create a problem? For instance, you won't have it when doing RB or arbitrary algorithms... If you always need it, then it should be part of the macro.

qualibration_graphs/superconducting/calibrations/CZ_calibration_fixed_couplers/20a_cz_conditional_phase.py
TheoLaudatQM: It is not in the ones from qua-libs (tof, res spec vs power, rabi chevron...)
In the end it is only in the ones you added recently...

qualibration_graphs/superconducting/calibrations/CZ_calibration_fixed_couplers/19_chevron_11_02.py
TheoLaudatQM: But then I don't think that it is the right way to do it, because it goes against the spirit of running nodes through qualibrate.
If a parameter needs to be user defined, then it should be a node parameter.
I think that it would make more sense to add the 2Q gates in the populate quam file, similarly to the 1Q gates, so that users can freely define the shapes and parameters without having to modify the chevron node.

qualibration_graphs/superconducting/calibrations/CZ_calibration_fixed_couplers/19_chevron_11_02.py
TheoLaudatQM: It looks very ugly and possibly confusing for customers.
Since everything is already contained in this line 
```
cz_pulse = CosineBipolarPulse(length=100, amplitude=0.5, id="cz_bipolar_pulse", flat_length=10)
```
Then why not doing it inside the function directly to abstract it away from the node?

qualibration_graphs/superconducting/calibrations/CZ_calibration_fixed_couplers/19_chevron_11_02.py
TheoLaudatQM: Merged doesn't mean available to customers (cf. the [changelog ](https://github.com/qua-platform/quam/blob/main/CHANGELOG.md)). For installable packages such as quam and py-qua-tools, you need to release a new version

qualibration_graphs/superconducting/calibrations/CZ_calibration_fixed_couplers/20b_cz_conditional_phase_error_amp.py
Deepakkhurrana: I think it would be good idea to see what happens to error amplification if you remove this wait time? If the "flat effect" messes up things, we need to add it to the macro itself

qualibration_graphs/superconducting/calibrations/CZ_calibration_fixed_couplers/19_chevron_11_02.py
paulQM: Let's ask Serwan to make a release then :)

qualibration_graphs/superconducting/calibrations/CZ_calibration_fixed_couplers/20a_cz_conditional_phase.py
paulQM: Ok i'll remove it then. I think I got that from Brad's pi vs flux node and thought it was needed

qualibration_graphs/superconducting/calibrations/CZ_calibration_fixed_couplers/20b_cz_conditional_phase_error_amp.py
paulQM: Ok i will remove it then :)

qualibration_graphs/superconducting/calibrations/CZ_calibration_fixed_couplers/19_chevron_11_02.py
paulQM: OK, in this case we can create the macro instances with default values in the populate quam, and then the user can pick his durations. There is however a potential issue that could arise, the chevron node will fit the flat duration and apply it to the already present macro, but if the total duration that was defined previously is shorter than the fitted flat section it will create an error. I propose to check that the defined total duration is longer than the fitted flat section and when that is not the case apply the +12. What do you think?

qualibration_graphs/superconducting/calibrations/CZ_calibration_fixed_couplers/19_chevron_11_02.py
paulQM: and add requirement version

paulQM: @TheoLaudatQM , i updated pyproject with the latest prerelease of quam containing bipolar pulses. All comments are now resolved and we can merge

qualibration_graphs/superconducting/calibrations/CZ_calibration_fixed_couplers/19_chevron_11_02.py
TheoLaudatQM: I don't like it since it is very manual and the 12 is still arbitrary.
It would have been much better to define the bipolar pulse with two durations (flat and rise/fall) so that the total length is inferred from the two.
What can be done now is at least to update the total length automatically from the fitted flat length and the rise/fall duration. I don't see why it can't be automated and users have to adjust it manually.

paulQM: Maybe i am missing something but to me this is equally arbitrary. The +12 is nothing more than the rise/fall time, you would still need to give it an initial arbitrary value. Both notations are equivalent, why do you like more defining the rise/fall instead of the total length? Are you assuming people will stick to a fixed rise/fall and the nodes should only update the flat duration and keep the same rise/fall?

TheoLaudatQM: Users don't have to chose 12ns for the rise/fall duration as it is set in generate quam, so if they change it they have to change it in the node as well, which is not good.
From what I understood, the node is only affecting the flat section, and we don't have a node to calibrate the rise and fall part, do we? How do you choose those?
The advantage of providing the flat length and rise/fall, removes the need for manually having to adjust the total length of the pulse in the node.

qualibration_graphs/superconducting/calibration_utils/cz_phase_compensation/parameters.py
```diff
@@ -20,7 +20,7 @@ class NodeSpecificParameters(RunnableParameters):
 
     num_shots: int = 100
     num_frames: int = 17
-    operation: Literal["cz_flattop", "cz_unipolar"] = "cz_unipolar"
+    operation: Literal["cz_flattop", "cz_unipolar", "cz_bipolar"] = "cz_unipolar"
```
OziEgri, line 23: Exactly same new Literal definition was added  in qualibration_graphs/superconducting/calibration_utils/cz_conditional_phase/parameters.py
Isn't there a common parameters file to hold a single def instead of both?

paulQM: These are the gates we currently add to the quam state but nothing prevents the user to manually define a new one so maybe we should think of actually changing the type to `str` and assert that the operation exists before execution instead. Storing gate names as enums on a separate file might be uncomfortable once people start doing more advanced custom gates. What do you think @OziEgri ?

OziEgri: I think most people are used to work with a config/define file where all definitions are placed in the same place. If you feel it is too much at the moment, maybe move all definitions to top of the current file.

qualibration_graphs/superconducting/calibration_utils/cz_conditional_phase_error_amp/__init__.py
```diff
@@ -0,0 +1,12 @@
+from .analysis import FitResults, fit_raw_data, log_fitted_results, process_raw_dataset
+from .parameters import Parameters
+from .plotting import plot_raw_data_with_fit
+
+__all__ = [
```
OziEgri, line 5: Nice, this is how it should be done to deal with all wildcard imports

qualibration_graphs/superconducting/calibration_utils/cz_conditional_phase_error_amp/analysis.py
```diff
@@ -0,0 +1,293 @@
...
+    optimal_amplitude: float
+    success: bool
+
+
+def fix_oscillation_phi_2pi(fit_data):
+    """Extract and fix the phase parameter from oscillation fit data."""
+    # Extract the phase parameter from the fit results
+    phase = fit_data.sel(fit_vals="phi")
+    # Normalize phase to [0, 1] range (representing 0 to 2π)
+    phase = (phase / (2 * np.pi)) % 1
```
OziEgri, line 26: Maybe: phase = np.mod(phase / (2 * np.pi), 1.0)

qualibration_graphs/superconducting/calibrations/CZ_calibration_fixed_couplers/19_chevron_11_02.py
```diff
@@ -349,68 +341,42 @@ def plot_data(node: QualibrationNode[Parameters, Quam]):
...
-                )
+                if node.parameters.update_all_pulses:
+                    qp.macros["cz_bipolar"].flux_pulse_control.amplitude = node.results["fit_results"][qp.name]["cz_amp"]
+                    qp.macros["cz_flattop"].flux_pulse_control.amplitude = node.results["fit_results"][qp.name]["cz_amp"]
+                    # Round up to the upper 4 ns to be compatible with the hardware time resolution
+                    qp.macros["cz_flattop"].flux_pulse_control.flat_length = int(
+                        np.ceil(node.results["fit_results"][qp.name]["cz_len"] / 4) * 4
+                    )
+                    qp.macros["cz_bipolar"].flux_pulse_control.flat_length = int(
+                        np.ceil(node.results["fit_results"][qp.name]["cz_len"] / 2) * 2
```
OziEgri, line 366: As you are using this round up a lot, maybe consider: 
def ceil_to_even(x):
    return int(np.ceil(x / 2) * 2)

cz_len = ceil_to_even(node.results["fit_results"][qp.name]["cz_len"])


similar method can be used to round to the nearest multiple of 4

qualibration_graphs/superconducting/calibrations/CZ_calibration_fixed_couplers/20a_cz_conditional_phase.py
```diff
@@ -61,7 +61,7 @@
 
 # Be sure to include [Parameters, Quam] so the node has proper type hinting
 node = QualibrationNode[Parameters, Quam](
-    name="20_cz_conditional_phase",  # Name should be unique
+    name="20a_cz_conditional_phase",  # Name should be unique
```
OziEgri, line 64: IDK if initializing name with unique string is done often, but if it is, consider adding a service function for it: 
used_names = set()

def create_unique_name(name: str) -> str:
    if name in used_names:
        raise ValueError(f"Name '{name}' is already used.")
    used_names.add(name)
    return name

qualibration_graphs/superconducting/calibrations/CZ_calibration_fixed_couplers/99_CZ_calibration_graph.py
```diff
@@ -20,11 +20,15 @@ class Parameters(GraphParameters):
     parameters=Parameters(),
     nodes={
         "chevron": library.nodes["19_chevron_1102"].copy(name="chevron"),
-        "conditional_phase": library.nodes["20_cz_conditional_phase"].copy(name="conditional_phase"),
+        "conditional_phase": library.nodes["20a_cz_conditional_phase"].copy(name="conditional_phase"),
```
OziEgri, line 23: consider defining this string in a config/define file, as it is used in many places.

OziEgri (commented):
I left a few minor comments on code style. Hopefully in a few weeks I’ll know enough to give feedback on the physics side as well.

PR #367, opened 2025-10-21
NV basic files readout method

NicolasStaudenmaier:
This PR yields a bunch of features and fixes for the collection of NV center basic scripts. The main feature is the usage of a new readout method for NV centers. Other relevant points are fixing a bug in `03_counter.py`, introducing corrections in case of 2GHz sampling rate and time bins for time tagging with QOP >= 350 and fixing deprecation warnings and typos. 

All programs where applicable are updated with the new readout method. In this method, a reference measurement is done at the end of the readout laser pulse. The average of the signal (first measure) and the reference (second measure) can be used to calculate a normalized trace. Here this is done in post-processing, not with the pulse streamer. This methods avoids the need of measuring "dark counts" for reference. For NV centers this method is preferred because it allows to acquire readout statistics much faster. Compared to many other qubit platforms this is a central bottleneck. 

Additional features/fixes are summarized as follows:
- fix bug in `03_counter.py`,
- add commented alternative for 2GHz sampling rate (`02_raw_adc_traces`) and 0.5ns time tag bins (`04_calibrate_delays`) with QOP >= 350,
- fix deprecated syntax (remove `None` if no ADC stream is used, `timeTaggingParameters` in config), 
- fix minor typos.

### DIFF
--- a/Quantum-Control-Applications/Optically addressable spin qubits/NV center in a confocal setup/04_calibrate_delays.py
+++ b/Quantum-Control-Applications/Optically addressable spin qubits/NV center in a confocal setup/04_calibrate_delays.py
@@ -108,9 +108,9 @@ with program() as calib_delays:
 
     with stream_processing():
         # Directly derive the histograms in the stream processing
-        times_st.histogram(
-            [[i, i + (resolution_samp - 1)] for i in range(0, total_samples, resolution_samp)]
-        ).save("times_hist")
+        times_st.histogram([[i, i + (resolution_samp - 1)] for i in range(0, total_samples, resolution_samp)]).save(
+            "times_hist"
+        )
         times_st_dark.histogram(
             [[i, i + (resolution_samp - 1)] for i in range(0, total_samples, resolution_samp)]
         ).save("times_hist_dark")


Quantum-Control-Applications/Optically addressable spin qubits/NV center in a confocal setup/02_raw_adc_traces.py
```diff
@@ -81,17 +81,21 @@
     # Fetch results and convert traces to volts
     adc1 = u.raw2volts(res_handles.get("adc1").fetch_all())
     adc1_single_run = u.raw2volts(res_handles.get("adc1_single_run").fetch_all())
+    # Determine time axis from the sampling rate: 1e9 for OPX+, 1e9 or 2e9 for OPX1000
+    time_array = np.arange(len(adc1))  # for 1e9 sampling rate
+    # time_array = 0.5 * np.arange(len(adc1))  # for 2e9 sampling rate
```
TheoLaudatQM: Wouldn't it better to use the `sampling_rate` parameter that is already defined in the config? (at least for the OPX1000 ones, but then it could be added to the OPX+ ones)

Quantum-Control-Applications/Optically addressable spin qubits/NV center in a confocal setup/04_calibrate_delays.py
```diff
@@ -28,11 +28,12 @@
 wait_between_runs = 10_000  # [ns]
 n_avg = 1_000_000
 
-resolution = 12  # ns
+resolution = 12  # in samples: 1ns for QOP <3.5.0 (and 2.x.x), 0.5ns for QOP >=3.5.0
 # total measurement length (ns), add 2*laser_delay to ensure that the window is larger than the laser pulse
 meas_len = int(initialization_len + 2 * laser_delay)
 # Time vector for plotting
 t_vec = np.arange(0, meas_len, 1)
+# t_vec = np.arange(0, meas_len, 0.5)  # for QOP >=3.5.0
```
TheoLaudatQM: Is this only dependent on the version, or also the sampling rate?
Would it make sense to detect the QOP version from `qmm.capabilities` and do the conversion automatically?
I am just afraid that people will oversee this and run into issues...

Quantum-Control-Applications/Optically addressable spin qubits/NV center in a confocal setup/04_calibrate_delays.py
```diff
@@ -98,6 +99,7 @@
 
     with stream_processing():
         # Directly derive the histograms in the stream processing
+        # NOTE: use `2 * meas_len` for QOP version >= 3.5.0 because time bins are 0.5ns
```
TheoLaudatQM: Same consideration here, I would actually always set the resolution defined above in ns, so that users won't need to think about it in the plotting section, as having a different unit depending whether you are on QOP350 or below is confusing.

Quantum-Control-Applications/Optically addressable spin qubits/NV center in a confocal setup/04_calibrate_delays.py
```diff
@@ -149,6 +151,7 @@
         # Plot data
         plt.cla()
         plt.plot(t_vec[::resolution] + resolution / 2, times_hist / 1000 / (resolution / u.s) / iteration)
+        # plt.plot(t_vec[::resolution] + resolution / 2, times_hist / 1000 / (0.5 * resolution / u.s) / iteration)  # for QOP >=3.5.0
```
TheoLaudatQM: Same here.

Quantum-Control-Applications/Optically addressable spin qubits/NV center in a confocal setup/configuration.py
```diff
@@ -181,7 +180,7 @@ def IQ_imbalance(g, phi):
                 "long_readout": "long_readout_pulse_1",
             },
             "outputs": {"out1": ("con1", 1)},
-            "outputPulseParameters": {
+            "timeTaggingParameters": {
```
TheoLaudatQM, line 185: Does this depend on the QOP version, or only qm-qua?

Quantum-Control-Applications/Optically addressable spin qubits/NV center in a confocal setup/02_raw_adc_traces.py
NicolasStaudenmaier: I didn't use it because it doesn't exist in the OPX+ config. If you think it make sense to just add it and to use it here, I can do it.

Quantum-Control-Applications/Optically addressable spin qubits/NV center in a confocal setup/04_calibrate_delays.py
NicolasStaudenmaier: It is the time vector associated to the time tag samples. It only depends on the QOP version (from QOP350 it is 0.5ns, prior it is 1ns). 
I haven't used `qmm.capabilities` so far and also don't find it in the documentation. Does it make sense to add it if it's not documented?

Quantum-Control-Applications/Optically addressable spin qubits/NV center in a confocal setup/configuration.py
NicolasStaudenmaier: It is qm-qua. `outputPulseParameters` is deprecated syntax already since 1.2.2 (precisely 1.2.2a3).

Quantum-Control-Applications/Optically addressable spin qubits/NV center in a confocal setup/04_calibrate_delays.py
TheoLaudatQM: You have a mention of it in the [documentation](https://docs.quantum-machines.co/latest/docs/API_references/qm_manager_api/?h=capabili#qm.quantum_machines_manager.QuantumMachinesManager.set_capabilities_offline), although it isn't very extensive....
If it is too complicated to automated, then let's keep it as is and just bear in mind that it can be a source of user errors

Quantum-Control-Applications/Optically addressable spin qubits/NV center in a confocal setup/02_raw_adc_traces.py
TheoLaudatQM: I think that it would be nice as it will avoid the need for customers to manually change the scripts based on a parameter that they will change in the config already, but I leave it to you.

NicolasStaudenmaier: Added it.

Quantum-Control-Applications/Optically addressable spin qubits/NV center in a confocal setup/04_calibrate_delays.py
NicolasStaudenmaier: I use `qmm.version_dict()` to determine the QOP version and convert the string to a tuple for version comparison. From that the time bin is set and used for further calculations. I tested the two `04_calibrate_delays` scripts with random data and both work fine.

Quantum-Control-Applications/Optically addressable spin qubits/NV center in a confocal setup/04_calibrate_delays.py
NicolasStaudenmaier: See other comment.

Quantum-Control-Applications/Optically addressable spin qubits/NV center in a confocal setup/04_calibrate_delays.py
NicolasStaudenmaier: See other comment.

PR #389, opened 2025-11-28
Update the Octave connection for the latest QOP244 onwards

TheoLaudatQM:


MichalGQM (approved):
Thanks for modifying it!

PR #410, opened 2026-01-22
Feat/res spectroscopy part1

KU-QM:
Commented out most of the resonator spectroscopy script. Currently only the Parameters, description and the update are configured. OK these before moving onto next steps.

### DIFF
--- a/qualibration_graphs/quantum_dots/calibration_utils/resonator_spectroscopy/parameters.py
+++ b/qualibration_graphs/quantum_dots/calibration_utils/resonator_spectroscopy/parameters.py
@@ -18,7 +18,6 @@ class NodeSpecificParameters(RunnableParameters):
 
 class Parameters(
     NodeParameters,
-    VideoModeCommonParameters,
     BaseExperimentNodeParameters,
     CommonNodeParameters,
     NodeSpecificParameters,
--- a/qualibration_graphs/quantum_dots/calibrations/loss_divincenzo/02a_resonator_spectroscopy.py
+++ b/qualibration_graphs/quantum_dots/calibrations/loss_divincenzo/02a_resonator_spectroscopy.py
@@ -138,7 +138,7 @@ def simulate_qua_program(node: QualibrationNode[Parameters, Quam]):
 
 # %% {Execute}
 @node.run_action(
-    skip_if=node.parameters.load_data_id is not None or node.parameters.simulate or node.parameters.run_in_video_mode
+    skip_if=node.parameters.load_data_id is not None or node.parameters.simulate
 )
 def execute_qua_program(node: QualibrationNode[Parameters, Quam]):
     """Connect to the QOP, execute the QUA program and fetch the raw data and store it in a xarray dataset called "ds_raw"."""
@@ -177,7 +177,7 @@ def load_data(node: QualibrationNode[Parameters, Quam]):
 
 
 # %% {Analyse_data}
-@node.run_action(skip_if=node.parameters.simulate or node.parameters.run_in_video_mode)
+@node.run_action(skip_if=node.parameters.simulate)
 def analyse_data(node: QualibrationNode[Parameters, Quam]):
     """Analyse the raw data and store the fitted data in another xarray dataset "ds_fit" and the fitted results in the "fit_results" dictionary."""
 #     node.results["ds_raw"] = process_raw_dataset(node.results["ds_raw"], node)
@@ -193,7 +193,7 @@ def analyse_data(node: QualibrationNode[Parameters, Quam]):
 
 
 # %% {Plot_data}
-@node.run_action(skip_if=node.parameters.simulate or node.parameters.run_in_video_mode)
+@node.run_action(skip_if=node.parameters.simulate)
 def plot_data(node: QualibrationNode[Parameters, Quam]):
     """Plot the raw and fitted data in specific figures whose shape is given by sensors.grid_location."""
 #     fig_raw_phase = plot_raw_phase(node.results["ds_raw"], node.namespace["sensors"])
@@ -209,7 +209,7 @@ def plot_data(node: QualibrationNode[Parameters, Quam]):
 
 
 # %% {Update_state}
-@node.run_action(skip_if=node.parameters.simulate or node.parameters.run_in_video_mode)
+@node.run_action(skip_if=node.parameters.simulate)
 def update_state(node: QualibrationNode[Parameters, Quam]):
     """Update the relevant parameters if the sensor_name data analysis was successful."""
     with node.record_state_updates():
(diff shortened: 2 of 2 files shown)


qualibration_graphs/quantum_dots/calibration_utils/resonator_spectroscopy/parameters.py
```diff
@@ -2,27 +2,23 @@
...
     """Span of frequencies to sweep in MHz. Default is 30 MHz."""
     frequency_step_in_mhz: float = 0.1
     """Step size for frequency sweep in MHz. Default is 0.1 MHz."""
-    dc_control: bool = False
-    """If an associated external DC offset exists."""
 
 
 class Parameters(
     NodeParameters,
+    VideoModeCommonParameters,
```
TheoLaudatQM: Do you plan to use video mode for this one?

qualibration_graphs/quantum_dots/calibrations/loss_divincenzo/02a_resonator_spectroscopy.py
```diff
@@ -61,151 +61,151 @@ def custom_param(node: QualibrationNode[Parameters, Quam]):
...
-    fig_fit_amplitude = plot_raw_amplitude_with_fit(
-        node.results["ds_raw"], node.namespace["sensors"], node.results["ds_fit"]
-    )
-    plt.show()
-    # Store the generated figures
-    node.results["figures"] = {
-        "phase": fig_raw_phase,
-        "amplitude": fig_fit_amplitude,
-    }
+# node.machine = Quam.load()
```
TheoLaudatQM: You can keep the machine at this stage and all the run_actions with docstrings. Just comment what is inside of it. For instance:


```
# %% {Analyse_data}
@node.run_action(skip_if=node.parameters.simulate or node.parameters.run_in_video_mode)
def analyse_data(node: QualibrationNode[Parameters, Quam]):
    """Analyse the raw data and store the fitted data in another xarray dataset "ds_fit" and the fitted results in the "fit_results" dictionary."""
#     node.results["ds_raw"] = process_raw_dataset(node.results["ds_raw"], node)
#     node.results["ds_fit"], fit_results = fit_raw_data(node.results["ds_raw"], node)
#     node.results["fit_results"] = {k: asdict(v) for k, v in fit_results.items()}

#     # Log the relevant information extracted from the data analysis
#     log_fitted_results(node.results["fit_results"], log_callable=node.log)
#     node.outcomes = {
#         sensor_name: ("successful" if fit_result["success"] else "failed")
#         for sensor_name, fit_result in node.results["fit_results"].items()
#     }
```

qualibration_graphs/quantum_dots/calibrations/loss_divincenzo/02a_resonator_spectroscopy.py
```diff
@@ -220,38 +220,38 @@ def update_state(node: QualibrationNode[Parameters, Quam]):
             q.readout_resonator.intermediate_frequency = float(node.results["fit_results"][q.name]["frequency"])
```
TheoLaudatQM: Since we are dealing with the sensors here, I wouldn't call it q:
```
for s in node.namespace["sensors"]:
            if node.outcomes[q.name] == "failed":
                continue
            s.readout_resonator.intermediate_frequency = float(node.results["fit_results"][q.name]["frequency"])
```

TheoLaudatQM: Does the readout frequency depend on the qubit that you are sensing?
Say I have a resonator coupled to a sensor dot that is coupled to 2 qubits, will I use the same readout frequency for probing the two qubits or will it depend on the qubit/quantum dot pair that I want to measure?

sebastianorbell-qm: The thresholding information will change but i think the frequency could be the same. We can check the literature

qualibration_graphs/quantum_dots/calibration_utils/resonator_spectroscopy/parameters.py
KU-QM: Yes, we can open video mode directly as a 1D sensor sweep. It is not crucial, since I doubt live plotting is entirely necessary for resonator spectroscopy, but just in-case.

qualibration_graphs/quantum_dots/calibrations/loss_divincenzo/02a_resonator_spectroscopy.py
KU-QM: The frequency should be the same I think. But this is not done per qubit, it should be done per sensor. I think the letter 'q' might be a bit misleading here. The readout_resonator attr is only applicable to SensorDot objects

TheoLaudatQM: Right, but I think that the Qualibrate node has `qubits` as a native parameter, which means that users will want to run nodes on defined qubits and not sensors. For consistency, it means that the node outcomes should also be referenced with the qubit name and not the sensor name. This can probably be done by mapping each qubit with a sensor and connecting the node outcome for each qubit to the corresponding sensors, what do you think?

qualibration_graphs/quantum_dots/calibration_utils/resonator_spectroscopy/parameters.py
TheoLaudatQM: Then wouldn't it be simpler to write this node without the video mode integration for simplicity?

KU-QM: My thinking was, what if the user is calibrating this from the ground up with no knowledge of the resonant frequency. In that case, it would help to be able to jump around quickly via video mode and look for the resonance. I'm very open to suggestion though - if we think that it is not necessary then we can get rid of it. The videomode runner is not super complex to add to a node, so I thought we may as well. Thoughts?

TheoLaudatQM: How would video mode help finding the resonance?
What I would do is just perform a quick sweep over the full LF-fem bandwidth, identify the resonance and run the node with a reduced span and increased resolution

qualibration_graphs/quantum_dots/calibrations/loss_divincenzo/02a_resonator_spectroscopy.py
KU-QM: I see your point. On the one hand, it would make it more consistent with the other nodes in quantum_dots, and also with the superconducting node for resonator spectroscopy. However, a few counter-arguments: 
- At this stage of the experiment, since we are (most likely) working with BaseQuamQD instead of LossDiVincenzoQuam, there is no guarantee that our quam even has a 'qubits' attr. This is why in calibration_utils/common_utils/experiment.py, the BaseExperimentNodeParameters has 'sensor_names' at the end. During this part of the experiment, the user might not have a good idea even of what the qubits in their system are, so having them input qubit names doesn't seem ideal to me.
- Unlike the superconducting counterpart, the resonators are coupled to the sensors. This means that conceptually, it feels to me like it makes more sense to run this node sensor-first rather than qubit-first. 

There is an argument to make the input quantum dots rather than sensor_names, in a similar way to what would be done for qubits. However, if we are changing this anyway (from node.parameters.qubits), we might as well keep the sensor version, since that seems more experimentally intuitive. What do you think?

qualibration_graphs/quantum_dots/calibration_utils/resonator_spectroscopy/parameters.py
KU-QM: You can do 1D resonator frequency sweeps in video mode. The way that I had started writing the node, opening video mode from this node will directly open a 1D sweep with the right parameters (span, averaging). In the annotationtab, you can also save the new IF to the Quam state. If you think this isn't necessary to the node, I can remove it!

PR #491, opened 2026-03-06
Sweep detuning l3

KU-QM:


qualibration_graphs/quantum_dots/calibrations/loss_divincenzo/05b_PSB_search_opx_sweep_detuning.py
```diff
@@ -225,13 +232,26 @@ def load_data(node: QualibrationNode[Parameters, Quam]):
 # %% {Analyse_data}
 @node.run_action(skip_if=node.parameters.simulate)
 def analyse_data(node: QualibrationNode[Parameters, Quam]):
-    """Analyse the raw data and store the fitted data in another xarray dataset "ds_fit" and the fitted results in the "fit_results" dictionary."""
+    """Fit sigmoid to sensor signal vs detuning to locate the PSB transition."""
+    node.results["ds_fit"], fit_results = fit_raw_data(node.results["ds_raw"], node)
+    node.results["fit_results"] = {k: {kk: vv for kk, vv in v.items() if kk != "_diag"} for k, v in fit_results.items()}
+    node.namespace["_fit_results_full"] = fit_results
+    log_fitted_results(node.results["fit_results"], log_callable=node.log)
```
sebastianorbell-qm, line 239: I thought that for each detuning we would plot the IQ clouds, fit a GMM or Barthel type model, and calculate the readout visibility / SNR - or does this happen in a different node?

KU-QM: Thought that was the fixed detuning version? Wasn't sweep detuning should just be to find the right detuning value?

PR #504, opened 2026-03-24
fix: Removing simultaneous operation blocks

sebastianorbell-qm:
Remove simultaneous-operation blocks for multiplexing calibrations. Proper multiplexing support requires broader long-term restructuring, so the current implementation was not viable.


qualibration_graphs/quantum_dots/calibrations/loss_divincenzo/09a_time_rabi_parity_diff.py
```diff
@@ -100,81 +105,76 @@ def create_qua_program(node: QualibrationNode[Parameters, Quam]):
...
                     # ---------------------------------------------------------
                     # Step 3: X180 - apply pi pulse
                     # ---------------------------------------------------------
-                    for i, qubit in batched_qubits.items():
-                        # X180 macro handles X180 pulse
-                        qubit.x180(duration=t)
+                    # Keep the operating-point hold synchronized with the
+                    # stretched Gaussian pulse duration.
+                    qubit.voltage_sequence.step_to_voltages({}, duration=t * 4)
+                    qubit.xy.play(reference_pulse_name, duration=t)
```
KU-QM: Wouldn't it be more consistent to use macros? I thought macros can track the state - or does that not work yet with a variable duration?

qualibration_graphs/quantum_dots/calibrations/loss_divincenzo/10_T1_parity_diff.py
```diff
@@ -125,55 +125,47 @@ def create_qua_program(node: QualibrationNode[Parameters, Quam]):
...
                     # Step 3: X180 - apply pi pulse and idle
                     # ---------------------------------------------------------
                     align()
-                    for i, qubit in batched_qubits.items():
-                        qubit.x180()
-                        # qubit.xy.wait(t)
-                        wait(t)
-                        qubit.voltage_sequence.step_to_voltages({}, duration=t * 4)
+                    qubit.x180()
+                    wait(t)
```
KU-QM: I wonder if this wait is necessary. Why not just use the empty step_to_voltages?

qualibration_graphs/quantum_dots/calibrations/loss_divincenzo/10a_ramsey_parity_diff.py
```diff
@@ -100,66 +100,58 @@ def create_qua_program(node: QualibrationNode[RamseyParameters, Quam]):
...
-                        for i, qubit in batched_qubits.items():
-                            qubit.x90()
-                            align()
-                            wait(t)
-                            qubit.voltage_sequence.step_to_voltages({}, duration=t * 4)
-                            align()
-                            qubit.x90()
+                        qubit.x90()
+                        align()
+                        wait(t)
```
KU-QM: Here as well - do we need both a wait and an empty step_to_voltages

qualibration_graphs/quantum_dots/calibrations/loss_divincenzo/08a_power_rabi.py
```diff
@@ -264,11 +257,10 @@ def update_state(node: QualibrationNode[Parameters, Quam]):
...
 
             opt_prefactor = node.results["fit_results"][q.name]["opt_amp"]
-            current_amp = q.xy.operations[node.parameters.operation].amplitude
+            reference_pulse = get_xy_reference_pulse(q)
+            current_amp = reference_pulse.amplitude
             new_amplitude = current_amp * opt_prefactor
-            q.xy.operations[node.parameters.operation].amplitude = new_amplitude
-            if node.parameters.operation == "x180" and node.parameters.update_x90:
-                q.xy.operations["x90"].amplitude = new_amplitude / 2
+            reference_pulse.amplitude = new_amplitude
```
KU-QM: Why not update the x90 if pulse is x180? Could be more consistent with the superconducting version

qualibration_graphs/quantum_dots/calibrations/loss_divincenzo/08a_power_rabi.py
```diff
@@ -264,11 +257,10 @@ def update_state(node: QualibrationNode[Parameters, Quam]):
                 continue
 
             opt_prefactor = node.results["fit_results"][q.name]["opt_amp"]
```
KU-QM, line 259: I think maybe I need a bit of clarification. In this update state, you find the referenced pulse using the new fn

``` def get_xy_reference_pulse_name(qubit: AnySpinQubit) -> str:
    """Resolve the pulse name backing the qubit's default XY macros."""
    if qubit.xy is None:
        raise ValueError(f"Qubit '{qubit.id}' has no XY drive configured.")

    xy_drive_macro = qubit.macros.get(SingleQubitMacroName.XY_DRIVE)
    if xy_drive_macro is None:
        raise KeyError(f"Qubit '{qubit.id}' is missing the '{SingleQubitMacroName.XY_DRIVE}' macro.")

    pulse_name = getattr(xy_drive_macro, "reference_pulse_name", None)
    if pulse_name is None:
        raise ValueError(f"Qubit '{qubit.id}' XY-drive macro has no reference_pulse_name configured.")
    if pulse_name not in qubit.xy.operations:
        raise KeyError(f"Reference pulse '{pulse_name}' is not defined on qubit '{qubit.id}' XY drive.")

    return pulse_name```

So all the macros use the same pulse under the hood? In that case, shouldn't the update be for the macro object, which can then rescale the underlying pulse when played? The update state at the moment seems to update the pulse directly

KU-QM (approved):
definitely necessary to remove - let's revisit the simultaneous stuff at some point. I have some points that I need clarification

qualibration_graphs/quantum_dots/calibrations/loss_divincenzo/08a_power_rabi.py
sebastianorbell-qm: they are all connected to the same calibrated waveform - so you do not need to update them separately

qualibration_graphs/quantum_dots/calibrations/loss_divincenzo/10_T1_parity_diff.py
sebastianorbell-qm: you are correct

qualibration_graphs/quantum_dots/calibrations/loss_divincenzo/09a_time_rabi_parity_diff.py
sebastianorbell-qm: yes - I am thinking of the best way to expose the duration of the gaussian to the macros - then i will update this

qualibration_graphs/quantum_dots/calibrations/loss_divincenzo/08a_power_rabi.py
sebastianorbell-qm: I will try to simplify it

PR #505, opened 2026-03-25
Adjust detuning value scaling in calibration script

Kyung-hoon-Jung0:
hot fix for the detuning value in "else" statement


paulQM (approved):
Nice catch

PR #507, opened 2026-03-29
Fixed wrong units in plotting. Nanoseconds instead of microseconds.

Tomer-Firon:


qualibration_graphs/superconducting/calibration_utils/T2echo/plotting.py
```diff
@@ -84,11 +84,11 @@ def plot_individual_data_with_fit(ax: Axes, ds: xr.Dataset, qubit: dict[str, str
         raise RuntimeError("The dataset must contain either 'I' or 'state' for the plotting function to work.")
 
     ax.set_title(qubit["qubit"])
-    ax.set_xlabel("Idle_time (µs)")
+    ax.set_xlabel("Idle_time (ns)")
     ax.text(
         0.1,
         0.9,
-        f'T2e = {fit["T2_echo"].values*1e-3:.1f} ± {fit["T2_echo_error"].values*1e-3:.1f} µs',
+        f'T2e = {fit["T2_echo"].values*1e-3:.1f} ± {fit["T2_echo_error"].values*1e-3:.1f} ns',
```
TheoLaudatQM: Here you can see that the value of T2echo is multiplied by 1e-3 so the unit should be µs.

qualibration_graphs/superconducting/calibration_utils/T2echo/plotting.py
```diff
@@ -84,11 +84,11 @@ def plot_individual_data_with_fit(ax: Axes, ds: xr.Dataset, qubit: dict[str, str
         raise RuntimeError("The dataset must contain either 'I' or 'state' for the plotting function to work.")
 
     ax.set_title(qubit["qubit"])
-    ax.set_xlabel("Idle_time (µs)")
+    ax.set_xlabel("Idle_time (ns)")
```
TheoLaudatQM, line 87: Since T2echo should always be on the order of several µs, I am in favor of multiplying the idle time by 1e-3 when plotting instead and keep the unit in µs.

TheoLaudatQM (changes requested):

qualibration_graphs/superconducting/calibration_utils/T2echo/plotting.py
Tomer-Firon: Discussed offline — keeping units as ns here since the plotted values are consistent with this choice.

TheoLaudatQM (approved):

PR #512, opened 2026-04-20
Chore/quam config sc docs

KevinAVR:
qua-libs now adapted to run with IQCC. The modifications are in docs and qualang-tools compatibilty

Does not break any path of working directly with an OPX

### DIFF
--- a/qualibration_graphs/superconducting/pyproject.toml
+++ b/qualibration_graphs/superconducting/pyproject.toml
@@ -12,11 +12,11 @@ classifiers = [
 ]
 dependencies = [
     "quam>=0.5.0a1",
-    "qualang-tools>=0.19.0",
+    "qualang-tools>=0.21.0,<0.22.0",
     "qualibrate>=1.0.2",
     "qm-qua>=1.2.1",
     "tomli>=2.0.1; python_version < '3.11'",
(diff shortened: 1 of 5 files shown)


qualibration_graphs/superconducting/pyproject.toml
```diff
@@ -12,11 +12,11 @@ classifiers = [
 ]
 dependencies = [
     "quam>=0.5.0a1",
-    "qualang-tools>=0.19.0",
+    "qualang-tools>=0.21.0,<0.22.0",
```
TheoLaudatQM, line 15: Same as in quam-builder, this will break the compatibility with python 3.13 which was requested by some customers

qualibration_graphs/superconducting/pyproject.toml
```diff
@@ -12,11 +12,11 @@ classifiers = [
...
     "qualibrate>=1.0.2",
     "qm-qua>=1.2.1",
     "tomli>=2.0.1; python_version < '3.11'",
-    "quam-builder@git+https://github.com/qua-platform/quam-builder.git",
+    "quam-builder @ git+https://github.com/qua-platform/quam-builder.git@main",
```
TheoLaudatQM, line 19: Is `main` really needed here?

KevinAVR:
already stablished path via qualibrate

PR #515, opened 2026-05-08
Resonator frequency vs coupler flux calibration 

Deepakkhurrana:
Summary:

1. Adds 02d_resonator_spectroscopy_vs_coupler_flux.py, a new calibration node that measures resonator frequency as a function of coupler flux bias for all qubit pairs.
2. Adds supporting utilities under calibration_utils/resonator_spectroscopy_vs_coupler_flux/ (analysis, plotting, parameters) and calibration_utils/pair_grid.py for topology-aware coupler subplot layout.

Why:

Resonator frequency shifts with coupler flux, so calibrating this dependence is a prerequisite for operating qubits at different coupler flux points.

Implementation notes:

Plotting uses a custom QubitPairGrid that positions coupler subplots on a doubled grid derived from control/target qubit grid_location, giving a topology-aware chip layout view. This object should move to qualibration_libs eventualy

PR #516, opened 2026-05-11
Qubit frequency vs coupler flux calibration

Deepakkhurrana:
Summary:

1. Adds 03c_qubit_spectroscopy_vs_coupler_flux.py, a new calibration node that sweeps qubit saturation spectroscopy over coupler flux bias for all qubit pairs, mapping out qubit frequency vs coupler flux and detecting avoided crossings.

2. Adds supporting utilities under calibration_utils/qubit_spectroscopy_vs_coupler/ (analysis, plotting, parameters).

Why:

Qubit frequency shifts with coupler flux due to dispersive coupling between the qubit and coupler modes. Measuring this dependence and locating avoided crossings is necessary to characterise idle and interaction points of the coupler, and to understand how much the qubit frequency is pulled at different coupler operating points.

Implementation notes:

1. Avoided crossing detection uses a derivative-based approach on the smoothed peak frequency curve, followed by a local hyperbolic fit with EM-style branch assignment to extract the crossing flux position, bare frequencies, and coupling strength g.
2. Plotting uses QubitPairGrid (from calibration_utils/pair_grid.py) to position coupler subplots in a topology-aware chip layout view — same approach as 02d (PR #515). This object should eventually move to qualibration_libs.

### DIFF
--- a/qualibration_graphs/superconducting/calibration_utils/qubit_spectroscopy_vs_coupler/parameters.py
+++ b/qualibration_graphs/superconducting/calibration_utils/qubit_spectroscopy_vs_coupler/parameters.py
@@ -1,0 +1,51 @@
+"""Parameter definitions for qubit spectroscopy vs coupler flux calibration."""
+
+from typing import ClassVar, Literal, Optional
+
+from qualibrate import NodeParameters
+from qualibrate.core.parameters import RunnableParameters
+from qualibration_libs.parameters import CommonNodeParameters, QubitPairExperimentNodeParameters
+
+
+class NodeSpecificParameters(RunnableParameters):
+    """Node-specific parameters for qubit spectroscopy vs coupler flux calibration."""
+
+    num_shots: int = 50
+    """Number of averages to perform. Default is 50."""
+    operation: str = "saturation"
+    """Operation to perform. Default is "saturation"."""
+    operation_amplitude_factor: float = 0.1
+    """Amplitude factor for the operation. Default is 0.1."""
+    operation_len_in_ns: Optional[int] = None
+    """Length of the operation in ns. Default is the predefined pulse length."""
+    frequency_span_in_mhz: float = 100.0
+    """Frequency span in MHz. Default is 100 MHz."""
+    frequency_step_in_mhz: float = 0.5
+    """Frequency step in MHz. Default is 0.5 MHz."""
+    min_flux_in_v: float = -0.025
+    """Minimum flux bias in volts. Default is -0.025 V."""
+    max_flux_in_v: float = 0.025
+    """Maximum flux bias in volts. Default is 0.025 V."""
+    num_flux_points: int = 11
+    """Number of flux points. Default is 11."""
+    input_line_impedance_in_ohm: Optional[int] = 50
+    """Input line impedance in ohms. Default is 50 Ohm."""
+    line_attenuation_in_db: Optional[int] = 0
+    """Line attenuation in dB. Default is 0 dB."""
+    measure_qubit: Literal["control", "target"] = "target"
+    """Which qubit to measure: 'control' or 'target'. Default is 'target'."""
+    settle_time_in_ns: int = 20000
+    """Settle time in ns. Default is 5000 ns."""
+    buffer_time_in_ns: int = 100
+    """Buffer time before readout in ns. Default is 100 ns."""
+
+
+class Parameters(
+    NodeParameters,
+    CommonNodeParameters,
+    NodeSpecificParameters,
+    QubitPairExperimentNodeParameters,
+):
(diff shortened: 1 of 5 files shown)


qualibration_graphs/superconducting/calibration_utils/pair_grid.py
paulQM, line 1: is this file repeated from #515 ?

paulQM (changes requested):
just clarify the comment, the rest looks good

qualibration_graphs/superconducting/calibration_utils/qubit_spectroscopy_vs_coupler/parameters.py
```diff
@@ -0,0 +1,51 @@
...
+    num_flux_points: int = 11
+    """Number of flux points. Default is 11."""
+    input_line_impedance_in_ohm: Optional[int] = 50
+    """Input line impedance in ohms. Default is 50 Ohm."""
+    line_attenuation_in_db: Optional[int] = 0
+    """Line attenuation in dB. Default is 0 dB."""
+    measure_qubit: Literal["control", "target"] = "target"
+    """Which qubit to measure: 'control' or 'target'. Default is 'target'."""
+    settle_time_in_ns: int = 20000
+    """Settle time in ns. Default is 5000 ns."""
```
sam-vizvary-QM, line 38: Default is 20000 or 5000?

qualibration_graphs/superconducting/calibrations/1Q_calibrations/03c_qubit_spectroscopy_vs_coupler_flux.py
```diff
@@ -0,0 +1,343 @@
...
+                                "const",
+                                amplitude_scale=dc / qp.coupler.operations["const"].amplitude,
+                                duration=duration + settle_t + buffer_t,
+                            )
+                            measured_qubit.xy.wait(settle_t)
+                            # Apply saturation pulse to the measured qubit
+                            measured_qubit.xy.play(
+                                operation,
+                                amplitude_scale=operation_amp,
+                                duration=duration,
```
sam-vizvary-QM: I think there’s a correctness issue with duration here. It’s computed inside the first loop over multiplexed_qubit_pairs, but then reused later in the pulse loop as a single variable. If operation_len_in_ns is None and different measured qubits have different configured operation lengths, every pair in the batch will end up using the duration from the last qubit processed in the previous loop.

sam-vizvary-QM: I think we want to store duration in an array and then call it using [ii]. Maybe we should discuss.

qualibration_graphs/superconducting/calibration_utils/pair_grid.py
Deepakkhurrana: Yes, I will rebase on main after 515 is merged.

qualibration_graphs/superconducting/calibration_utils/qubit_spectroscopy_vs_coupler/parameters.py
Deepakkhurrana: Ah! I think 5000ns is reasonable default. I will update accordingly.

qualibration_graphs/superconducting/calibrations/1Q_calibrations/03c_qubit_spectroscopy_vs_coupler_flux.py
Deepakkhurrana: Fair point. I think as you suggested, following block after `measured_qubits_map` should fix it. What do you think?  
`operation_durations = {
                ii: (
                    operation_len * u.ns
                    if operation_len is not None
                    else measured_qubits_map[ii].xy.operations[operation].length * u.ns
                ) // 4
                for ii in multiplexed_qubit_pairs
            }`

sam-vizvary-QM: Yes this looks good to me.

Deepakkhurrana: the change is implemented now.

Deepakkhurrana:
All the comments have been addressed.

qualibration_graphs/superconducting/calibration_utils/pair_grid.py
Deepakkhurrana: After rebasing on main, this file is removed now.

paulQM (approved):

sam-vizvary-QM (approved):
Looks good!

PR #517, opened 2026-05-19
Ramsey vs coupler flux calibration node

Deepakkhurrana:
**Summary**

New
- Adds a new qualibration node and utils to characterize how tunable-coupler flux shifts the measured qubit frequency via Ramsey interferometry, following the same patterns as 03c. 
- Adds supporting utilities under calibration_utils/ramsey_vs_coupler_flux/ (analysis, plotting, parameters).

Changed
- Renames the existing Ramsey-vs-flux node to 09a and updates bringup/retuning graphs accordingly.

**Why**
Coupler flux shifts the qubit frequency via dispersive coupling. We need that curve to pick operating points and track dispersion calibrations. 03c gets it with spectroscopy; 09b gets it with Ramsey. It pairs with 09a (Ramsey vs qubit flux) for a consistent flux-characterization set on tunable-coupler devices.

**Notes for reviewers**
- Complements 03c (spectroscopy vs coupler flux) with Ramsey-based frequency tracking vs flux.
- No parabola / avoided-crossing fit on the frequency curve (oscillation fit per flux point only).
- Optional follow-ups: shared helper for measured-qubit namespace setup; node.outcomes gating in update_state (like 03c).


paulQM (approved):
Looks great!

PR #518, opened 2026-05-21
qubit_xy-coupler_z delay calibration

Deepakkhurrana:
**Summary**

New

Adds a new qualibration node and utils to correct the delay between qubit XY and coupler Z channel. 
Adds supporting utilities under calibration_utils/xyc_delay/ (analysis, plotting, parameters).

Changed

Renames the existing 16_xyz_delay node to 16a. 

Why
The goal is to ensure that the coupler flux pulse reaches the qubit at the same time as the XY drive and correct
for any latency differences between the two control lines. By applying the XY and coupler Z pulses simultaneously, the qubit
frequency is shifted during the XY rotation, which can lead to incomplete rotations if the pulses are not properly aligned
in time.

Notes for reviewers

Optional follow-ups: Sweep coupler flux around anti crossing point to find strong coupling point where analysis does not fail.


paulQM (changes requested):
## Plotting bug (`calibration_utils/xyc_delay/plotting.py`)

In `plot_individual_data_with_fit`, the curve is plotted from `fit_q` instead of `ds_q`:

```python
def plot_individual_data_with_fit(ax, ds_q, fit_q=None):
    fit_q.difference.plot(ax=ax)  # should be ds_q.difference.plot(ax=ax)
    if fit_q is not None and fit_q.success.data:
        ax.axvline(fit_q.flux_delay.data, ...)
```

**Issues:**
- `ds_q` is passed in and documented as the data to plot, but is never used for the line plot.
- The `if fit_q is not None` guard only protects the fitted-delay vline; if `fits` were ever `None`, the first line would raise before that check.
- On the normal path, `fit_q.difference` likely matches `ds_q.difference`, so this may not show up in routine runs—but it is still the wrong object and makes `ds_q` misleading.

**Suggested fix:**
```python
ds_q.difference.plot(ax=ax)
if fit_q is not None and fit_q.success.data:
    ax.axvline(fit_q.flux_delay.data, color="red", linestyle="--", label="fitted center")
    ax.legend()
```

paulQM:
**Quick thought on the fit (`xyc_delay/analysis.py`)**

Right now `fit_routine` is basically the "v0" approach I threw together: look for exactly two sign flips in `difference` and take the midpoint. Works fine when you get a clean e−g dip, but it's pretty easy to break—noisy traces, values sitting near zero (`np.sign(0)` fun), or flux not at a good coupling point and you hit the assert and land on `flux_delay=0` / `success=False`.

Wondering if this PR (or a tiny follow-up right after) is a good moment to make that a bit sturdier **and** share it with **16a** XY–qubit Z delay too? Same overall picture for both nodes (e−g difference vs `relative_time`), so something like light smoothing before sign detection, or a simple fallback when you don't get exactly two crossings, could live in one place instead of us carrying two copies of the quick version (`xyc_delay` + `xyx_delay`).

Totally fine to ship as-is and iterate—just wanted to flag it while we're already in this code.

Deepakkhurrana:
> **Quick thought on the fit (`xyc_delay/analysis.py`)**
> 
> Right now `fit_routine` is basically the "v0" approach I threw together: look for exactly two sign flips in `difference` and take the midpoint. Works fine when you get a clean e−g dip, but it's pretty easy to break—noisy traces, values sitting near zero (`np.sign(0)` fun), or flux not at a good coupling point and you hit the assert and land on `flux_delay=0` / `success=False`.
> 
> Wondering if this PR (or a tiny follow-up right after) is a good moment to make that a bit sturdier **and** share it with **16a** XY–qubit Z delay too? Same overall picture for both nodes (e−g difference vs `relative_time`), so something like light smoothing before sign detection, or a simple fallback when you don't get exactly two crossings, could live in one place instead of us carrying two copies of the quick version (`xyc_delay` + `xyx_delay`).
> 
> Totally fine to ship as-is and iterate—just wanted to flag it while we're already in this code.

Deepakkhurrana:
> **Quick thought on the fit (`xyc_delay/analysis.py`)**
> 
> Right now `fit_routine` is basically the "v0" approach I threw together: look for exactly two sign flips in `difference` and take the midpoint. Works fine when you get a clean e−g dip, but it's pretty easy to break—noisy traces, values sitting near zero (`np.sign(0)` fun), or flux not at a good coupling point and you hit the assert and land on `flux_delay=0` / `success=False`.
> 
> Wondering if this PR (or a tiny follow-up right after) is a good moment to make that a bit sturdier **and** share it with **16a** XY–qubit Z delay too? Same overall picture for both nodes (e−g difference vs `relative_time`), so something like light smoothing before sign detection, or a simple fallback when you don't get exactly two crossings, could live in one place instead of us carrying two copies of the quick version (`xyc_delay` + `xyx_delay`).
> 
> Totally fine to ship as-is and iterate—just wanted to flag it while we're already in this code.

I have added now a more robust fit analysis. Main improvements are:

- Replaced the zero-crossing logic entirely with a triangle peak fit (scipy.curve_fit). The triangle is the physically correct model — it's the cross-correlation of the two equal-duration rectangular pulses. No sign changes, no np.sign(0) edge cases.
- The offset parameter in the model absorbs the baseline directly, so traces sitting far from zero are handled naturally without any upfront subtraction.
- Added four explicit quality checks: edge proximity (triangle truncated at scan boundary), fit uncertainty from the covariance diagonal, amplitude polarity, and SNR against the wing noise. Each fires with a descriptive message rather than landing silently on flux_delay=0.
- The fallback on failure is now a baseline-subtracted argmax rather than zero — still a physically grounded estimate, and success=False is set so downstream code knows to treat it with suspicion. 

You can see the results below. 

About combining the utils
- I think we can not combine all the code since we have have different data structures for qubit and coupler nodes. However, some redundancy can be reduced by f`rom calibration_utils.xyx_delay.analysis import log_fitted_results, _extract_relevant_fit_parameters, triangle_peak`.

image
Screenshot 2026-05-22 at 15 11 56

paulQM (approved):
LGTM

qualibration_graphs/superconducting/calibration_utils/xyc_delay/analysis.py
```diff
@@ -0,0 +1,140 @@
...
+    if data == "I":
+        difference -= difference.mean()
+
+    ds["difference"] = difference
+    return ds
+
+
+def fit_raw_data(ds: xr.Dataset, node: QualibrationNode) -> Tuple[xr.Dataset, dict[str, FitParameters]]:
+    """
+    Fit the qubit frequency and FWHM for each qubit in the dataset.
```
cjwu-qm: This description seems to be not accurate

qualibration_graphs/superconducting/calibration_utils/xyx_delay/analysis.py
```diff
@@ -93,26 +95,74 @@ def _extract_relevant_fit_parameters(fit: xr.Dataset, node: QualibrationNode):
...
+def triangle_peak(t, amp, t0, half_width, offset):
+    """
+    Cross-correlation of two equal-duration rectangular pulses.
+    Both XY and coupler flux pulses have the same length by construction
+    (flux waveform is passed in already matching x180 length).
+    """
+    return amp * np.maximum(0, 1 - np.abs(t - t0) / half_width) + offset
 
+
+def fit_routine(da, node):
```
cjwu-qm: I see that fit routines are repeated in both files (xyx and xyc). This means that in the future if something is changed in the fitting, we need to remember to change both. Is there a better way to do this in qua-libs? (I am not sure what is the answer and this should not block the merge, but just curious)

cjwu-qm (approved):
Looks good to me, just have some minor comments. My iqcc token expired so weren't able to test it on a real machine

qualibration_graphs/superconducting/calibration_utils/xyc_delay/analysis.py
Deepakkhurrana: Thanks for noticing! Corrected now.

qualibration_graphs/superconducting/calibration_utils/xyx_delay/analysis.py
Deepakkhurrana: I think it is a good point. I have done following changes now: 

- Rename calibration_utils `xyx_delay` → `xyz_delay` to match XY-Z delay
  semantics and `16a_xyz_delay.py`; remove legacy `xyx_delay` package from qua-libs
- Point `16a_xyz_delay` and `xyc_delay` imports at `xyz_delay`
- Extract shared `fit_delay_trace(da, xy_duration)` in `xyz_delay/analysis.py`;
  keep thin `fit_routine` wrappers in `xyz_delay` (qubit id) and `xyc_delay`
  (measured_qubit_name) for experiment-specific qubit lookup
- Reuse `log_fitted_results`, `_extract_relevant_fit_parameters`, and
  `triangle_peak` from `xyz_delay` in `xyc_delay` where applicable

paulQM:
@Deepakkhurrana should we merge this?

PR #519, opened 2026-05-27
Decoupling flux calibration via zero of residual ZZ coupling using JAZZ sequence 

Deepakkhurrana:
**What**
- Adds qualibration node `19_zz_off_jazz` and utils package `calibration_utils/JAZZ_ZZ/`.
- Measures residual ZZ coupling by sweeping coupler amplitude and interaction time in a JAZZ echo sequence (`x90 → coupler → π echo → virtual Z → coupler → x90`).
- Fits damped cosines per amplitude to extract `J_eff`, decay rate, and updates `qubit_pair.coupler.decouple_offset` from the fitted bias point.

**Why** 
- Tunable-coupler QPUs need a dedicated step to minimize residual ZZ at the operating point before two-qubit calibrations.

**Notes for reviewers**
-  Aligns with other coupler-flux nodes (03c, 09b) for structure and plotting, so review and maintenance stay consistent.
-  non-deterministic `align` used in favour of explicit `waits`
- Coupling and decay figures can do both linear and log plots

### DIFF
--- a/qualibration_graphs/superconducting/calibrations/1Q_calibrations/19_zz_off_jazz.py
+++ b/qualibration_graphs/superconducting/calibrations/1Q_calibrations/19_zz_off_jazz.py
@@ -1,28 +1,24 @@
 # %% {Imports}
 from dataclasses import asdict
 
-import matplotlib.pyplot as plt
 import numpy as np
 import xarray as xr
 from calibration_utils.JAZZ_ZZ import (
-    FitResults,
     Parameters,
     fit_raw_data,
     log_fitted_results,
     plot_decay_rate_data,
-    plot_oscillation_data,
-    plot_raw_data_with_fit,
+    plot_effective_coupling,
+    plot_fit_data,
+    plot_raw_data,
     process_raw_dataset,
 )
-from qm import SimulationConfig
 from qm.qua import *
-from qualang_tools.bakery import baking
 from qualang_tools.loops import from_array
 from qualang_tools.multi_user import qm_session
-from qualang_tools.results import fetching_tool, progress_counter
+from qualang_tools.results import progress_counter
 from qualang_tools.units import unit
 from qualibrate import QualibrationNode
-from qualibration_libs.core import tracked_updates
 from qualibration_libs.data import XarrayDataFetcher
 from qualibration_libs.parameters import get_qubit_pairs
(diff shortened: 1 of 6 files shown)


paulQM (approved):
LGTM

qualibration_graphs/superconducting/calibrations/1Q_calibrations/19_zz_off_jazz.py
```diff
@@ -0,0 +1,334 @@
...
+                            partner_qubit.xy.wait(t)
+                            measured_qubit.xy.play("x180")
+                            partner_qubit.xy.play("x180")
+                            qp.coupler.wait(measured_qubit.xy.operations["x180"].length // 4)
+                            measured_qubit.xy.frame_rotation_2pi(virtual_detuning_phase)
+                            qp.coupler.play(
+                                "const",
+                                amplitude_scale=amp / qp.coupler.operations["const"].amplitude,
+                                duration=t,
+                            )
```
HuanQM, line 175: Here, the sequence plays two coupler pulses, each with a duration of t, so the total coupler pulse time is 2t. However, the analysis (fits, and plots) and the virtual detuning phase both use t. Should these be made consistent?

Deepakkhurrana: Good catch! I double checked and I think there is no physics issue. The code is self-consistent — both the ZZ phase and the detuning phase scale with the same t, so the extraction ζ = ωm − ωb is clean. I have updated the analysis.py to reflect that. 
@paulQM, do you think there is an issue?

paulQM: yes, t is the ZZ interaction time. The JAZZ sequence plays it twice to have the echo effect of substracting the common phases. The phase aquired on the first ramsey arm is refocussed in the second arm. The remaining phase is the one acquired by exciting the other qubit, i.e. the one introduced by the ZZ coupling.

HuanQM: Got it!
And the description you added to analysis.py makes this very clear. Thanks!

HuanQM (approved):
LGTM

PR #520, opened 2026-05-27
fix/add readout_GEF setup to GEF calibration nodes 14 and 15

Deepakkhurrana:
**What**

- `14_gef_readout_frequency_optimization` and `15_iq_blobs_gef`: at the start of `create_qua_program`, create `resonator.operations["readout_GEF"]` from readout when it is missing (1.5× length, thresholds cleared), matching the pattern used in the legacy GEF nodes.
-` calibration_utils/readout_gef_frequency_optimization/` and `calibration_utils/iq_blobs_ef/`: supporting updates for these nodes (parameters, analysis, plotting).
- Small fixes in the node scripts (e.g. data_fetcher.get("n", 0), docstrings, pylint).

**Why**
- After the `quam-builder` upgrade (https://github.com/qua-platform/quam-builder/pull/116), new machines get both `readout` and `readout_GEF` from `add_default_transmon_pulses()`. Older saved QuAM states often only have readout.
- Nodes 14 and 15 are the first calibrations that measure with readout_GEF (three-level g/e/f readout). On a legacy state that would cause: `KeyError: 'readout_GEF'`

- The update lets 14 and 15 run on old states (pulse created in memory, then saved) and on new populates (pulse already there—unchanged). After 14 → 15 and save, downstream nodes that use readout_state_gef() (e.g. CZ / coupler calibrations) can find `readout_GEF` and `gef_centers` in the state.

### DIFF
--- a/qualibration_graphs/superconducting/calibrations/1Q_calibrations/14_gef_readout_frequency_optimization.py
+++ b/qualibration_graphs/superconducting/calibrations/1Q_calibrations/14_gef_readout_frequency_optimization.py
@@ -1,4 +1,5 @@
 # %% {Imports}
+import dataclasses
 from dataclasses import asdict
 
 import matplotlib.pyplot as plt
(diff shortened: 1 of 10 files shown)


qualibration_graphs/superconducting/calibrations/1Q_calibrations/14_gef_readout_frequency_optimization.py
```diff
@@ -97,6 +94,18 @@ def create_qua_program(node: QualibrationNode[Parameters, Quam]):
     node.namespace["qubits"] = qubits = get_qubits(node)
     num_qubits = len(qubits)
 
+    for qubit_obj in qubits:
```
paulQM, line 97: nice addition!

paulQM (approved):
Nice

PR #526, opened 2026-06-12
Introduce QubitRoles for physics-driven qubit role resolution in 2Q gate calibrations

Deepakkhurrana:
**Summary**

Replaces the ad-hoc `moving_qubit` / `other_qubit` helper functions (which relied on the `qubit_pair.moving_qubit` config field) with a new `QubitRoles` named tuple that resolves qubit roles — moving, stationary, leakage, high, low — from first principles using qubit frequencies and anharmonicities.

**Why**

Previously, which qubit "moves" during a flux pulse was determined by a manually-set moving_qubit: "target" | "control" attribute on the pair. This was error-prone and required the user to know the correct role assignment upfront. The new `QubitRoles.resolve() `classmethod derives the correct role automatically:

iSWAP: the higher-frequency qubit tunes down to the lower one.
CZ (via |11⟩↔|20⟩): if freq_diff > |anharmonicity_high|, the high-frequency qubit moves; otherwise the low-frequency qubit moves. The leakage qubit is always the high-frequency one.

**Changes**
- `parameters.py:` Added `QubitRoles(NamedTuple)` with a `resolve()` classmethod and a `verify_moving_qubit()` helper that cross-checks the resolved role against the legacy qubit_pair.moving_qubit field.

- `__init__.py`: Exported `QubitRoles` and `verify_moving_qubit`.

- `analysis.py`: Renamed all control / target variable references to moving / stationary throughout the analysis pipeline (interaction map, fit extraction, dataset coordinates). Removed the direct import of the old `moving_qubit` function; roles are now read from `node.namespace["qubit_roles_map"]`.

- `30_cz_iswap_flux_bootstrap.py (node 30)`: Builds a `qubit_roles_map` in the setup step, passes it through the namespace and persists it in `node.results `so historical re-analysis is fully reproducible. QUA stream variables renamed from `state_control / state_target / I_control / I_target `to `state_moving / state_stationary / I_moving / I_stationary`.

**Reviewer Notes**
- No changes to the calibration logic or sweep parameters — this is purely a naming/role-resolution refactor.
- After this PR is approved, similar changes will be done on 31, 32a, 32b.

### DIFF
--- a/qualibration_graphs/superconducting/calibration_utils/cz_iswap_flux_bootstrap/parameters.py
+++ b/qualibration_graphs/superconducting/calibration_utils/cz_iswap_flux_bootstrap/parameters.py
@@ -160,14 +160,26 @@ class QubitRoles(NamedTuple):
 def verify_moving_qubit(
     qp,
     gate_type: Literal["cz", "iswap"] = "cz",
+    operation: str = "cz_unipolar",
+    repair_routing: bool = False,
     log_callable: Optional[Callable[[str], None]] = None,
 ) -> None:
-    """Verify that qp.moving_qubit matches the physics-based calculation; update in-place if not.
+    """Verify and reconcile ``qp.moving_qubit``, then ``macros[operation]`` flux Z-line routing.
 
-    Recalculates the moving qubit from qubit frequencies and anharmonicity, then compares
-    it to the ``moving_qubit`` field stored on the QUAM qubit-pair object. If they disagree,
-    logs a warning, updates ``qp.moving_qubit`` in memory, and relies on the caller's state save
-    to persist the correction to state json.
+    Recalculates the moving qubit from qubit frequencies and anharmonicity via
+    :func:`QubitRoles.resolve`, updates ``qp.moving_qubit`` on the pair if needed, then
+    checks that ``macros[operation].flux_pulse_qubit`` is registered on that qubit's Z
+    channel. ``CZGate.apply()`` reads ``qp.moving_qubit`` to pick the Z line, so the
+    label is corrected before pulse routing is verified or repaired.
+
+    Flux routing is verified using ``flux_pulse_qubit.id`` as the Z-line operation name
+    (the same name passed to ``moving_qubit.z.play(...)`` at runtime). If the pulse is
+    missing from the moving qubit or present only on the stationary qubit's Z line,
+    raises :class:`ValueError` unless ``repair_routing=True``.
+
+    With ``repair_routing=True``, a new Z-line entry is created on the moving qubit and wired
+    to ``macros[operation].flux_pulse_qubit`` (same pattern as ``populate_quam``); any existing
+    entry on the stationary Z line is left unchanged.
 
     Parameters
     ----------
@@ -176,9 +188,21 @@ def verify_moving_qubit(
     gate_type:
         ``"cz"`` (default) or ``"iswap"`` — selects the physics rule used by
         :func:`QubitRoles.resolve`.
+    operation:
+        Pair macro to verify (e.g. ``"cz_unipolar"``), matching ``node.parameters.operation``.
+    repair_routing:
+        If True, add the flux pulse on the moving Z line instead of raising.
     log_callable:
         Callable that accepts a single string and emits it to the node's output
         (e.g. ``node.log``). Defaults to ``logging.getLogger(__name__).info``.
+
+    Raises
+    ------
+    ValueError
+        If ``operation`` is not in ``qp.macros``, if its flux pulse is not on the moving
+        qubit's Z line (and ``repair_routing`` is False), if the moving qubit has no Z
+        channel, or if ``qp.moving_qubit`` is set to a value other than ``"control"`` or
+        ``"target"``.
     """
     if log_callable is None:
         log_callable = logging.getLogger(__name__).info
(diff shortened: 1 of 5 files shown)


qualibration_graphs/superconducting/calibration_utils/cz_iswap_flux_bootstrap/parameters.py
```diff
@@ -107,26 +107,171 @@ class Parameters(
...
+        )
+        qp.moving_qubit = computed_label
+    elif stored != computed_label:
+        log_callable(
+            f"WARNING Pair {qp.name}: moving={stored_mq.name}, stationary={stored_sq.name} "
+            f"disagrees with recalculation: moving={roles.moving.name}, stationary={roles.stationary.name} "
+            f"for {gate_type} gate — updating. "
+            f"State will be persisted at the end of the node. [{physics_ctx}]"
+        )
+        qp.moving_qubit = computed_label
```
paulQM, line 248: When `stored != computed_label`, this updates the `moving_qubit` label in memory, but gate macros are wired at QUAM population time: `flux_pulse_qubit` operations live on a specific qubit's Z line (see `populate_quam_lf_mw_fems.py`). Meanwhile node 30 now drives prep/readout from `QubitRoles.resolve()`, and `qp.macros[operation].apply(...)` still uses whatever Z-line routing was baked in at populate.

Should `verify_moving_qubit()` also verify that the expected flux pulses are present on the resolved moving qubit's Z line? And if they are on the wrong qubit, should it relocate/add them (or at least raise a hard error rather than silently "fixing" only the string field)?

qualibration_graphs/superconducting/calibration_utils/cz_iswap_flux_bootstrap/parameters.py
```diff
@@ -107,26 +107,171 @@ class Parameters(
...
+    delta_mhz = (roles.high.f_01 - roles.low.f_01) / 1e6
+    alpha_mhz = roles.high.anharmonicity / 1e6
+    physics_ctx = (
+        f"high={roles.high.name} ({roles.high.f_01/1e9:.4f} GHz), "
+        f"low={roles.low.name} ({roles.low.f_01/1e9:.4f} GHz), "
+        f"δ={delta_mhz:.1f} MHz, α_high={alpha_mhz:.1f} MHz"
+    )
+
+    stored = getattr(qp, "moving_qubit", None)
+    stored_mq = qp.qubit_control if stored == "control" else qp.qubit_target
```
paulQM: Any value other than the literal string `"control"` is treated as `"target"` here (including `None` after the earlier branch, typos, or corrupted state). That means a bad `moving_qubit` value produces misleading `stored_mq` / `stored_sq` names in the warning log without surfacing that the field itself is invalid.

Should this explicitly validate that `stored in ("control", "target")` and error (or at least warn) on unexpected values before mapping to qubits?

paulQM (commented):
Two logic questions on `verify_moving_qubit()` — see inline comments.

qualibration_graphs/superconducting/calibration_utils/cz_iswap_flux_bootstrap/parameters.py
Deepakkhurrana: Thanks for the comment!  Agreed. Updating `qp.moving_qubit `alone doesn't fix where the pulse lives; `CZGate.apply() `plays `flux_pulse_qubit.id` on whichever qubit the label points to, while prep/readout follow `QubitRoles.resolve()`.

`verify_moving_qubit() `now:

- Resolves roles from physics and updates `qp.moving_qubit` first (label before Z-line work, since macros read the label at runtime).
- Checks `macros[operation].flux_pulse_qubit.id` is registered on that moving qubit's z.operations.
- `repair_routing=True`: adds a Z-line entry wired to the macro (length/amplitude/… refs, same pattern as populate). Does not remove the pulse from the stationary qubit if it was there.
- `repair_routing=False`: raises ValueError with an actionable message instead of silently fixing only the label.

qualibration_graphs/superconducting/calibration_utils/cz_iswap_flux_bootstrap/parameters.py
Deepakkhurrana: Agreed. Silent fallback was misleading. We now validate before mapping to qubits:

`stored` is `None` → warn and set from physics (unchanged).
`stored` not in ("control", "target") → `ValueError` with the invalid value quoted (e.g. typo 'contorl').
`stored_mq` / `stored_sq` are only computed inside the stale-label branch, where stored is already known valid.

So corrupted state fails fast instead of producing wrong qubit names in a warning log.

qualibration_graphs/superconducting/calibration_utils/cz_iswap_flux_bootstrap/parameters.py
paulQM: Regarding:

> repair_routing=True: adds a Z-line entry wired to the macro (length/amplitude/… refs, same pattern as populate). Does not remove the pulse from the stationary qubit if it was there.

Do you think we should remove the dead pulses? Any risk of cross-reference in the nodes?

Deepakkhurrana: I was thinking about it as well. Not sure tbh. I did not remove the stationary Z-line pulse during `repair_routing` because it could be intentional from user during populate for some other tests/purpose.

paulQM (approved):

HuanQM (approved):
LGTM

PR #528, opened 2026-06-15
Refactor CZ chevron (node 31) for QubitRoles and tunable-coupler support

Deepakkhurrana:
## Summary

Refactors the CZ chevron calibration (node 31) to use physics-driven **moving/stationary qubit roles** instead of hardcoded control/target, and adds **tunable-coupler** support. Renames the node file to `31_chevron_11_20.py` to reflect the |11⟩↔|20⟩ interaction.

## What

- **Qubit roles**: Resolves moving vs. stationary qubits via `QubitRoles` / `verify_moving_qubit` (from the node 30 bootstrap utils). The moving qubit carries the flux pulse; the stationary qubit is the |11⟩↔|20⟩ leakage channel used for fitting.
- **Tunable coupler**: Detects `macros[operation].coupler_flux_pulse` and holds the coupler at its CZ bias during each flux sweep. The coupler is played separately at 4 ns granularity (not baked) to avoid `strict_timing_` conflicts with baked qubit pulses.
- **Analysis & plotting**: Fitting and detuning math now use the **moving** qubit's `freq_vs_flux_01_quad_term`; fit data comes from the **stationary** qubit. Plotting uses `QubitPairGrid` chip topology and produces separate figures for moving and stationary qubits.
- **Parameters & state update**: Adds `operation` (default `"cz_unipolar"`) and `use_saved_detuning`; removes `update_all_pulses`. State update writes fitted amplitude/length only to the configured macro (macro must already exist from populate).
- **Re-analysis**: Saves `pulse_amplitudes` and `qubit_roles` in results so loaded data can be re-analyzed without re-running the sweep.

### DIFF
--- a/qualibration_graphs/superconducting/calibrations/CZ_calibrations/31_chevron_11_20.py
+++ b/qualibration_graphs/superconducting/calibrations/CZ_calibrations/31_chevron_11_20.py
@@ -97,7 +97,12 @@ def create_qua_program(node: QualibrationNode[Parameters, Quam]):
     # qp.moving_qubit in-memory if they disagree; state is persisted at the end of the node.
     qubit_roles_map = {}
     for qp in qubit_pairs:
-        verify_moving_qubit(qp, log_callable=node.log)
+        verify_moving_qubit(
+            qubit_pair=qp,
+            operation=node.parameters.operation,
+            repair_routing=True,
+            log_callable=node.log,
+        )
         qubit_roles_map[qp.name] = QubitRoles.resolve(qp)
     node.namespace["qubit_roles_map"] = qubit_roles_map
 
--- a/qualibration_graphs/superconducting/calibrations/CZ_calibrations/31_chevron_11_20.py
+++ b/qualibration_graphs/superconducting/calibrations/CZ_calibrations/31_chevron_11_20.py
@@ -1,14 +1,16 @@
 # %% {Imports}
 from dataclasses import asdict
-
 import matplotlib.pyplot as plt
 import numpy as np
 import xarray as xr
 from calibration_utils.chevron_cz import (
     Parameters,
     baked_waveform,
+    estimate_cz_flux_amplitude,
     fit_raw_data,
     log_fitted_results,
+    QubitRoles,
+    verify_moving_qubit,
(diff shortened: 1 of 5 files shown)


paulQM (approved):
LGTM

qualibration_graphs/superconducting/calibrations/CZ_calibrations/31_chevron_11_20.py
```diff
@@ -78,22 +79,32 @@ def custom_param(node: QualibrationNode[Parameters, Quam]):
...
     # Get the qubit pairs to be calibrated
     node.namespace["qubit_pairs"] = qubit_pairs = get_qubit_pairs(node)
     num_qubit_pairs = len(qubit_pairs)
 
+    # Verify that qp.moving_qubit in state matches the recalculation from qubit frequencies and
+    # anharmonicity, then precompute roles for QUA program loops. Logs a warning and corrects
+    # qp.moving_qubit in-memory if they disagree; state is persisted at the end of the node.
+    qubit_roles_map = {}
+    for qp in qubit_pairs:
+        verify_moving_qubit(qp, log_callable=node.log)
```
HuanQM: I am thinking if it is better to pass `operation=node.parameters.operation` into `verify_moving_qubit`, like:
`verify_moving_qubit(qp, operation=node.parameters.operation, log_callable=node.log)`

This won't affect the sweep and experiment since this node does not use `macros[operation].apply()`. However, `verify_moving_qubit` currently defaults to `cz_unipolar`. If a user sets `parameters.operation` to a different macro and `cz_unipolar` is somehow deleted in `qp.macros`, this will raise an unnecessary error.

HuanQM (approved):
LGTM overall, only minor comment.

qualibration_graphs/superconducting/calibrations/CZ_calibrations/31_chevron_11_20.py
Deepakkhurrana: Thanks for noticing this. I have added the operation as well as routing option because for fixed freq coupler architecture, this will the first node that wound be run.

HuanQM (approved):

PR #530, opened 2026-06-17
Upgrade the Qubit Spectroscopy experiment to the latest version taken from the CS_installations repo

ohad-lev-qm:
**feat(qubit-spectroscopy)**: upgrade node 03a_qubit_spectroscopy with Lorentzian fit and "quality gates".

## Summary

Upgrade node `03a_qubit_spectroscopy` and `calibration_utils.qubit_spectroscopy` with a so-called "v2" analysis features taken from the `CS_installations` repo. The QUA sequence is unchanged; post-processing, plotting, and QUAM update behavior are improved.

- Replace the legacy `peaks_dips` path with an explicit Lorentzian + linear-background `curve_fit`, including smart peak initialization (Savgol + `find_peaks`), a refit window, and an `fwhm_floor_hz` guard against fake-narrow noise-spike fits.

- Add R² / FWHM / contrast "quality gates" and new node parameters (`r2_threshold`, `max_fwhm_mhz`, `min_contrast`); failed fits set `node.outcomes` to `"failed"` and are skipped in `update_state`.

- Refresh plot panels with a window-clipped fit overlay, a values box for QUAM-bound outputs, and clear failed-fit indicators; apply per-qubit saturation amplitude bound checks.

## Plotting demo

**Before:**
figures amplitude

**After:**
figures amplitude


qualibration_graphs/superconducting/calibrations/1Q_calibrations/03a_qubit_spectroscopy.py
```diff
@@ -1,8 +1,20 @@
+"""Qubit spectroscopy calibration (v2: improved Lorentzian-fit analysis).
```
Deepakkhurrana, line 1: No need to say versions since it continuous updates happen on the nodes all the time.

qualibration_graphs/superconducting/calibrations/1Q_calibrations/03a_qubit_spectroscopy.py
```diff
@@ -1,8 +1,20 @@
...
+
+Same QUA program as the previous 03_qubit_spectroscopy node; uses the
```
Deepakkhurrana, line 3: I think we should not add new description on top. Please expand on description which already is there in the node. We can be concise in the node and put more elaborate comments in analysis if required.

qualibration_graphs/superconducting/calibrations/1Q_calibrations/03a_qubit_spectroscopy.py
```diff
@@ -12,97 +24,106 @@
...
-In order to facilitate the qubit search, the qubit pulse duration and amplitude can be changed manually
-from the node parameters.
 
-The data is post-processed to determine the qubit resonance frequency and the width of the peak.
+import sys
+import os
 
-Note that it can happen that the qubit is excited by the image sideband or LO leakage instead of the desired sideband.
-This is why calibrating the qubit mixer is highly recommended when using external mixers or the Octave.
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), "../../../.."))
```
Deepakkhurrana, line 44: This is not required along with corresponding imports

qualibration_graphs/superconducting/calibrations/1Q_calibrations/03a_qubit_spectroscopy.py
```diff
@@ -12,97 +24,106 @@
...
+    - (optional) qubit.xy.operations["x180"/"x90"].amplitude
 """
 
 
-# Be sure to include [Parameters, Quam] so the node has proper type hinting
 node = QualibrationNode[Parameters, Quam](
-    name="03a_qubit_spectroscopy",  # Name should be unique
-    description=description,  # Describe what the node is doing, which is also reflected in the QUAlibrate GUI
-    parameters=Parameters(),  # Node parameters defined under quam_experiment/experiments/node_name
+    name="03a_qubit_spectroscopy",
```
Deepakkhurrana, line 81: Why were the comments removed? This applies in general to all the comments removed

qualibration_graphs/superconducting/calibration_utils/qubit_spectroscopy/analysis.py
```diff
@@ -1,154 +1,527 @@
+"""Analysis utilities for qubit spectroscopy calibration (v2).
```
Deepakkhurrana, line 1: I think we should remove v2 here as well.

qualibration_graphs/superconducting/calibration_utils/qubit_spectroscopy/node.py
```diff
@@ -1,11 +1,13 @@
+"""Node utilities for qubit spectroscopy calibration."""
```
Deepakkhurrana, line 1: This is not subject of this PR but Why do we need this file I wonder? IT is a very simple function which can live inside paramter.py

qualibration_graphs/superconducting/calibration_utils/qubit_spectroscopy/plotting.py
```diff
@@ -1,87 +1,217 @@
+"""Plotting utilities for qubit spectroscopy calibration (v2).
+
+Per-qubit panel keeps the dual-axis (absolute RF + detuning) layout from the
```
Deepakkhurrana, line 3: Lets not say legacy here, I think these nodes are always under update. And also we are overwriting the node so not sure how to access legacy.

qualibration_graphs/superconducting/calibration_utils/qubit_spectroscopy/plotting.py
```diff
@@ -1,87 +1,217 @@
...
+
+# Per-panel sizing. Width per column accommodates the dual axis labels +
+# values box; height per row leaves space for title and ticks.
+_PER_COL_INCH = 6.0
+_PER_ROW_INCH = 4.5
+_MIN_WIDTH_INCH = 8.0
+_MIN_HEIGHT_INCH = 6.0
+
+
+def _setup_grid_figure(grid) -> None:
```
Deepakkhurrana, line 46: Not sure what this means here? Why did we have to pass so many hardcoded params for plots?

qualibration_graphs/superconducting/calibration_utils/qubit_spectroscopy/plotting.py
```diff
@@ -1,87 +1,217 @@
...
-    - The function creates a grid of subplots, one for each qubit.
-    - Each subplot contains the raw data and the fitted curve.
+    nrows, ncols = grid.all_axes.shape
+    width = max(_MIN_WIDTH_INCH, ncols * _PER_COL_INCH)
+    height = max(_MIN_HEIGHT_INCH, nrows * _PER_ROW_INCH)
+    grid.fig.set_size_inches(width, height)
+    grid.fig.tight_layout(pad=1.5, w_pad=2.0, h_pad=2.0)
+
+
+def _apply_tick_fontsize(ax: Axes, size: int = _FS_TICK) -> None:
```
Deepakkhurrana, line 61: Is this really needed, can be inline. Also for other helpers, I think we should make it inline if it is used once.

qualibration_graphs/superconducting/calibration_utils/qubit_spectroscopy/__init__.py
```diff
@@ -1,11 +1,23 @@
+"""Qubit spectroscopy calibration utilities (v2)."""
+
 from .parameters import Parameters
-from .analysis import process_raw_dataset, fit_raw_data, log_fitted_results
+from .analysis import (
+    process_raw_dataset,
+    fit_raw_data,
+    log_fitted_results,
+    lorentzian_peak_linbg,
```
Deepakkhurrana, line 8: Do we need to import them out of the module? Does any other node need them?

Deepakkhurrana (changes requested):
Thanks for the update! :)
The analysis upgrade (Lorentzian fit, quality gates, skipping bad QUAM updates) is solid, but the PR adds a lot of noise by treating this like a “v2 migration” instead of a straight improvement to 03a. “v2 / legacy / replaces peaks_dips” is repeated across the node, GUI description, and every utils file even though there’s no parallel old node to keep.  I’d trim the node back to the usual thin orchestration pattern, drop the versioning language, keep implementation detail in utils only, and avoid exporting internal helpers (fit_qubit_peak, etc.) unless something else actually needs them.

PR #535, opened 2026-07-09
Two qubit Randomized Benchmarking for CZ calibrations

Deepakkhurrana:
## Summary

Adds two-qubit randomized benchmarking to the CZ calibration graph: **standard RB** (`37a`) for 2Q Clifford fidelity and **interleaved CZ RB** (`37b`) for isolated CZ gate fidelity. Shared logic lives in `calibration_utils/two_qubit_rb/`.

## What

- **`37a_two_qubit_standard_rb`** — random 2Q Clifford circuits, OPX execution, exponential decay fit, QUAM update (`StandardRB`, `StandardRB_alpha`, `StandardRB_load_id`).
- **`37b_two_qubit_interleaved_cz_rb`** — CZ-interleaved circuits; CZ fidelity from reference 37a decay; QUAM update (`InterleavedRB`, `InterleavedRB_alpha`).
- **`calibration_utils/two_qubit_rb/`** — circuit generation/transpilation, QUA program (`switch_case` per layer), optional input streaming, on-disk RB cache, analysis with validation, and chip-grid plotting.

## Why

Characterize 2Q gate quality after CZ calibration. Standard RB measures overall 2Q Clifford error; interleaved RB isolates CZ error from surrounding Clifford noise. Both follow the existing Qualibrate pattern and support multiplexed qubit pairs.

### Execution
- `use_input_stream` for large sweeps (avoids OPX ~16k QUA variable limit).
- Barriered batch transpile (`transpile_per_clifford_barriered`) + DAG-layer encoding (`circuit_to_layer_ints`).
- Unified analysis for with/without input stream.
- Multiplexed qubit-pair support; transpiled-circuit disk cache (`.rb_cache/`).

### Metrics
- EPC / EPG in logs and plot titles.
- 37b optional `fidelity_threshold` for adaptive-graph routing on low CZ fidelity (out of scope for this PR)

### Plotting
- `rb_plot_style`: `error_bars` (default) or `per_sequence`.
- `rb_plot_log_x`: log circuit-depth axis.
- Fit curves only when `success=True`; failed fits show raw data + “fit failed” (no misleading fidelity).
- 37b Standard RB overlay fit hidden when interleaved fit fails.

### Hard validation (`success=False` → no QUAM update)
- Fit convergence failure.
- α ∉ (0, 1]; invalid A/B (A ≥ 0, B ≥ 0, A + B ≤ 1).
- Fidelity ∉ [0, 1] (includes CZ > 100% on 37b).
- Interleaved α > reference `StandardRB_alpha`.
- Reference α out of range.
- Fitted curve > 4σ from data.

Failure reasons logged via `fit_issues`.

### Soft warnings (log only)
- Non-monotonic shallow vs max depth (SPAM/readout / low shots).
- P(|00⟩) at shallowest depth ≪ 0.5 (readout/reset/SPAM).

Logged via `fit_warnings`.

## Prerequisites
- Single-qubit gates, CZ, readout, flux biases calibrated.
- **37b** requires a recent **37a** run on the same `operation` (`StandardRB_alpha` in QUAM).

## Test plan
- [ ] Run `37a` on one pair — small sweep (few depths / circuits / shots)
- [ ] Re-run with `use_input_stream=True` and a larger depth list
- [ ] Confirm QUAM updates on successful fit; no update on hard-fail
- [ ] Deliberately bad data or stale 37a reference → hard-fail logs; plot shows no fit line
- [ ] Check soft warnings appear in logs for noisy / low shallow-survival data
- [ ] Run `37b` after 37a on the same pair; verify CZ fidelity / EPG / EPC
- [ ] Try `rb_plot_style=per_sequence` and `rb_plot_log_x=True`
- [ ] Optional: `simulate=True` smoke test

## On QPU
figures rb
image


soonqm:
Looks good at a glance, but a few things I would add/fix
- Should calculate SEM and show it
- Should add corresponding coherence limit since we should already have T1, T2
- why is the [plot](https://private-user-images.githubusercontent.com/119570568/619522218-f22b1ec9-6d87-48ba-8c9c-de4cbaeda92f.png?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3ODM2NTc5MzUsIm5iZiI6MTc4MzY1NzYzNSwicGF0aCI6Ii8xMTk1NzA1NjgvNjE5NTIyMjE4LWYyMmIxZWM5LTZkODctNDhiYS04YzljLWRlNGNiYWVkYTkyZi5wbmc_WC1BbXotQWxnb3JpdGhtPUFXUzQtSE1BQy1TSEEyNTYmWC1BbXotQ3JlZGVudGlhbD1BS0lBVkNPRFlMU0E1M1BRSzRaQSUyRjIwMjYwNzEwJTJGdXMtZWFzdC0xJTJGczMlMkZhd3M0X3JlcXVlc3QmWC1BbXotRGF0ZT0yMDI2MDcxMFQwNDI3MTVaJlgtQW16LUV4cGlyZXM9MzAwJlgtQW16LVNpZ25hdHVyZT00ZTNjY2Q5N2FkMDg4MTRiZjM2ZGI4OWI1YzE3MDI2MWU4ZWI4NDgzNjkyNjdlY2RhZWU5MGY3MWZmYjc5YzliJlgtQW16LVNpZ25lZEhlYWRlcnM9aG9zdCZyZXNwb25zZS1jb250ZW50LXR5cGU9aW1hZ2UlMkZwbmcifQ.XKUQcpX-6eZ0Sre65CFoK2GL1Fu9KlrMYFfBSE-VEnI) using different color and label for interleaved?

I will have a closer look soon

Deepakkhurrana:
> Looks good at a glance, but a few things I would add/fix
> 
> * Should calculate SEM and show it
> * Should add corresponding coherence limit since we should already have T1, T2
> * why is the [plot](https://private-user-images.githubusercontent.com/119570568/619522218-f22b1ec9-6d87-48ba-8c9c-de4cbaeda92f.png?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3ODM2NTc5MzUsIm5iZiI6MTc4MzY1NzYzNSwicGF0aCI6Ii8xMTk1NzA1NjgvNjE5NTIyMjE4LWYyMmIxZWM5LTZkODctNDhiYS04YzljLWRlNGNiYWVkYTkyZi5wbmc_WC1BbXotQWxnb3JpdGhtPUFXUzQtSE1BQy1TSEEyNTYmWC1BbXotQ3JlZGVudGlhbD1BS0lBVkNPRFlMU0E1M1BRSzRaQSUyRjIwMjYwNzEwJTJGdXMtZWFzdC0xJTJGczMlMkZhd3M0X3JlcXVlc3QmWC1BbXotRGF0ZT0yMDI2MDcxMFQwNDI3MTVaJlgtQW16LUV4cGlyZXM9MzAwJlgtQW16LVNpZ25hdHVyZT00ZTNjY2Q5N2FkMDg4MTRiZjM2ZGI4OWI1YzE3MDI2MWU4ZWI4NDgzNjkyNjdlY2RhZWU5MGY3MWZmYjc5YzliJlgtQW16LVNpZ25lZEhlYWRlcnM9aG9zdCZyZXNwb25zZS1jb250ZW50LXR5cGU9aW1hZ2UlMkZwbmcifQ.XKUQcpX-6eZ0Sre65CFoK2GL1Fu9KlrMYFfBSE-VEnI) using different color and label for interleaved?
> 
> I will have a closer look soon
@soonqm 
- Should calculate SEM and show it
Addressed. Now there is a detailed logged info on EPC, EPG, methods used, EPC 1q and CZ contribution and coherence time flo
Screenshot 2026-07-10 at 15 43 40
or.

- Coherence limit
Addressed. Coherence-limited EPG is computed from QUAM T1 and T2echo on both qubits and the CZ pulse duration (Qiskit 2Q formula). On 37b it appears in the log (with a not-/coherence-limited verdict vs measured CZ EPG) and in the plot title as EPG (coherence limit) = …. On 37a the floor is logged on success only (no direct CZ EPG to compare).

- Different color/label for interleaved
Fixed. On 37b, standard RB reference (from 37a) is C0 / ○ (StandardRB Mean + dashed fit); interleaved data is C1 / △ (Interleaved Mean + fit). This keeps reference and interleaved curves distinguishable on one panel. 37a uses C0 / ○ only.
image

image

qualibration_graphs/superconducting/calibration_utils/two_qubit_rb/qua_utils.py
```diff
@@ -0,0 +1,788 @@
...
+
+            for qp in qubit_pair.values():
+
+                # reset the qubits
+                qp.qubit_control.reset(reset_type, simulate)
+                qp.qubit_target.reset(reset_type, simulate)
+
+            align()
+
+            # # Reset the frame of the qubits in order not to accumulate rotations
```
omrieoqm: are these comments needed?

qualibration_graphs/superconducting/calibrations/CZ_calibrations/37b_two_qubit_interleaved_cz_rb.py
```diff
@@ -0,0 +1,277 @@
+# pylint: disable=duplicate-code
+
```
omrieoqm, line 2: how is the EPG converted from the EPC ? is there an average_gates_per_2q_layer calculation somewhere ?

qualibration_graphs/superconducting/calibrations/CZ_calibrations/37b_two_qubit_interleaved_cz_rb.py
```diff
@@ -0,0 +1,277 @@
...
+# %% {Imports}
+from dataclasses import asdict
```
omrieoqm, line 4: Since this node is QOP compiler heavy I would recommend testing this node for multiple pairs mutliplexed/not_multiplexed and with/without active reset.

qualibration_graphs/superconducting/calibrations/CZ_calibrations/37b_two_qubit_interleaved_cz_rb.py
Deepakkhurrana: - EPG is calculated differently for standard RB and interleaved RB.
- In standard RB, EPG gives an ad hoc idea of how good a gate is on average (1Q + CZ blended). We go one step further and calculate an **implied CZ EPG** by factoring out stored 1Q fidelities from QUAM (with a warning that this is only a rough estimate — use 37b for a direct CZ EPG).
- In interleaved RB, EPG is a **direct measurement** of how good the CZ gate is.

Lets look at it in detail:

---

## Standard RB (37a) — blended average EPG

**What you measure:** Survival $P(|00\rangle)$ vs 2Q Clifford depth $m$.

**Step 1 — Fit the decay**

$$P(m) = A\,\alpha^m + B$$

**Step 2 — 2Q Clifford fidelity** from $\alpha$ (2 qubits, $d=4$):

$$F_{\mathrm{Clifford}} = 1 - \left(1 - \alpha - \frac{1-\alpha}{d}\right)$$

**Step 3 — Error per Clifford (EPC)**

$$\mathrm{EPC} = 1 - F_{\mathrm{Clifford}}$$

**Step 4 — Blended error per gate (EPG)** — divide EPC by the measured average native gate count per Clifford from transpilation:

$$\mathrm{EPG}_{\mathrm{avg}} = \frac{\mathrm{EPC}}{N_{\mathrm{gates/Clifford}}} = \frac{1 - F_{\mathrm{Clifford}}}{N_{\mathrm{gates/Clifford}}}$$

$N_{\mathrm{gates/Clifford}}$ comes from `summarize_transpiled_depth`: count all native gates (`rz`, `sx`, `x`, `cz`) in the transpiled RB circuits, pool over depths, divide by total Cliffords. Stored in namespace as `average_gates_per_clifford`.

**Interpretation:** $\mathrm{EPG}_{\mathrm{avg}}$ is an average over **all** gates in a 2Q Clifford (1Q + CZ), **not** CZ-only.

---

### Optional (37a only) — implied CZ EPG

Requires stored 1Q RB fidelities on both qubits (`gate_fidelity["averaged"]` from node 11a) and $N_{\mathrm{CZ}}$, $N_{\mathrm{1Q}}$ from transpile stats.

**Step 5a — 1Q error budget per Clifford**

Assume $N_{\mathrm{1Q}}/2$ single-qubit gates on control and $N_{\mathrm{1Q}}/2$ on target:

$$F_{\mathrm{1Q,budget}} = F_{\mathrm{1Q},c}^{\,N_{\mathrm{1Q}}/2} \cdot F_{\mathrm{1Q},t}^{\,N_{\mathrm{1Q}}/2}$$

$$\mathrm{EPC}_{\mathrm{1Q}} = 1 - F_{\mathrm{1Q,budget}}$$

**Step 5b — CZ residual at Clifford level**

$$\mathrm{EPC}_{\mathrm{CZ,residual}} = \mathrm{EPC} - \mathrm{EPC}_{\mathrm{1Q}}$$

**Step 5c — Implied CZ gate fidelity and EPG**

Assume multiplicative model: $F_{\mathrm{Clifford}} \approx F_{\mathrm{1Q,budget}} \cdot F_{\mathrm{CZ}}^{\,N_{\mathrm{CZ}}}$

$$F_{\mathrm{CZ,implied}} = \left(\frac{F_{\mathrm{Clifford}}}{F_{\mathrm{1Q,budget}}}\right)^{1/N_{\mathrm{CZ}}}$$

$$\mathrm{EPG}_{\mathrm{CZ,implied}} = 1 - F_{\mathrm{CZ,implied}} = 1 - \left(\frac{F_{\mathrm{Clifford}}}{F_{\mathrm{1Q,budget}}}\right)^{1/N_{\mathrm{CZ}}}$$

**Caveat:** This is **not** interleaved RB. It assumes 1Q and CZ errors combine multiplicatively and that stored 1Q fidelities are accurate. **Trust 37b for direct CZ EPG.**

---

## Interleaved RB (37b) — direct CZ EPG

**What you measure:** Same decay fit on **interleaved** sequences (CZ inserted between each Clifford).

**Step 1 — Fit** interleaved decay → $\alpha_{\mathrm{IRB}}$

**Step 2 — Reference** from 37a → $\alpha_{\mathrm{SRB}}$

**Step 3 — CZ gate fidelity** (Magesan et al., [arXiv:1210.7011](https://arxiv.org/pdf/1210.7011)):

$$F_{CZ} = 1 - \frac{d-1}{d}\left(1 - \frac{\alpha_{\mathrm{IRB}}}{\alpha_{\mathrm{SRB}}}\right), \quad d=4$$

**Step 4 — EPG**

$$\mathrm{EPG}_{CZ} = 1 - F_{CZ}$$

No `EPC / N_gates` and no `average_gates_per_2q_layer`. EPC on 37b is only the reference Clifford error (for context); it is **not** used to compute EPG.

**Interpretation:** $\mathrm{EPG}_{CZ}$ on 37b is the **CZ gate error** isolated by interleaved RB.

---

## Summary

| | **37a Standard** | **37b Interleaved** |
|---|---|---|
| **Primary fit** | $\alpha$ from 2Q Clifford RB | $\alpha_{\mathrm{IRB}}$ from interleaved RB |
| **Main EPG** | $\mathrm{EPC} / N_{\mathrm{gates/Clifford}}$ (blended) | $1 - F_{CZ}$ (CZ only) |
| **CZ EPG** | Implied only (Step 5, diagnostic) | Direct measurement |
| **Gate counting** | Yes — transpile stats | No |
| **Needs 37a first?** | No | Yes — $\alpha_{\mathrm{SRB}}$ |

qualibration_graphs/superconducting/calibrations/CZ_calibrations/37b_two_qubit_interleaved_cz_rb.py
Deepakkhurrana: Yes agreed! On it!

qualibration_graphs/superconducting/calibration_utils/two_qubit_rb/qua_utils.py
Deepakkhurrana: Thanks for notocing this. Unnecessary Comments are removed.

qualibration_graphs/superconducting/calibrations/CZ_calibrations/37b_two_qubit_interleaved_cz_rb.py
Deepakkhurrana: There is not much disagreement between thermal and active reset 

image
image

Deepakkhurrana: Also ran for three multiplexed qubit pairs (not adjacent ), the experiment runs fine. However the results are different (see pic underneath). Not 100% sure if it is expected or mutiplexing logic is interfering.  Thoughts? 

**Not Multiplexed** 

Screenshot 2026-07-13 at 12 37 16

**Multiplexed** 
Screenshot 2026-07-13 at 12 37 33

soonqm: 1. It seems that even at depth=0, the probability is higher with multiplexing. So, It might have shifted the IQ blob higher. Is this the same seed?
2. What is the results of qB4-qB2 pair without other pairs? Does it have any effect?

Deepakkhurrana: 1. Seed is same for both cases. I think it could be just qubit xtalk
2. In this case, there is not affect of multiplexing. Same result multiplexed or not

qualibration_graphs/superconducting/calibrations/CZ_calibrations/37b_two_qubit_interleaved_cz_rb.py
soonqm: 1. SEM is not really showing? I might have missed it but why not put it on the  plot since that is probably one of the most important steps to have uncertainty labelled
2. The implied 2Q fidelity formula is inaccurate and does not take into account the non-identity 2Q Pauli operators. See [Eq 10 here](https://arxiv.org/src/1712.06550v2/anc/threeq_supp.pdf)
image

soonqm:
Additional questions: do we meant for this 2QRB to be general for all other gates or just CZ? The calibration_utils/two_qubit_rb exclusively refers to CZ in most cases, maybe should be renamed?

qualibration_graphs/superconducting/calibration_utils/two_qubit_rb/parameters.py
```diff
@@ -0,0 +1,130 @@
...
+class NodeSpecificParameters(RunnableParameters):
+    """Node-specific parameters for two-qubit RB experiments."""
+
+    num_shots: int = 100
+    """Number of averages to perform. Default is 50."""
+    operation: Literal["cz_flattop", "cz_unipolar", "cz_bipolar", "cz_flattop_erf", "cz_SNZ"] = "cz_unipolar"
+    """Type of CZ operation to perform."""
+    use_state_discrimination: bool = True
+    """Whether to use state discrimination for readout. Default is True."""
+    circuit_depths: list[int] = [1, 2, 4, 8, 16, 32, 64]
```
soonqm: Should consider geomspace, linspace. Is there any reason/benefit to choosing specific value?

qualibration_graphs/superconducting/calibrations/CZ_calibrations/37b_two_qubit_interleaved_cz_rb.py
soonqm: 2. Is it more similar to the multiplex results or without?

qualibration_graphs/superconducting/calibrations/CZ_calibrations/37b_two_qubit_interleaved_cz_rb.py
Deepakkhurrana: 1. Ah! You mean SEM for the final fidelity itself? Sorry got it wrong last time. On it! 
2. Great catch, fixing it

qualibration_graphs/superconducting/calibrations/CZ_calibrations/37b_two_qubit_interleaved_cz_rb.py
Deepakkhurrana: > 2. Is it more similar to the multiplex results or without?

Without Multiplex

qualibration_graphs/superconducting/calibrations/CZ_calibrations/37b_two_qubit_interleaved_cz_rb.py
Deepakkhurrana: > * Ah! You mean SEM for the final fidelity itself? Sorry got it wrong last time. On it!

Is this better?
image

soonqm: Perfect!

qualibration_graphs/superconducting/calibration_utils/two_qubit_rb/qua_utils.py
```diff
@@ -0,0 +1,779 @@
...
+            for qp in qubit_pair.values():
+                qp.qubit_control.wait(4)
+                qp.qubit_target.wait(4)
+                qp.align()
+        with case_(36):  # CZ
+            for qp in qubit_pair.values():
+                qp.macros[cz_operation].apply()
+                qp.align()
+        with case_(37):  # idle_2q
+            for qp in qubit_pair.values():
```
soonqm, line 467: Case 35 and 37 is the same. Is this intentional?

qualibration_graphs/superconducting/calibration_utils/two_qubit_rb/qua_utils.py
```diff
@@ -0,0 +1,779 @@
...
+                # Wait for 4 cycles
+                qp.qubit_control.wait(4)
+                qp.qubit_target.wait(4)
+                qp.align()
+
+        with case_(38):  # readout and thermalization
+            align()
+            # Readout the qubits and save the state
+            for i, qp in qubit_pair.items():
+                wait(8)
```
soonqm: Do we need to wait?

qualibration_graphs/superconducting/calibration_utils/two_qubit_rb/qua_utils.py
Deepakkhurrana: Yes, intentional.

Opcodes 35 and 37 run the same QUA (wait(4) on both qubits), but they encode different circuit roles:

35: last slot in the 6×6 single-qubit parallel grid (idle on both qubits). Completes the encoding
37: explicit two-qubit `idle_2q` gate for interleaved RB and input-stream padding.

Same hardware action, different circuit semantics. We keep both so the opcode table is complete and idle_2q interleaving is unambiguous.

Deepakkhurrana:
I observed that analysis module was very messy. I have refactored it in this PR because it is mainly file restructure.  

## Analysis refactor

Split monolithic `analysis.py` into four modules:

| Module | Role |
|--------|------|
| `analysis.py` | Orchestration: `process_raw_dataset`, `fit_raw_data`, `RBMode` |
| `fit_utils.py` | Curve fitting, validation, SRB/IRB per-pair fit, 37a overlay |
| `fidelity.py` | α → fidelity, EPC/EPG, implied-CZ, stderr propagation |
| `reporting.py` | Result dataclasses + `log_srb_results` / `log_irb_results` |

**Node API:**
- 37a → `fit_raw_data(..., mode=RBMode.STANDARD)` + `log_srb_results`
- 37b → `fit_raw_data(..., mode=RBMode.INTERLEAVED)` + `log_irb_results`

**Also fixed:** ± on EPG logs, 37a `load_data_id` gate-stats restore, IRB overlay sequence-dimension mismatch, trimmed `__init__` exports.

qualibration_graphs/superconducting/calibrations/CZ_calibrations/37b_two_qubit_interleaved_cz_rb.py
Deepakkhurrana: the impled EPG calculation now is fixed and follows the above formula. Thanks @soonqm

Line 220 of the new module fidelity.py

qualibration_graphs/superconducting/calibration_utils/two_qubit_rb/qua_utils.py
Deepakkhurrana: Not really, it is removed now.

Deepakkhurrana:
> Additional questions: do we meant for this 2QRB to be general for all other gates or just CZ? The calibration_utils/two_qubit_rb exclusively refers to CZ in most cases, maybe should be renamed?

I think right now it is specific to CZ and lets keep it this way. When we bring in CR modules, we can go ahead with merging it into one.

qualibration_graphs/superconducting/calibration_utils/two_qubit_rb/parameters.py
Deepakkhurrana: It is a deliberate choice because sometimes decays are too fast and one wants more points at shallow depths and couple of points at higher depths to capture the tail. In this case linspace/geomspace will be limiting.

soonqm: np.geomspace(1, 64, 7).astype(int) = array([ 1,  2,  4,  7, 16, 32, 64]). Is this good enough without having to re-write the number?

Deepakkhurrana: Yes for sure. I will make this change.

Deepakkhurrana: Done!

soonqm:
```
    depths = np.geomspace(1, max_circuit_depth + 1, intervals, dtype=int) - 1
    depths = np.unique(depths)
```

Recommend doing the above to include 0, and also to avoid same depth when user chooses something like np.geomspace(1, 10, 100)

Deepakkhurrana: So `max_circuit_depth`  forces the users to use linsoace or logspace but sometimes they might wanna just use a custom list, in that case I think explicitly have a depth list is better idea

soonqm: I think that is unlikely? Considering we need to get a good fit of power law at the end

Deepakkhurrana: I agree!  But sometimes QPU are bad. In any case, this is all speculation. Lets go for `max_circuit_depth`. seems a more elegant  choice

qualibration_graphs/superconducting/calibration_utils/two_qubit_rb/fit_utils.py
```diff
@@ -0,0 +1,506 @@
...
+    """P(|00>) per random sequence at each depth, averaged over shots."""
+    survival = (ds_qp.state == 0).mean(dim="shots")
+    if "qubit_pair" in survival.dims:
+        survival = survival.squeeze("qubit_pair", drop=True)
+    return survival
+
+
+def _survival_stderr(ds_qp: xr.Dataset) -> xr.DataArray:
+    """Standard error of the mean survival probability at each circuit depth."""
+    n_samples = ds_qp.sizes["sequence"] * ds_qp.sizes["shots"]
```
soonqm: I believe that it should be number of sequence only when obtaining the standard error, see [qiskit](https://github.com/qiskit-community/qiskit-experiments/blob/46329108c6e75f76d142ac951c7d954ec22da617/qiskit_experiments/library/randomized_benchmarking/rb_analysis.py#L124)

qualibration_graphs/superconducting/calibrations/CZ_calibrations/37b_two_qubit_interleaved_cz_rb.py
soonqm: Got it. It still raise some questions on why does multiplexing improves the EPG in the presence of crosstalk.

qualibration_graphs/superconducting/calibration_utils/two_qubit_rb/fit_utils.py
Deepakkhurrana: YEs good catch! corrected now!

qualibration_graphs/superconducting/calibration_utils/two_qubit_rb/parameters.py
Deepakkhurrana: Change has been implemented. 

image

qualibration_graphs/superconducting/calibrations/CZ_calibrations/37a_two_qubit_standard_rb.py
```diff
@@ -0,0 +1,306 @@
...
+        for stat_key in ("average_gates_per_clifford", "avg_1q_per_clifford", "avg_cz_per_clifford"):
+            if stat_key not in node.namespace and stat_key in cached:
+                node.namespace[stat_key] = float(cached[stat_key])
+
+
+# %% {Analyse_data}
+@node.run_action(skip_if=node.parameters.simulate)
+def analyse_data(node: QualibrationNode[Parameters, Quam]):
+    """Analyse raw data, fit, log results, set outcomes and store structured fit results."""
+    node.results["ds_raw"] = process_raw_dataset(node.results["ds_raw"], node)
```
soonqm: I understand that overwriting ds_raw is the convention all this while but isn't it better to do
```
node.results["ds_proc"] = process_raw_dataset(node.results["ds_raw"], node)
node.results["ds_fit"], fit_results = fit_raw_data(node.results["ds_proc"], node)
```
such that we keep the raw data from OPX execution?

qualibration_graphs/superconducting/calibrations/CZ_calibrations/37b_two_qubit_interleaved_cz_rb.py
Deepakkhurrana: Yup! Valid point. I think this will clarified by eventual testing if these nodes on several QPUs and out of scope for this PR.

qualibration_graphs/superconducting/calibrations/CZ_calibrations/37b_two_qubit_interleaved_cz_rb.py
```diff
@@ -0,0 +1,267 @@
...
+        node.parameters.circuit_depths,
+        node.parameters.num_circuits_per_depth,
+        target_gate="cz",
+    )
+    cache_dir = Path(__file__).resolve().parents[2] / ".rb_cache"
+    cached = try_load(cache_dir, key)
+
+    if cached is not None:
+        circuits_as_ints = cached["circuits_as_ints"]
+        node.log(f"Loaded {len(circuits_as_ints)} cached interleaved RB circuits (key {key[:12]})")
```
soonqm: interesting magic number 12. Shouldn't log automatically cut it off? Also, do we just keep caching with no ways of cleaning up?

qualibration_graphs/superconducting/calibrations/CZ_calibrations/37b_two_qubit_interleaved_cz_rb.py
```diff
@@ -0,0 +1,267 @@
...
+    node.results["fit_results"] = {k: asdict(v) for k, v in fit_results.items()}
+    log_irb_results(fit_results, log_callable=node.log)
+
+    threshold = node.parameters.fidelity_threshold
+    outcomes: dict[str, str] = {}
+    for qp_name, fit_result in fit_results.items():
+        if not fit_result.success:
+            outcomes[qp_name] = "failed"
+            continue
+        if threshold is not None and fit_result.fidelity < threshold:
```
soonqm, line 239: Also should fail if the standard error is larger than fidelity

qualibration_graphs/superconducting/calibration_utils/two_qubit_rb/analysis.py
```diff
@@ -0,0 +1,210 @@
...
+        )
+        ds_fit, fit_results = _extract_irb_results(ds_fit, node)
+
+    ds_fit = _attach_shared_coherence_limits(ds_fit, node)
+    return ds_fit, fit_results
+
+
+def _annotate_shared_attrs(ds_fit: xr.Dataset) -> None:
+    """Attach display attrs shared by both modes (mutates in place)."""
+    attrs_by_var = {
```
soonqm, line 63: Is it really a.u.?

qualibration_graphs/superconducting/calibrations/CZ_calibrations/37b_two_qubit_interleaved_cz_rb.py
omrieoqm: In the IQCC code there is a hard coded average number of gates per layer (calculated from the switch_ block) and also a calculation of average number of layers per Clifford (calculated from the circuits). Using these two we can calculate the average number of gates per Clifford which then give the EPG from the EPC. Is this procedure used here as well?

qualibration_graphs/superconducting/calibrations/CZ_calibrations/37b_two_qubit_interleaved_cz_rb.py
omrieoqm: Interesting. Are you able run non multiplexed? It does not crash ? I would expect some difference but really haven't investigated this at length. We mostly run multiplexed (non-multiplexed is crashing last time I checked but maybe in latest QOP this is not the case)

omrieoqm: The fact that it is running on active reset is very encouraging as well. It used to crash but latest QOP made it much more stable. Regarding the performance difference: we are seeing this in other experiments as well and are still investigating it.

omrieoqm (approved):

qualibration_graphs/superconducting/calibrations/CZ_calibrations/37b_two_qubit_interleaved_cz_rb.py
soonqm: image

Also, should we still plot the fit if fail because of 

> 2026-07-15 09:56:44,829 - qualibrate - INFO - Node 37b_two_qubit_int... - Results for qubit pair qA2-qA4: FAIL!
        - CZ gate fidelity=100.6642% outside physical range [0, 100%].
        - Interleaved alpha=0.930009 exceeds reference StandardRB_alpha=0.921845 (implies CZ fidelity > 100%).
INFO:qualibrate:Node 37b_two_qubit_int... - Results for qubit pair qA2-qA4: FAIL!
        - CZ gate fidelity=100.6642% outside physical range [0, 100%].
        - Interleaved alpha=0.930009 exceeds reference StandardRB_alpha=0.921845 (implies CZ fidelity > 100%).

soonqm: Same note. When I run seed=1 for interleaved and 0 for standard RB, it seems that it will run and compare as it is. Should we force check for interleaved if the previous run is using the same seed?

There might be some bug where if I choose different number of intervals lead to something like the following for num=5 for interleaved and 7 for standard RB. The workaround is to check as mentioned above
image

qualibration_graphs/superconducting/calibrations/CZ_calibrations/37a_two_qubit_standard_rb.py
Deepakkhurrana: Good idea! I can make the changes in these nodes but I think this is a larger refactor across all nodes in general.

qualibration_graphs/superconducting/calibrations/CZ_calibrations/37b_two_qubit_interleaved_cz_rb.py
Deepakkhurrana: I removed the truncated cache key from the logs; messages now only indicate cache hit vs miss and the circuit count. The full SHA-256 hash remains the on-disk filename only now. 

There is no automatic cache eviction as of yet.  Entries under .rb_cache/ are local, deterministic, and safe to delete manually (a miss triggers regeneration). I am not sure how to deal with this because cache helps save quite a bit of time at higher depths. For now, I added .rb_cache/ to .gitignore so cache files are not tracked in git at least.

qualibration_graphs/superconducting/calibrations/CZ_calibrations/37b_two_qubit_interleaved_cz_rb.py
Deepakkhurrana: Agreed. We added this to the fit validation path (validate_fidelity_uncertainty in fidelity.py, called from _validate_rb_fit), not as a separate check in the 37b outcomes block.

If the propagated 1σ uncertainty on the fitted fidelity exceeds the point estimate (fidelity_stderr > fidelity), the fit is marked success=False with an explicit fit_issues message. This applies to both Standard and Interleaved RB, so logs, plots, and node.outcomes stay consistent.

The optional fidelity_threshold in 37b remains a separate acceptance policy (“is the gate good enough?”), distinct from this statistical significance check.

qualibration_graphs/superconducting/calibrations/CZ_calibrations/37b_two_qubit_interleaved_cz_rb.py
Deepakkhurrana: It is not crashing for non-multiplexed using the the code in this PR :) I would be very curious to know how this holds up at IQCC internal testing.

qualibration_graphs/superconducting/calibrations/CZ_calibrations/37b_two_qubit_interleaved_cz_rb.py
Deepakkhurrana: **Average gates per Clifford (EPG denominator)**

We compute this at circuit generation from the actual transpiled RB ensemble — no hardcoded constant. Line 145 in `def _count_transpiled_circuit_stats(qc: QuantumCircuit)` in `circuit_utils.py`

For each transpiled circuit, barriers mark Cliffords. We count:
- **All transpiled gates** (`sx`, `x`, `rz`, `cz`)
- **Physical gates only** (`sx`, `x`, `cz`) — used for EPG

Then:

`N_gates/Clifford = (sum of physical gates) / (sum of Cliffords)`
`EPG = EPC / N_gates/Clifford
`This updates automatically with seed, depths, and circuit count. Unlike IQCC’s fixed 0.6757, we derive it from the circuits (I checked for a  randomdepth: ~0.673 vs 0.6757, ~0.4% agreement).

Fix: The earlier version incorrectly included virtual rz in the denominator (implemented as frame_rotation in QUA, not a physical pulse), which understated EPG. We now exclude rz from physical gate counts and log both totals for transparency.

qualibration_graphs/superconducting/calibration_utils/two_qubit_rb/fidelity.py
```diff
@@ -0,0 +1,270 @@
...
+    alpha_stderr: float | None,
+    standard_rb_alpha: float,
+    *,
+    epc_reference: float | None = None,
+) -> IRBFidelity:
+    """Direct CZ gate fidelity from interleaved alpha vs. a reference Standard RB alpha."""
+    fidelity = interleaved_gate_fidelity_from_alpha(alpha, standard_rb_alpha)
+    epg = 1 - fidelity
+    epc = epc_reference if epc_reference is not None else 1 - clifford_fidelity_from_alpha(standard_rb_alpha)
+    fidelity_stderr = fidelity_stderr_from_alpha(alpha_stderr, interleaved=True, standard_rb_alpha=standard_rb_alpha)
```
soonqm: I didn't check manually but I don't think the error propagation is done correctly in the interleaved case since the stderr should also depends on standard_rb_alpha_stderr. Why not just use `uncertainties` for everything involving stderr?

qualibration_graphs/superconducting/calibration_utils/two_qubit_rb/parameters.py
```diff
@@ -0,0 +1,135 @@
...
+    seed: int = 0
+    """Random seed for circuit generation to ensure reproducibility. Default is 0."""
+    use_input_stream: bool = False
+    """Whether to use input streams for circuit execution. Default is False.
+    When True, the gate sequences are streamed to the OPX chunk-by-chunk via the
+    QUA input-stream feature instead of being declared as a single large
+    `declare(int, value=...)` array. This bypasses the OPX's ~16000 QUA variable
+    budget cap on declared arrays, enabling longer circuit depths and/or more
+    circuits per depth than the without-input-stream path can support."""
+    max_chunk_ints: int = 16000
```
soonqm: Seems to work most of the cases, but do you think making default a bit smaller like 15000 is safer? Have you tested if we could do for example >20 qubit pairs for example.

qualibration_graphs/superconducting/calibration_utils/two_qubit_rb/parameters.py
```diff
@@ -0,0 +1,135 @@
...
+            "circuit_depth": xr.DataArray(np.array(circuit_depths)),
+            "shots": xr.DataArray(np.arange(num_shots)),
+            "sequence": xr.DataArray(np.arange(num_circuits_per_depth)),
+        }
+    return {
+        "qubit_pair": xr.DataArray(qubit_pairs.get_names()),
+        "shots": xr.DataArray(np.arange(num_shots)),
+        "circuit_depth": xr.DataArray(np.array(circuit_depths)),
+        "sequence": xr.DataArray(np.arange(num_circuits_per_depth)),
+    }
```
soonqm, line 135: Are these not the same?

qualibration_graphs/superconducting/calibrations/CZ_calibrations/37b_two_qubit_interleaved_cz_rb.py
soonqm, line 1: I think the logic between 37a and 37b is very similar. Would it be better to just import the function directly? instead of copy pasting

qualibration_graphs/superconducting/calibration_utils/two_qubit_rb/parameters.py
Deepakkhurrana: So we flip the array for input stream because we run shots loop inside the circuit depth chunks to make the experiment faster. The without input stream deliberately follows the logic of all our other experiments where shot loop comes first. So without input stream = standard + simple; with input stream = bypass the size cap and run faster by pushing each chunk once and replaying shots on the OPX.

qualibration_graphs/superconducting/calibration_utils/two_qubit_rb/parameters.py
Deepakkhurrana: No I have not. I agree 15k would be safer. Changing it

qualibration_graphs/superconducting/calibrations/CZ_calibrations/37b_two_qubit_interleaved_cz_rb.py
Deepakkhurrana: Sorry I did not get it, which function is this?

qualibration_graphs/superconducting/calibration_utils/two_qubit_rb/parameters.py
soonqm: Ah, right. This is for the loop. Got it!

qualibration_graphs/superconducting/calibrations/CZ_calibrations/37b_two_qubit_interleaved_cz_rb.py
soonqm: I just diff 37a and 37b node directly. Anyway, it is fine to leave it as it is for now.

qualibration_graphs/superconducting/calibration_utils/two_qubit_rb/fidelity.py
Deepakkhurrana: Yes great catch! Fixing it. Let me see ho big a refactor uncertainties be and get back

qualibration_graphs/superconducting/calibration_utils/two_qubit_rb/analysis.py
Deepakkhurrana: Previous calibration utils used a.u. informally for any dimensionless quantity (including probabilities). That’s legacy convention, not correct metrology. For RB, P(|00>), fidelity, EPC/EPG, and α are dimensionless probabilities/rates → units: "1" is more accurate. We can fix RB attrs without changing the rest of the graph in this PR. Thoughts?

qualibration_graphs/superconducting/calibration_utils/two_qubit_rb/fidelity.py
Deepakkhurrana: The refactor has been implemented

qualibration_graphs/superconducting/calibration_utils/two_qubit_rb/analysis.py
soonqm: I would just leave it, i.e. not include units attr unless it is required anywhere else

Deepakkhurrana: Not required anywhere else, removed them now

soonqm (approved):

PR #537, opened 2026-07-15
quantum_dots: shared quam_config + test-harness infra

KU-QM:
Part 1 of migrating `quantum_dots` calibration updates into `release/nightly`, split into small reviewable PRs. This one is shared infra that later PRs build on:

- `quam_config`: `Quam` now subclasses `LossDiVincenzoQuam`, new `pulse_wirer.py` / `generate_quam.py` state-generation helpers, wiring examples reformatted.
- `validation_utils`: formatting-only changes, no behavior change.
- Adds `cma` dependency; root `pyproject.toml` moves from Poetry to a `uv` workspace.
- New shared test harness (`tests/quam_test_machine.py`, `save_quam_state.py`) and `quam_factory.py`/`shared_fixtures.py` updates used by later PRs' tests.
- `.gitignore`: ignore `**/.rb_cache/` build-cache output.

Excludes a few files found upstream that looked like local scratch/junk (stray notebooks, debug DB, a duplicate script) rather than intended production content — happy to revisit if any of those were wanted.

Made with [Cursor](https://cursor.com)


qualibration_graphs/quantum_dots/quam_config/wiring_examples/wiring_mwfem_cross_resonance.py
```diff
@@ -0,0 +1,85 @@
+import matplotlib.pyplot as plt
```
TheoLaudatQM: This one is not relevant here

qualibration_graphs/quantum_dots/quam_config/wiring_examples/wiring_opxp_external_mixers.py
```diff
@@ -0,0 +1,64 @@
+import matplotlib.pyplot as plt
```
TheoLaudatQM: Let's remove this one for now as we are focusing on the OPX1000

qualibration_graphs/quantum_dots/quam_config/wiring_examples/wiring_opxp_octave.py
```diff
@@ -0,0 +1,65 @@
+import matplotlib.pyplot as plt
```
TheoLaudatQM: Let's remove this one for now as we are focusing on the OPX1000

qualibration_graphs/quantum_dots/validation_utils/charge_stability/default.py
```diff
@@ -0,0 +1,184 @@
+from time import perf_counter
```
TheoLaudatQM: This should probably belong to `populate_quam.py`

qualibration_graphs/quantum_dots/validation_utils/time_dynamics/__init__.py
```diff
@@ -0,0 +1,8 @@
+"""Time dynamics validation utilities for quantum dots."""
```
TheoLaudatQM: Do we really need this time_dynamics folder here at this stage?
To me this is for simulating the physics using Seb's simulator, no?
If so then I am in favor of adding it in a second stage

qualibration_graphs/quantum_dots/quam_config/wiring_utils/connectivity_qd.py
```diff
@@ -0,0 +1,76 @@
+from typing import List
```
TheoLaudatQM: These wiring_utils functions should belong to qualang-tools

TheoLaudatQM (changes requested):
Let's try to clean-up the directory by moving what is relevant to the corresponding repos (qualang-tools, quam-builder, qualibration-libs).
Also let's focus on the bare minimum needed to set up the calibration nodes and keep the other things (physics simulator, timing report,...) for later in a future release for simplicity sakes.
I will add the generate_quam framework that I have and let's see if we can use it here

qualibration_graphs/quantum_dots/quam_config/wiring_examples/wiring_mwfem_cross_resonance.py
KU-QM: removed

qualibration_graphs/quantum_dots/quam_config/wiring_examples/wiring_opxp_external_mixers.py
KU-QM: removed

qualibration_graphs/quantum_dots/quam_config/wiring_examples/wiring_opxp_octave.py
KU-QM: removed

qualibration_graphs/quantum_dots/validation_utils/charge_stability/default.py
KU-QM: I think this is also for the simulator, as it sets up QArray. This can be removed for now, and we can add it back at some point when we need it

qualibration_graphs/quantum_dots/validation_utils/time_dynamics/__init__.py
KU-QM: removed

qualibration_graphs/quantum_dots/quam_config/wiring_utils/connectivity_qd.py
KU-QM: This confuses me - is this connectivity actually used anywhere? I see in the quam generators that most of them are getting the connectivity from qualang-tools anyway, which already includes a Quantum Dots Connectivity. The only difference here seems to be that the naming is "PLUNGER" vs "PLUNGER_GATE", and I don't see it used anywhere. I think we can just delete it and use the one already in qualang-tools

TheoLaudatQM: I am not sure, since you added it, I thought that it was because the components defined in qualang-tools were not good enough so you had to update them locally.
If it is not the case, then yes let's try to build everything from the qualang-tools wiring tool

qualibration_graphs/quantum_dots/quam_config/populate_quam_lf_mw_fems.py
```diff
@@ -0,0 +1,133 @@
...
+
+# XY / MW-FEM: QuAM uses IF = larmor_frequency - MW_upconverter (see XYDriveMW).
+# QM enforces |IF| <= 500 MHz. The old name ``LO`` here was really the *Larmor*
+# centre (~9.7 GHz), not the FEM LO; leaving upconverter at ~5 GHz made IF ~4.7 GHz.
+larmor_center_hz = 9.5e9
+mw_upconverter_hz = larmor_center_hz
+
+qubit_frequencies = [
+    i * 5e6 + larmor_center_hz for i, q in enumerate(machine.qubits.keys())
+]
```
TheoLaudatQM: Is it really the most convenient way of setting the qubit frequencies?
The 5MHz splitting seems pretty arbitrary, so wouldn't it make more sense to just set the frequencies as a list [9.1e9, 9.245e9, 9.647e9, 9.8e9] and then set the upconverter frequency so that all the IF are within +/- 400MHz?

qualibration_graphs/quantum_dots/quam_config/populate_quam_lf_mw_fems.py
```diff
@@ -0,0 +1,133 @@
...
+#################################
+###### Qubits Points Update #####
+#################################
+
+for i, q in enumerate(machine.qubits.values()):
+    q.xy.opx_output.band = 3
+    # Same params for each qubit for now. Subject to change.
+    q.macros[VoltagePointName.INITIALIZE].update(ramp_duration=2000, hold_duration=200)
+    q.macros[VoltagePointName.MEASURE].update(buffer_duration=240)
+    q.macros[VoltagePointName.EMPTY].update(hold_duration=80)
```
TheoLaudatQM: How can we access the meaning of those parameters easily? For instance, I don't know what `buffer_duration` is, and there is no direct docstring accessible from here, is there an API for customers to get it?

qualibration_graphs/quantum_dots/quam_config/populate_quam_lf_mw_fems.py
```diff
@@ -0,0 +1,133 @@
...
+
+    # MW FEM LO on this XY line (shared port → same value each iteration is fine).
+    q.xy.opx_output.upconverter_frequency = mw_upconverter_hz
+
+    # Absolute drive / Larmor frequency (RF), not the OPX IF.
+    q_xy = q.macros[SingleQubitMacroName.XY_DRIVE]
+    q_xy.update(frequency=qubit_frequencies[i])
```
TheoLaudatQM: What is the advantage of doing this instead of the usual `q.xy.RF_frequency=qubit_frequencies[i]`?

qualibration_graphs/quantum_dots/quam_config/populate_quam_lf_mw_fems.py
```diff
@@ -0,0 +1,133 @@
...
+
+    q.xy.operations[f"{DrivePulseName.GAUSSIAN}_x90"].amplitude = 0.17
```
TheoLaudatQM: Why only setting the amplitude of the Gaussian drive and not the other pulses?

qualibration_graphs/quantum_dots/quam_config/populate_quam_lf_mw_fems.py
```diff
@@ -0,0 +1,133 @@
...
+    qdp.add_point(
+        point_name=VoltagePointName.MEASURE,
+        voltages={d.id: (i + 1) * 0.025 for d in qdp.quantum_dots},
+        duration=1000,
+    )
+    qdp.add_point(
+        point_name=VoltagePointName.EXCHANGE,
+        voltages={d.id: (i + 1) * -0.025 for d in qdp.quantum_dots},
+        duration=1000,
+    )
```
TheoLaudatQM: This programmatic way of defining the points is nice, but is it really usable in practice?
There is no reason that those points are equally spaced, yes?

qualibration_graphs/quantum_dots/quam_config/populate_quam_lf_mw_fems.py
```diff
@@ -0,0 +1,133 @@
...
+
+##############################
+###### Sensor Properties #####
+##############################
+
+resonator_frequencies = [i*5e6 + 250e6 for i, s in enumerate(machine.sensor_dots.keys())]
```
TheoLaudatQM: Same here, I think that it would make more sense to manually specify the list of frequencies

TheoLaudatQM (changes requested):

qualibration_graphs/quantum_dots/quam_config/populate_quam_lf_mw_fems.py
KU-QM: I wanted this script to be able to run with no constraints on the number of qubits in the state, hence the generator. Since this is an example that the user would fill out themselves, I thought this could be somehting that they could manually edit. Do you think this is a bit confusing?

qualibration_graphs/quantum_dots/quam_config/populate_quam_lf_mw_fems.py
KU-QM: Commented this for now. We can either add docstrings in quam-builder, or add comments to explain here, or just leave this out since these are just macro default overrides and not strictly necessary

qualibration_graphs/quantum_dots/quam_config/populate_quam_lf_mw_fems.py
KU-QM: Now uses larmor_frequency instead.

qualibration_graphs/quantum_dots/quam_config/populate_quam_lf_mw_fems.py
KU-QM: Now cycles through the enum and changes the x90 and x180 for each

qualibration_graphs/quantum_dots/quam_config/populate_quam_lf_mw_fems.py
KU-QM: Commented for now. I think this is useful for us debugging, as some nodes specifically require these points to exist. We don't want to be manually adding these points every time we create a new quam state for testing

qualibration_graphs/quantum_dots/quam_config/wiring_examples/wiring_lffem_mwfem_qdac.py
```diff
@@ -0,0 +1,117 @@
...
+connectivity.add_voltage_gate_lines("source", name="", constraints=qdac2_spec(index=2, out_port=10))
+connectivity.add_voltage_gate_lines("drain", name="", constraints=qdac2_spec(index=2, out_port=11))
+# Add sensor dots
+connectivity.add_sensor_dot_voltage_gate_lines(sensor_dots, constraints=qdac_lf_spec)
+# Add resonators
+connectivity.add_sensor_dot_resonator_line(sensor_dots, shared_line=False)
+# Add drive lines
+connectivity.add_quantum_dot_drive_lines(plunger_dots, shared_line=True, use_mw_fem=True)
+# Add the barrier gates for each quantum dot pair
+connectivity.add_quantum_dot_pairs(quantum_dot_pairs, constraints=lf_fem_spec(out_slot=6))
```
KU-QM: Is this constraint just for an example of how to set an FEM constraint?

TheoLaudatQM: yes sort of and show that you can parametrize where the different channels are, here plungers and sensors are on the 1st fem while the barriers are on the second one

qualibration_graphs/quantum_dots/quam_config/populate_quam_lf_mw_fems.py
TheoLaudatQM: But the input parameters are macros dependent, no?
If so how would you pass the docstring to the user here?

KU-QM: It's something to think about - I agree that having a list of the possible arg inputs would be very useful for the user to be able to add their own macro overrides. I think we should see (maybe not now) whether we can add some sort of documentation to each macro object so that the update function's docstring can be displayed on the IDE or something. Not sure I have a good suggestion for now

TheoLaudatQM: Yeah it is definitely something to think about, because I have no idea how to quickly find where a specific macros is defined and what are the parameters exposed to the users at the moment.
To me this shouldn't be part of populate quam, but directly on quam-builder so that the relevant information can be accessed from any node

qualibration_graphs/quantum_dots/quam_config/README.md
```diff
@@ -0,0 +1,92 @@
...
+- Copy the entire content of the chosen example script into `generate_quam.py`.
+- Modify the `generate_quam.py` script:
+  - Adjust static parameters like IP addresses and cluster names.
+  - Define the available instruments in your setup.
+  - Specify the qubit IDs present in your system.
+  - Define any custom channel addresses if necessary.
+  - Review the wiring allocation logic and adjust if needed.
+- Run the script: Execute `python generate_quam.py`. This uses the definitions in the script (leveraging `quam-builder` tools internally) to generate the QUAM state files (e.g., `wiring.json` and `state.json`) in your QUAM state folder (default: `quantum_dots/quam_state`). The `state.json` created at this stage contains the structure but is largely empty of specific operational parameters. This generated QUAM state should typically be populated with target values in the next step.
+
+![opx1000_wiring](../.img/opx1000_wiring.PNG)
```
TheoLaudatQM: @KU-QM I believe that the figure is missing

TheoLaudatQM (approved):

PR #539, opened 2026-07-15
loss_divincenzo: resonator spectroscopy nodes (02a/02b/02c)

KU-QM:
Part 3 of the quantum_dots migration series (see #537, #538).

Updates `02a_resonator_spectroscopy`, `02b_resonator_spectroscopy_vs_power`, and adds new `02c_resonator_spectroscopy_vs_detuning`, plus their `calibration_utils` and tests/artifacts.

Made with [Cursor](https://cursor.com)

PR #540, opened 2026-07-15
loss_divincenzo: sensor gate sweep nodes (03a/03a_opx/03c)

KU-QM:
Part 4 of the quantum_dots migration series (see #537-#539).

Updates `03a_sensor_gate_sweep`, adds `03a_sensor_gate_sweep_opx` and `03c_sensor_gate_sweep_dac`, plus their `calibration_utils`.

Note: removed a hardware-specific external DAC/gate-manager driver dependency that had leaked into `charge_stability_dac`'s production code; the affected hardware-execution call site in `03c` now raises `NotImplementedError` with a comment on what to plug in. Simulation/analysis paths are unaffected.

Made with [Cursor](https://cursor.com)


qualibration_graphs/quantum_dots/calibrations/loss_divincenzo/03a_sensor_gate_sweep_opx.py
```diff
@@ -0,0 +1,312 @@
...
+    log_fitted_results,
+    generate_simulated_dataset,
+    plot_all,
+)
+from calibration_utils.common_utils.experiment import get_sensors
+from qualibration_libs.runtime import simulate_and_plot
+from qualibration_libs.data import XarrayDataFetcher
+
+# %% {Node initialisation}
+description = """
```
KU-QM, line 30: Potentially also add option to step to detuning point at the readout moment

qualibration_graphs/quantum_dots/calibrations/loss_divincenzo/03a_sensor_gate_sweep_opx.py
```diff
@@ -0,0 +1,312 @@
...
+        #   n_st         : stream reporting shot index to PC (progress bar)
+        I, I_st, Q, Q_st, n, n_st = node.machine.declare_qua_variables(num_IQ_pairs=num_sensors)
+        # Real-time variable holding the plunger gate voltage
+        offset = declare(fixed)
+
+        # If several sensors share the same AWG resources, they are grouped into batches
+        for multiplexed_sensors in sensors.batch():
+            align()  # sync all channels in this batch before starting
+
+            # Extract the VoltageSequence objects in this batch
```
KU-QM: This might be unintuitive for the user

qualibration_graphs/quantum_dots/calibrations/loss_divincenzo/03a_sensor_gate_sweep_opx.py
```diff
@@ -0,0 +1,308 @@
...
+        # Allocate real-time variables on the OPX:
+        #   I[i], Q[i]   : demodulated quadratures for sensor i
+        #   I_st[i], Q_st[i] : buffers collecting I/Q before transfer to PC
+        #   n            : shot counter
+        #   n_st         : stream reporting shot index to PC (progress bar)
+        I, I_st, Q, Q_st, n, n_st = node.machine.declare_qua_variables(num_IQ_pairs=num_sensors)
+        # Real-time variable holding the plunger gate voltage
+        offset = declare(fixed)
+
+        # If several sensors share the same AWG resources, they are grouped into batches
```
TorbjornRasmussenQM, line 132: AWG resources -> OPX resources

TorbjornRasmussenQM (approved):

qualibration_graphs/quantum_dots/calibrations/loss_divincenzo/03a_sensor_gate_sweep_opx.py
```diff
@@ -0,0 +1,308 @@
...
+
+                        # Play the "readout" pulse and integrate I/Q into I[i], Q[i]
+                        sensor.readout_resonator.measure("readout", qua_vars=(I[i], Q[i]))
+                        # Append this voltage point's I/Q to the stream buffer
+                        save(I[i], I_st[i])
+                        save(Q[i], Q_st[i])
+                    align()
+
+                # At the end of each 1D sweep, play a compensation pulse to account for any charge build-up in the bias tee
+                apply_compensation_pulse(multiplexed_sensors, node.parameters.max_compensation_voltage)
```
TheoLaudatQM: I am challenging the necessity of having a helper_utils.py file under calibration_utils.
For instance here why not simply do the following, which is also more explicit for the user?

```
    for i, sensor in multiplexed_sensors.items():
        sensor.voltage_sequence.apply_compensation_pulse(node.parameters.max_compensation_voltage)
```

qualibration_graphs/quantum_dots/calibrations/loss_divincenzo/03a_sensor_gate_sweep_opx.py
```diff
@@ -0,0 +1,308 @@
...
+        #   I_st[i], Q_st[i] : buffers collecting I/Q before transfer to PC
+        #   n            : shot counter
+        #   n_st         : stream reporting shot index to PC (progress bar)
+        I, I_st, Q, Q_st, n, n_st = node.machine.declare_qua_variables(num_IQ_pairs=num_sensors)
+        # Real-time variable holding the plunger gate voltage
+        offset = declare(fixed)
+
+        # If several sensors share the same OPX resources, they are grouped into batches
+        for multiplexed_sensors in sensors.batch():
+            refresh_voltage_sequences(node, multiplexed_sensors)
```
TheoLaudatQM: I am challenging the necessity of having a helper_utils.py file under calibration_utils.
For instance here why not simply do the following, which is also more explicit for the user?
```
        for multiplexed_sensors in sensors.batch():
            for i, sensor in multiplexed_sensors.items():
                node.machine.reset_voltage_sequence(sensor.voltage_sequence.gate_set.id, track_integrated_voltage=True)
```

KU-QM: The intent was: 
1. sensor.voltage_sequence.gate_set.id/name might have been unintuitive for the user, so I thought we could abstract it away. Same reason why in later nodes we have extract_vgs_id function
2. Additionally, in-case the user wanted to use sensors from different gate-sets, although admittedly this is a secondary concern

qualibration_graphs/quantum_dots/calibrations/loss_divincenzo/03a_sensor_gate_sweep_opx.py
KU-QM: helper_utils removed

KU-QM:
@TheoLaudatQM removed helper_utils, and as discussed, added to common_utils/experiment a common check. This function: 
- Extracts the gate set ID for each element in the Sequence. The intent is to input list of sensors/dots etc directly
- Optionally allows you to reset the voltage sequence associated with the gate set
- Throws an error if the len of the set of vgs IDs is not exactly 1

Once this PR is merged, we can add this to the other nodes (<=06)

PR #542, opened 2026-07-15
loss_divincenzo: charge stability and readout time optimization nodes (05a-05e)

KU-QM:
Part 6 of the quantum_dots migration series (see #537-#541).

Updates `05a_charge_stability_opx`, `05b_charge_stability_qdac`, `05d_charge_stability_demo`, adds `05c_charge_state_readout_time_optimization` and `05e_charge_stability_opx_dac`, plus their `calibration_utils` and tests.

Note: removed the same hardware-specific DAC/gate-manager driver dependency as #540 (this branch pulls in a fresh copy of that shared module); the `05e` hardware-execution call site now raises `NotImplementedError` with a note on what to plug in. Simulation/analysis paths are unaffected.


qualibration_graphs/quantum_dots/calibration_utils/charge_stability/parameters.py
```diff
@@ -0,0 +1,185 @@
...
+    y_points: int = 121
+    """Number of measurement points in the Y axis."""
+    x_span: float = 1
+    """The X axis span in volts"""
+    y_span: float = 1
+    """The Y axis span in volts"""
+    ramp_duration: int = 100
+    """The ramp duration to each pixel. Set to zero for a step."""
+    hold_duration: int = 100
+    """Dwell time on each point in nanoseconds. If using the QDAC, this must be slow enough."""
```
ChloeBO-QM: "slow enough" is not very specific. Is it possible to give a duration or at least an order of magnitude?

qualibration_graphs/quantum_dots/calibration_utils/charge_stability/parameters.py
```diff
@@ -0,0 +1,185 @@
...
+    per_line_compensation: bool = True
+    """Whether to send a compensation pulse at the end of each scan line."""
+    max_compensation_voltage: float = 0.05
+    """The maximum voltage for the compensation pulse."""
+    x_center: Optional[float] = None
+    """The center of the X axis sweep. If dc_control = True, then this will be applied to the external source. Else, it will be applied by the OPX."""
+    y_center: Optional[float] = None
+    """The center of the Y axis sweep. If dc_control = True, then this will be applied to the external source. Else, it will be applied by the OPX."""
```
ChloeBO-QM: Based on the code in `05a_charge_stability_opx.py` it seems that `dc_control` should be a parameter defined here. However, it seems to be missing from the list of node parameters.

qualibration_graphs/quantum_dots/calibration_utils/charge_stability/parameters.py
```diff
@@ -0,0 +1,185 @@
...
+    NodeSpecificParameters,
+):
+    x_from_qdac: bool = False
+    "Check to perform 2D map using the QDAC instead of the OPX"
+    y_from_qdac: bool = False
+    "Check to perform 2D map using the QDAC instead of the OPX"
+    post_trigger_wait_ns: int = 10000
+    """A pause in the QUA programme to allow the QDAC to get to the correct level."""
+    qdac_dwell_time_us: float = 200
+    """The dwell time in microseconds for the QDAC."""
```
ChloeBO-QM: Does this serve the same purpose as `hold_duration` in the main parameter list (but for the qdac instead of the opx)? If so, perhaps change the name of the variable to `qdac_hold_duration` for consistency.

PR #545, opened 2026-07-15
loss_divincenzo: qubit spectroscopy nodes (08a/08b/08c)

KU-QM:
Part 9 of the quantum_dots migration series (see #537-#544).

Adds `08a_qubit_spectroscopy_chirp`, `08b_qubit_spectroscopy`, `08c_qubit_initialisation_optimiser_init_duration`, plus their `calibration_utils` and tests/artifacts.

Note: the old `08a_power_rabi`/`08b_power_rabi_error_amplification` nodes move to category 09 in the next PR, so they're left untouched here to keep that rename self-contained.

PR #546, opened 2026-07-15
loss_divincenzo: power Rabi and time Rabi nodes (09a/09b/10a/10b)

KU-QM:
Part 10 of the quantum_dots migration series (see #537-#545).

Renumbers power Rabi from category 08 to 09 (`09a_power_rabi`, adds `09b_power_rabi_error_amplification`) and time Rabi from 09 to 10 (`10a_time_rabi`, `10b_time_rabi_chevron`, dropping the `_parity_diff` suffix), plus their `calibration_utils` and tests/artifacts.

Note: the T1/Ramsey nodes currently at 10/10a/10b/10c are untouched here — they move to category 11 in the next PR, kept separate to keep each rename self-contained.


qualibration_graphs/quantum_dots/calibration_utils/common_utils/plot_style.py
```diff
@@ -0,0 +1,67 @@
+"""Shared matplotlib styling for quantum-dots calibration plots."""
```
KU-QM, line 1: I am a bit confused about this file. This is meant to set a standard plotting style for all the nodes? Should this be standardised across all the nodes? These being used in select calibration scripts would lead to a lack of uniformity. Are we going to re-apply this uniformity at the end, in release/nightly?

qualibration_graphs/quantum_dots/calibration_utils/power_rabi_error_amplification/parameters.py
```diff
@@ -0,0 +1,37 @@
...
+    """Minimum amplitude prefactor. Narrow window around expected a_π after node 09a."""
+    max_amp_factor: float = 1.15
+    """Maximum amplitude prefactor. Narrow window around expected a_π after node 09a."""
+    amp_factor_step: float = 0.001
+    """Step size for the amplitude prefactor sweep. Default is 0.001."""
+    max_n_pulses: int = 40
+    """Number of pulses in the error-amplified power Rabi pulse sequence."""
+    operation: Literal["x180", "x90"] = "x180"
+    """The operation to perform to drive the qubit."""
+    parity_measurement: bool = False
```
KU-QM, line 23: This should not be necessary, seeing as ParityDiffAnalysisParameters (subclassed below) already includes this

qualibration_graphs/quantum_dots/calibration_utils/power_rabi/parameters.py
```diff
@@ -0,0 +1,37 @@
...
+    """Number of averages to perform. Default is 100."""
+    min_amp_factor: float = 0.001
+    """Minimum amplitude factor for the operation. Default is 0.001."""
+    max_amp_factor: float = 1.99
+    """Maximum amplitude factor for the operation. Default is 1.99."""
+    amp_factor_step: float = 0.01
+    """Step size for the amplitude factor. Default is 0.01."""
+    operation: Literal["x180", "x90"] = "x180"
+    """The operation to perform to drive the qubit."""
+    parity_measurement: bool = False
```
KU-QM, line 21: Should not be necessary based on ParityDiffAnalysisParameters

qualibration_graphs/quantum_dots/calibration_utils/power_rabi/plotting.py
```diff
@@ -0,0 +1,263 @@
...
+
+from calibration_utils.common_utils.plot_style import (
+    apply_qubit_outcome_style,
+    empty_figure,
+    qubit_success,
+)
+from calibration_utils.power_rabi.analysis import FFT_FREQ_MIN, FFT_FREQ_MAX, compute_fft_diagnostic
+
+
+def _get_qubit_names_from_ds(
```
KU-QM, line 20: I wonder if this belongs in measurement_utils

qualibration_graphs/quantum_dots/calibration_utils/power_rabi_error_amplification/analysis.py
```diff
@@ -0,0 +1,367 @@
...
+            "mean_signal": mean_signal,
+            "mean_signal_fit": mean_signal_fit,
+            "resonance_idx": resonance_idx,
+        },
+    }
+
+
+def process_raw_dataset(ds: xr.Dataset, node: QualibrationNode) -> xr.Dataset:
+    """Build conditional-expectation variables from joint-outcome streams in ``ds_raw``."""
+    from calibration_utils.measurement_utils.measurement_streams import process_streams
```
KU-QM, line 227: Any reason not to have this at the top?

qualibration_graphs/quantum_dots/calibration_utils/time_rabi/parameters.py
```diff
@@ -0,0 +1,35 @@
...
+    """Number of averages to perform. Default is 100."""
+    min_wait_time_in_ns: int = 16
+    """Minimum pulse duration in nanoseconds. Must be larger than 4 clock cycles. Default is 16 ns."""
+    max_wait_time_in_ns: int = 10_000
+    """Maximum pulse duration in nanoseconds. Default is 10000 ns (10 us)."""
+    time_step_in_ns: int = 52
+    """Step size for the pulse duration sweep in nanoseconds. Default is 52 ns."""
+    operation: Literal["x180", "x90"] = "x180"
+    """The operation to perform to drive the qubit."""
+    parity_measurement: bool = False
```
KU-QM, line 21: Already in ParityDiffAnalysisParameters

qualibration_graphs/quantum_dots/calibration_utils/time_rabi_chevron/init_utils.py
```diff
@@ -0,0 +1,306 @@
...
+
+import numpy as np
+from scipy.optimize import curve_fit
+
+_logger = logging.getLogger(__name__)
+
+FFT_FREQ_MIN = 0.0005  # 1/ns (exclude DC) → t_π max ~1000 ns
+FFT_FREQ_MAX = 0.03  # 1/ns → t_π min ~17 ns; restricts to plausible Rabi range
+
+
```
KU-QM, line 31: Why is this not part of analysis.py? If it needs to be its own file, then init_utils is not an intuitive name for what it is doing

qualibration_graphs/quantum_dots/calibrations/loss_divincenzo/09a_power_rabi.py
```diff
@@ -0,0 +1,297 @@
...
+from qualibration_libs.data import XarrayDataFetcher
+from qualibration_libs.parameters.experiment import get_qubits
+from qualibration_libs.runtime import simulate_and_plot
+
+# %% {Node initialisation}
+description = """
+        POWER RABI
+This sequence parks the qubit at the manipulation bias point, plays the selected qubit operation (e.g. x180) at
+different amplitude prefactors, and measures the spin state. Joint-outcome streams are averaged and reduced to
+conditional expectations for analysis. Rabi oscillations in the analysis signal versus amplitude prefactor are
```
KU-QM, line 37: The line "Joint-outcome streams are averaged and reduced to
conditional expectations for analysis." is specific to parity measurements, however the user can choose which measurement type to use. Maybe make this line more generic

qualibration_graphs/quantum_dots/calibrations/loss_divincenzo/09a_power_rabi.py
```diff
@@ -0,0 +1,297 @@
...
+
+        # Real-time variables:
+        # a  : current amplitude prefactor for the manipulation gate
+        # n  : shot counter
+        # p2 : post-manipulation measurement outcome (0 = empty, 1 = loaded)
+        # p1 : pre-manipulation measurement outcome (only used when parity_measurement=True)
+        a = declare(fixed)
+        n = declare(int)
+
+        p2, p1, parity_streams = declare_streams(node, qubits, stream_fn=declare_output_stream)
```
KU-QM, line 126: Might be worth adding a line here specifying that if parity_measurement = False, then p1 = None. 

Also perhaps consider changing p2,p1 to p1,p0 to match the measurement_utils version

qualibration_graphs/quantum_dots/calibrations/loss_divincenzo/09a_power_rabi.py
```diff
@@ -0,0 +1,297 @@
...
+
+                # ── INNER LOOP: sweep amplitude prefactor ────────────────
+                with for_(*from_array(a, amps)):
+
+                    # Optional pre-measurement at the empty bias point (parity readout)
+                    if node.parameters.parity_measurement:
+                        qubit.empty()
+                        a1 = qubit.measure()
+
+                    # Heralded initialization to the target spin state at the manipulation bias
```
KU-QM, line 144: Something to consider - should we do this in the preparation Python part as a tracked change instead of having these as run-time args? Could be easier, and would allow us to use qubit.initialize() with no args

qualibration_graphs/quantum_dots/calibrations/loss_divincenzo/09a_power_rabi.py
```diff
@@ -0,0 +1,297 @@
...
+
+# %% {Update_state}
+@node.run_action(skip_if=node.parameters.simulate)
+def update_state(node: QualibrationNode[Parameters, Quam]):
+    """Update the relevant parameters if the qubit data analysis was successful."""
+    with node.record_state_updates():
+        for q in node.namespace["qubits"]:
+            if node.outcomes[q.name] == "failed":
+                continue
+
```
KU-QM, line 287: Confused about this. We are allowing the user to pick the operation to be either x90 or x180 (based on Parameters class Literal attribute type), but the extracted opt_amp is for a half-period (pi-pulse). I don't see any specific handling of the power in the case of an x90. If the user picks x90, then this means that they will be setting the x90 power as the pi pulse (x180) power? This might be a bug

qualibration_graphs/quantum_dots/calibrations/loss_divincenzo/09a_power_rabi.py
```diff
@@ -0,0 +1,297 @@
...
+            opt_prefactor = node.results["fit_results"][q.name]["opt_amp"]
+            q.macros[node.parameters.operation].update(amplitude_scale=opt_prefactor)
+            if node.parameters.operation == "x180":
+                q.macros["x90"].update(amplitude_scale=opt_prefactor / 2)
```
KU-QM, line 291: Do we also want to keep this auto-update of x90? Maybe the user would like to calibrate them separately

qualibration_graphs/quantum_dots/calibrations/loss_divincenzo/10a_time_rabi.py
```diff
@@ -0,0 +1,295 @@
...
+
+
+# Any parameters that should change for debugging purposes only should go in here
+# These parameters are ignored when run through the GUI or as part of a graph
+@node.run_action(skip_if=node.modes.external)
+def custom_param(node: QualibrationNode[Parameters, Quam]):
+    """Allow the user to locally set the node parameters for debugging purposes, or execution in the Python IDE."""
+    # You can get type hinting in your IDE by typing node.parameters.
+    node.parameters.qubits = ["q1", "q2"]
+    node.parameters.use_simulated_data = True
```
KU-QM, line 88: comment out for uniformity

qualibration_graphs/quantum_dots/calibrations/loss_divincenzo/10a_time_rabi.py
```diff
@@ -0,0 +1,295 @@
...
+    """Build the 1D pulse-duration sweep and the QUA pulse sequence."""
+    # ── Experiment parameters (Python side) ──────────────────────────────
+
+    node.namespace["qubits"] = qubits = get_qubits(node)
+
+    n_avg = node.parameters.num_shots  # repetitions averaged at each duration point
+    operation = node.parameters.operation  # qubit gate whose duration is swept (x180 or x90)
+
+    # Duration axis [ns]: sweep around the expected π-pulse length (quantised to 4 ns on the OPX)
+    pulse_durations = np.arange(
```
KU-QM, line 108: In the PSB nodes, I have been adding a 4ns granularity check via a function in calibration_utils/.../helper_utils.py. Maybe we can make a global 4ns granularity checker in common_utils

KU-QM (commented):
Biggest issue I think is the handling of the x90

qualibration_graphs/quantum_dots/calibration_utils/common_utils/plot_style.py
TheoLaudatQM: Yes that was my idea in order to create homogeneous figures when possible and specify whether the fit converged or not on the figure directly and I believe it can be quite handy

qualibration_graphs/quantum_dots/calibration_utils/time_rabi_chevron/init_utils.py
TheoLaudatQM: Good catch this was actually how it was done at Quantum Motion and I overlooked it during the review. I refactored it in analysis.py

qualibration_graphs/quantum_dots/calibration_utils/power_rabi/plotting.py
TheoLaudatQM: Yep, I uniformized it and fixed it

qualibration_graphs/quantum_dots/calibrations/loss_divincenzo/09a_power_rabi.py
TheoLaudatQM: Yes good point!

qualibration_graphs/quantum_dots/calibrations/loss_divincenzo/09a_power_rabi.py
TheoLaudatQM: fixed as discussed

qualibration_graphs/quantum_dots/calibrations/loss_divincenzo/09a_power_rabi.py
TheoLaudatQM: fixed as discussed

qualibration_graphs/quantum_dots/calibrations/loss_divincenzo/09a_power_rabi.py
TheoLaudatQM: Fiixed as discussed

qualibration_graphs/quantum_dots/calibrations/loss_divincenzo/09a_power_rabi.py
TheoLaudatQM: I removed it as we said that parity meas will be removed eventually.

qualibration_graphs/quantum_dots/calibrations/loss_divincenzo/10a_time_rabi.py
TheoLaudatQM: Let's focus on this during the uniformity section to make sure that this is handled the same way accross all the node

qualibration_graphs/quantum_dots/calibrations/loss_divincenzo/09a_power_rabi.py
```diff
@@ -0,0 +1,270 @@
...
+                    # Perform the initialize macro
+                    qubit.initialize()
+                    align()
+
+                    # Play the x180 gate at the current amplitude prefactor
+                    qubit.x180(amplitude_scale=a)
+                    align()
+
+                    # Thresholded PSB readout → averaged state probability
+                    assign(state[i], Cast.to_int(qubit.measure()))
```
KU-QM: Do we not want to expose the qubit.measure() macro as a separate line?

qualibration_graphs/quantum_dots/calibrations/loss_divincenzo/09b_power_rabi_error_amplification.py
```diff
@@ -0,0 +1,275 @@
...
+import xarray as xr
+
+from qm.qua import *
+
+from qualang_tools.loops import from_array
+from qualang_tools.multi_user import qm_session
+from qualang_tools.results import progress_counter
+
+from qualibrate.core import QualibrationNode
+from quam_config import Quam
```
KU-QM: Should this not be 
`from quam_config import QubitQuam as Quam`? 
quam_config's Quam class is subclassing BaseQuamQD, which should not have qubits

qualibration_graphs/quantum_dots/calibrations/loss_divincenzo/09a_power_rabi.py
```diff
@@ -0,0 +1,270 @@
...
+                    save(state[i], state_st[i])
+
+                    # Return gate voltages to zero before the next shot to avoid accumulation of fixed point errors
+                    align()
```
KU-QM, line 146: This align could potentially be confusing if we don't explicitly use the measure macro on a new line

qualibration_graphs/quantum_dots/calibrations/loss_divincenzo/09b_power_rabi_error_amplification.py
```diff
@@ -0,0 +1,275 @@
...
+    """Update the relevant parameters if the qubit data analysis was successful."""
+    with node.record_state_updates():
+        for q in node.namespace["qubits"]:
+            if node.outcomes[q.name] == "failed":
+                continue
+
+            opt_prefactor = node.results["fit_results"][q.name]["opt_amp"]
+            q.macros[node.parameters.operation].update(amplitude_scale=opt_prefactor)
+            if node.parameters.operation == "x180":
+                q.macros["x90"].update(amplitude_scale=opt_prefactor / 2)
```
KU-QM, line 275: So we automatically update x90 whenever we calibrate the x180, but not the other way around?

qualibration_graphs/quantum_dots/calibrations/loss_divincenzo/10a_time_rabi.py
```diff
@@ -0,0 +1,268 @@
...
+# %% {Node initialisation}
+description = """
+        TIME RABI
+After initialization to the target spin state, this sequence applies an XY drive pulse whose duration is
+swept and measures the spin state with thresholded PSB readout. Averaged state probabilities versus pulse duration
+are fitted to extract the π-pulse duration.
+
+Prerequisites:
+    - Having calibrated the resonators coupled to the sensor dots.
+    - Having calibrated the voltage points (empty, initialization, measurement).
```
KU-QM: We no longer need empty if we have removed the parity diff stuff

qualibration_graphs/quantum_dots/calibrations/loss_divincenzo/10b_time_rabi_chevron.py
```diff
@@ -0,0 +1,297 @@
...
+                save(n, n_st)
+
+                # ── MIDDLE LOOP: sweep drive detuning ────────────────────
+                with for_(*from_array(df, dfs)):
+
+                    # ── INNER LOOP: sweep pulse duration in clock cycles ─────────────────
+                    with for_(*from_array(t, pulse_durations // 4)):
+
+                        # Retune the XY drive to (calibrated IF + df)
+                        qubit.xy.update_frequency(intermediate_frequency + df)
```
KU-QM: I think we should update to the stored intermediate frequency, initialize, and then detune. Otherwise the initialize macro will also be detuned, which might mess up the initialization

qualibration_graphs/quantum_dots/calibrations/loss_divincenzo/10b_time_rabi_chevron.py
```diff
@@ -0,0 +1,297 @@
...
+        for qubit in node.namespace["qubits"]:
+            if node.outcomes[qubit.name] == "failed":
+                continue
+
+            fit_result = node.results["fit_results"][qubit.name]
+            qubit.macros[node.parameters.operation].update(
+                duration=fit_result["optimal_duration"],
+            )
+            # Shift Larmor by the offset between fitted resonance and the current XY IF.
+            qubit.larmor_frequency += fit_result["optimal_frequency"] - qubit.xy.intermediate_frequency
```
KU-QM, line 300: Maybe I'm being stupid - why are we subtracting the IF here?

qualibration_graphs/quantum_dots/calibration_utils/power_rabi/analysis.py
```diff
@@ -0,0 +1,367 @@
...
+    """Return ``ds_raw`` unchanged (thresholded ``state`` needs no stream post-processing)."""
+    return ds
+
+
+def fit_raw_data(
+    ds: xr.Dataset,
+    node: QualibrationNode,
+) -> Tuple[xr.Dataset, Dict[str, Dict[str, Any]]]:
+    """Fit optimal amplitude per qubit from power-Rabi ``state`` data."""
+    qubit_names = [str(v) for v in ds.qubit.values]
```
KU-QM, line 314: Just to be fully crystal clear, we are relying on the analysis being positional here. I do not want to mix up the qubit's results due to a positional mixup

KU-QM (commented):
The removal of the parity diff stuff looks good. If has passed the analysis and plotting tests without mixing up the qubits, then I think we are good. I think nodes themselves are looking clear

qualibration_graphs/quantum_dots/calibration_utils/power_rabi/analysis.py
TheoLaudatQM: Not really, we are just extracting the list of qubit names, but then the data for each qubit is taken from the dataset so I don't think that there is any risk here unless I missed something

qualibration_graphs/quantum_dots/calibrations/loss_divincenzo/10b_time_rabi_chevron.py
TheoLaudatQM: Conceptually I agree with you, but that might be tricky to keep track of as the same applies if one changes the x180 amplitude at the begining of a node for instance. Especially since one cannot easily inspect the macros when working on the QUA program. Wouldn't it be cleaner and more robust to ensure that all the right parameters are used in the macro directly? This way it is the responsability of the macros to ensure that the parameters are properly set.

qualibration_graphs/quantum_dots/calibrations/loss_divincenzo/09b_power_rabi_error_amplification.py
TheoLaudatQM: I'll add it to the description

qualibration_graphs/quantum_dots/calibrations/loss_divincenzo/10b_time_rabi_chevron.py
TheoLaudatQM: Nevermind, we'll do as you suggested

qualibration_graphs/quantum_dots/calibrations/loss_divincenzo/10b_time_rabi_chevron.py
TheoLaudatQM: After double checking, fit_result["optimal_frequency"] is q.xy.if+optimal_detuning. This means that to update the total RF_frequency you want to add the optimal detuning only which leads to fit_result["optimal_frequency"] - qubit.xy.intermediate_frequency

PR #560, opened 2026-07-21
loss_divincenzo: Common Utils & Video Mode Utils

KU-QM:


qualibration_graphs/quantum_dots/calibration_utils/parity_diff_utils/parity_streams.py
```diff
@@ -0,0 +1,384 @@
...
+
+    Returns:
+        ``(p2, p1, streams)`` where
+
+        * ``p2`` – QUA ``int`` variable for the post-sequence measurement.
+        * ``p1`` – QUA ``int`` variable for the pre-sequence measurement, or
+          ``None`` when ``parity_pre_measurement`` is ``False``.
+        * ``streams`` – dict mapping stream-key → ``{item.name: stream}``.
+          Keys are ``"p0_p0"``, ``"p0_p1"``, ``"p1_p0"``, ``"p1_p1"`` when
+          ``parity_pre_measurement`` is ``True``, or ``"p"`` otherwise.
```
TheoLaudatQM: Why not using the same indexing convention here (p1/2 for the variable and p0/1 for the streams)?
It would seem to make more sense to either use indices 0 & 1 or 1 & 2 for both, no?

qualibration_graphs/quantum_dots/calibration_utils/parity_diff_utils/parity_streams.py
```diff
@@ -0,0 +1,384 @@
...
+                save(0, streams["p0_p1"][name])
+                save(0, streams["p1_p0"][name])
+                save(1, streams["p1_p1"][name])
+    else:
+        save(p2, streams["p"][name])
+
+
+def buffer_parity_streams(node, name: str, streams: dict, *buffer_dims: int) -> None:
+    """Buffer and save parity streams inside a ``stream_processing()`` block.
+
```
TheoLaudatQM, line 142: Here I would add a usage example for when you do a parity measurement in a 2D sweep, something like:
```
with stream_processing():
    buffer_parity_streams(node, "q1", streams, (21, 52))
```

qualibration_graphs/quantum_dots/calibration_utils/parity_diff_utils/parity_streams.py
```diff
@@ -0,0 +1,384 @@
...
+
+def get_parity_item_names(
+    ds: xr.Dataset,
+    analysis_signal: str = "E_p2_given_p1_0",
+    *,
+    item_names: Optional[Iterable[str]] = None,
+    item_dim: str = "qubit",
+    legacy_prefixes: Sequence[str] = ("p0_p0", "p"),
+) -> list[str]:
```
TheoLaudatQM, line 275: I am not sure what this one is doing and how to use it in practice.
Also, the input variables are missing in the docstring.
I believe that a more detailed explanation and usage example is needed here

qualibration_graphs/quantum_dots/calibration_utils/parity_diff_utils/parity_streams.py
```diff
@@ -0,0 +1,384 @@
...
+
+    - ``E_p2_given_p1_0_{name}`` = P(second=1 | first=0) = p0_p1 / (p0_p0 + p0_p1)
+    - ``E_p2_given_p1_1_{name}`` = P(second=1 | first=1) = p1_p1 / (p1_p0 + p1_p1)
+
+    Division by zero yields NaN.
+
+    Args:
+        ds: Dataset containing ``p0_p0_{name}``, ``p0_p1_{name}``,
+            ``p1_p0_{name}``, ``p1_p1_{name}`` for each name.
+        qubit_names: List of qubit (or qubit-pair) names.
```
TheoLaudatQM, line 371: Some input arguments are missing and I believe that a usage example would be nice as well

qualibration_graphs/quantum_dots/calibration_utils/parity_diff_utils/parity_streams.py
```diff
@@ -0,0 +1,384 @@
...
+
+def process_parity_streams(
+    ds: xr.Dataset,
+    item_names: Iterable[str],
+    *,
+    parity_pre_measurement: bool,
+    item_dim: str = "qubit",
+    sweep_dims: Optional[Sequence[str]] = None,
+) -> xr.Dataset:
+    """Normalize parity streams into conditional-expectation variables.
```
TheoLaudatQM, line 417: The input arguments are missing and I believe that a usage example would be nice as well

qualibration_graphs/quantum_dots/calibration_utils/common_utils/qubit_readout_helper.py
```diff
@@ -0,0 +1,151 @@
...
+
+    return batches
+
+
+def get_qubits_batched_by_readout(
+    node: QualibrationNode,
+    sensor_batches: BatchableList[SensorDot] = None,
+    execution_mode: Literal["parallel", "sequential"] = "sequential",
+) -> BatchableList[AnySpinQubit]:
+    """
```
TheoLaudatQM: A usage example would be nice also showing the structure of the BatchableList for different configuration

qualibration_graphs/quantum_dots/calibration_utils/common_utils/experiment.py
```diff
@@ -0,0 +1,487 @@
+import logging
```
TheoLaudatQM: I believe that those should belong to qualibration_libs, no?

qualibration_graphs/quantum_dots/calibration_utils/common_utils/annotation.py
```diff
@@ -0,0 +1,152 @@
+"""Utilities for annotating matplotlib figures with qualibrate metadata."""
+
+from __future__ import annotations
+
+from typing import Any, Dict, Optional
+
+from matplotlib.figure import Figure
+
+
+def stamp_snapshot(
```
TheoLaudatQM: This should probably belong to qualibration_libs, under the plotting section

qualibration_graphs/quantum_dots/calibration_utils/common_utils/annotation.py
```diff
@@ -0,0 +1,152 @@
+"""Utilities for annotating matplotlib figures with qualibrate metadata."""
```
TheoLaudatQM: I think that all the public functions in this folder should be exposed to users via an __init__ file, no?

qualibration_graphs/quantum_dots/calibration_utils/gate_virtualization/__init__.py
```diff
@@ -0,0 +1,86 @@
+from .base_parameters import GateVirtualizationBaseParameters, get_voltage_arrays
```
TheoLaudatQM: Why not having these different virtualization calibration nodes following the same structure as the other nodes, like having the main node under `calibrations` and the parameters, analysis and plotting under `calibration_utils`?

qualibration_graphs/quantum_dots/calibration_utils/run_video_mode/__init__.py
```diff
@@ -0,0 +1,10 @@
+from .video_mode_specific_parameters import VideoModeCommonParameters
```
TheoLaudatQM: Same here, I am not sure I am following the logic.

1. Why do we need the simulated_video_mode here?
2. Why not having the video_mode parameters under calibration_utils?
3. Why having parameters.py and video_mode_specific_parameters.py?
4. Here you have the qdac_video_mode, but not the opx_video_mode

TheoLaudatQM (changes requested):
I am not sure I fully understand why gate_virtualization and video_mode are set up this way and not like the other nodes are implemented.
In addition, I believe that several of the commo_utils functions should belong to qualibration_utils, especially since many of them are private and not exposed to customers.
In my opinion, only the parity_streams.py and qubit_readout_helper.py belong here, but I would even place them under calibration_utils.

qualibration_graphs/quantum_dots/calibration_utils/run_video_mode/__init__.py
KU-QM: 1. The simulated video mode was first made for the march meeting, working with node 05d_charge_stability_demo. This seemed like a useful addition at the time (similar to the simulated data generators for each node) which is why it was kept in for the QuMo eval. Happy to keep or remove it 
2. Can you clarify - do you mean common_utils? If so, then I think I agree to move it 
3. Initially, the intent was to have a node specifically to run video mode with (node 00_run_video_mode), which is why this directory is called 'run_video_mode' and we have a corresponding parameters.py. I agree that the VideoModeSpecificParameters should probably belong elsewhere, as perhaps this entire directory, if we choose to not include node 00_run_video_mode. 
4. The OPX only version is in video_mode_utils.py. qdac_video_mode was made to maybe eventually have a node that allows you to do a QDAC + OPX video mode. The naming scheme is not great, admittedly

qualibration_graphs/quantum_dots/calibration_utils/gate_virtualization/__init__.py
KU-QM: I think the main nodes would be included in calibrations/gate_virtualization, which are not PR'd yet. The calibration_utils for gate_virtualization is only included here, since the sensor gate sweep node analysis seemed to use it - but may be cleaner to just separate the sensor gate sweep from this directory.

qualibration_graphs/quantum_dots/calibration_utils/common_utils/experiment.py
```diff
@@ -0,0 +1,101 @@
...
+class VideoModeCommonParameters(RunnableParameters):
+    run_in_video_mode: bool = True
+    """Optionally open Video Mode with the qualibration node."""
+    virtual_gate_set_id: Optional[str] = None
+    """Name of the associated VirtualGateSet in your QPU. """
+    video_mode_port: int = 8002
+    """Localhost port to open VideoMode with"""
+    dc_control: bool = False
+    """If an associated external DC offset exists."""
+    result_type: Literal["I", "Q", "Amplitude", "Phase"] = "I"
```
TorbjornRasmussenQM: Should these be in run_video_mode/video_mode_utils.py instead? or are they actually common

qualibration_graphs/quantum_dots/calibration_utils/run_video_mode/__init__.py
```diff
@@ -0,0 +1,10 @@
...
+
+
+def create_video_mode(*args, **kwargs):
+    from .video_mode_utils import create_video_mode as _create_video_mode
+
+    return _create_video_mode(*args, **kwargs)
```
TheoLaudatQM: Why not just importing `create_video_mode` from `.video_mode_utils` here?

qualibration_graphs/quantum_dots/calibration_utils/run_video_mode/__init__.py
TheoLaudatQM: This no longer exist, shall we delete it?

qualibration_graphs/quantum_dots/calibration_utils/run_video_mode/video_mode_utils.py
```diff
@@ -0,0 +1,206 @@
...
+        except Exception:
+            pass
+        _DASHBOARD_SERVER = None
+
+    # 2. Kill any process using the port (works even if server ref is lost)
+    result = subprocess.run(["lsof", "-ti", f":{port}"], capture_output=True, text=True)
+    pids = result.stdout.strip().split()
+    for pid in pids:
+        try:
+            os.kill(int(pid), signal.SIGKILL)
```
TheoLaudatQM, line 53: `os` and `signal` don't exist

qualibration_graphs/quantum_dots/calibration_utils/common_utils/experiment.py
KU-QM: I am open to suggestions. The calibration_utils folder run_video_mode is now intended to be for a node 00_run_video_mode, so mirroring other nodes, it should contain params and helper utils. In the same way as the QuantumDotExperimentNodeParameters, the thinking was that the common node parameters lives here and not in the node-specific folder. However, I can see the argument that it lives in either place

PR #562, opened 2026-07-30
Add OPX+ and Octave to the wiring example together with the populate script

TheoLaudatQM:


qualibration_graphs/quantum_dots/quam_config/wiring_examples/wiring_opxp_octave_qdac.py
```diff
@@ -0,0 +1,137 @@
...
+def qdac_config(ip: str):
+    return {
+        "driver_module": "qcodes_contrib_drivers.drivers.QDevil.QDAC2",
+        "driver_class": "QDac2",
+        "connection": {
+            "visalib": "@py",
+            "address": f"TCPIP::{ip}::5025::SOCKET",
+        },
+        "channel_method": "channel",
+        "accessor": "limited_dc_constant_V",
```
KU-QM, line 20: If it is changed in the lffem/mwfem wiring, then it should be changed here too. 

In fact, why is it "limited_dc_constant_V"? The API I know is just "dc_constant_V"

qualibration_graphs/quantum_dots/quam_config/wiring_examples/wiring_opxp_octave_qdac.py
```diff
@@ -0,0 +1,137 @@
...
+# with lf_fem_spec replaced by opx_spec(con=1).
+connectivity.add_quantum_dot_voltage_gate_lines(
+    1, True, opx_spec(con=1) & qdac2_spec(index=1, out_port=1, trigger_in_port=1)
+)
+connectivity.add_quantum_dot_voltage_gate_lines(
+    2, True, opx_spec(con=1) & qdac2_spec(index=1, out_port=2, trigger_in_port=2)
+)
+connectivity.add_quantum_dot_voltage_gate_lines(3, False, qdac_opx_spec)
+
+# Global gates on QDAC2
```
KU-QM, line 85: I believe we should leave this out until we fix this. Right now, these are counted as valid voltagegate lines, which are added into the GateSet without an OPX output. This must be fixed probably in quam-builder

KU-QM (commented):
Some minor changes. But looks good

qualibration_graphs/quantum_dots/quam_config/wiring_examples/wiring_opxp_octave_qdac.py
TheoLaudatQM: Yes indeed, I just forgot to update it in the OPX+ version.
To be honest I didn't add `limited_dc_constant_V` so I just reverted it back to `dc_constant_V` as it doesn't seem to exist...

qualibration_graphs/quantum_dots/quam_config/wiring_examples/wiring_opxp_octave_qdac.py
TheoLaudatQM: I just fixed it under the release nightly branch of quam builder which now allows global gates and sensor gates to be a qdac channel only (not included in the virtual gate set), which seems to work. I raise an error for plungers and barriers in case a user tries to do it because it creates issues with the way the qubits and qubit pairs are defined.
What do you think?

qualibration_graphs/quantum_dots/quam_config/populate_quam_lf_mw_fems.py
```diff
@@ -152,6 +159,9 @@ def _apply_fem_output_port_delays(machine: Quam) -> None:
     s.readout_resonator.operations["readout"].amplitude = readout_amplitude
     s.readout_resonator.operations["readout"].length = readout_length
 
+for q in machine.qubits.values():
+    q.preferred_readout_quantum_dot = qubit_readout_dot_mapping[q.name]
+
```
KU-QM, line 164: This breaks the chain for me. 
- The mapping needs to be a Quantum Dot (i.e. "virtual_dot_2", or if we want to map using the qubit name, machine.qubits["q.name"].quantum_dot). This is because the ancillary quantum dot paired to the qubit for readout is not necessarily a qubit by itself

KU-QM:
Rebased version is merged in PR #591

PR #567, opened 2026-08-20
Upgrade resonator spectroscopy fitting to be robust for bring-up and modular for routine use

ohad-lev-qm:
- Replace the peaks_dips-based fit with a bring-up-grade Lorentzian-dip fitter (analysis.py):
    - Noise-relative dip detection with background removal.
    - A fit ladder over window size and linear/quadratic backgrounds.
    - Multi-dip/ambiguity reporting.
    
- Report two independent fit outcomes instead of one: whether a resonator dip was actually found (usable frequency), and whether the fitted lineshape itself was clean.

- Add optional no-dip span-escalation (escalation.py) for fresh-from-fab bring-up, doubling the sweep span and re-centering the readout LO when no resonator is found.

- Add grid-size-aware plotting (plotting.py) so figure/font sizes scale with the qubit grid instead of a fixed 15x9 layout.

- Add optional plot_detrended_phase and plot_iq_circle diagnostics plots.

- Add a manual per-qubit re-fit window overrides (parameters.py).

### DIFF
--- a/qualibration_graphs/superconducting/calibration_utils/resonator_spectroscopy/analysis.py
+++ b/qualibration_graphs/superconducting/calibration_utils/resonator_spectroscopy/analysis.py
@@ -344,12 +344,7 @@ def _select_fit_outputs(
     centre_inside_fit_window: bool = fit_win_lo <= f0_out <= fit_win_hi
 
     success_shape: bool = bool(
-        fit_trusted
-        and good_r2
-        and fwhm_in_range
-        and fwhm_not_too_broad
-        and contrast_ok
-        and centre_inside_fit_window
+        fit_trusted and good_r2 and fwhm_in_range and fwhm_not_too_broad and contrast_ok and centre_inside_fit_window
     )
 
     return SelectedFitOutputs(
--- a/qualibration_graphs/superconducting/calibration_utils/resonator_spectroscopy/analysis.py
+++ b/qualibration_graphs/superconducting/calibration_utils/resonator_spectroscopy/analysis.py
@@ -1,105 +1,622 @@
 import logging
-from dataclasses import dataclass
-from typing import Tuple, Dict
+import warnings
+from dataclasses import dataclass, field
+from typing import Any, Callable
+
 import numpy as np
 import xarray as xr
+from numpy.typing import NDArray
+from scipy.ndimage import median_filter
+from scipy.optimize import curve_fit
+from scipy.signal import find_peaks, peak_widths, savgol_filter
 
 from qualibrate import QualibrationNode
 from qualibration_libs.data import add_amplitude_and_phase, convert_IQ_to_V
-from qualibration_libs.analysis import peaks_dips
+
+
+def lorentzian_dip_linbg(
+    f: NDArray[np.float64],
+    f0: float,
+    fwhm: float,
+    amp: float,
+    bg0: float,
+    bg1: float
+) -> NDArray[np.float64]:
+    """Inverted Lorentzian with linear background.
+
+    R(f) = [bg0 + bg1*(f - fc)] - amp / [1 + ((f - f0)/(fwhm/2))^2]
+
+    fc is the centre of the frequency array so that bg1 has units of V/Hz
+    without absorbing a large offset into bg0.
+    """
+    fc = f.mean()
+    return (bg0 + bg1 * (f - fc)) - amp / (1 + ((f - f0) / (fwhm / 2)) ** 2)
+
+
+def lorentzian_dip_quadbg(
+    f: NDArray[np.float64],
+    f0: float,
+    fwhm: float, amp: float,
+    bg0: float,
+    bg1: float,
+    bg2: float
+) -> NDArray[np.float64]:
+    """Inverted Lorentzian with quadratic background (curved / bowl baselines)."""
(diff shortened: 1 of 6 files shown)


qualibration_graphs/superconducting/calibrations/1Q_calibrations/02a_resonator_spectroscopy.py
```diff
@@ -156,53 +157,166 @@ def execute_qua_program(node: QualibrationNode[Parameters, Quam]):
...
 
+def _analyse(node: QualibrationNode[Parameters, Quam]) -> None:
+    """Process + fit + log + outcomes (shared between the main pass and the retries)."""
+    node.results["ds_raw"] = process_raw_dataset(node.results["ds_raw"], node)
+    node.results["ds_fit"], fit_results = fit_raw_data(node.results["ds_raw"], node)
+    node.results["fit_results"] = {k: asdict(v) for k, v in fit_results.items()}
+
+    # Log the relevant information extracted from the data analysis
+    log_fitted_results(node.results["fit_results"], log_callable=node.log)
+    node.outcomes = {
```
Deepakkhurrana, line 176: Ambiguous fits should not update state

When ambiguous=True (multiple comparable dips), we still set outcomes=successful and write RF_frequency from the top candidate. That can lock in a feedline neighbour on wide bring-up scans.

I would suggest treating ambiguous as failed for outcomes / state update (keep success for “dip found” diagnostics if useful), so the machine is only updated once the resonator is disambiguated.     

```
 qubit_name: (
            "successful"
            if fit_result["success"] and not fit_result.get("ambiguous", False)
            else "failed"
        )
```

qualibration_graphs/superconducting/calibration_utils/resonator_spectroscopy/analysis.py
```diff
@@ -1,105 +1,623 @@
...
+
+    A fit is only trusted if its centre landed close to the detected dip
+    (``best["f0"]`` within ``2*fwhm_init`` or 1 MHz of ``f0_init``) — otherwise
+    it ran away and the frequency/FWHM fall back to the pre-fit estimates.
+    ``success_shape`` gates on R², FWHM range, and contrast; frequency success
+    never depends on it, since the dip's existence and position were already
+    established before any fit was attempted.
+    """
+    nan5 = np.full(5, np.nan)
+    if best is not None and abs(best["f0"] - f0_init) <= max(2 * fwhm_init, 1e6):
```
Deepakkhurrana: `success_shape` on runaway fits

If the Lorentzian centre drifts away from the detected dip, we correctly fall back to the dip frequency and clear popt to NaN — but `success_shape` can still pass using that discarded fit’s R². Please require the trusted-fit branch (fit_trusted) for `success_shape`, so we never report a clean lineshape with no stored popt. Something like this


```
nan5 = np.full(5, np.nan)
fit_trusted = best is not None and abs(best["f0"] - f0_init) <= max(2 * fwhm_init, 1e6)
if fit_trusted:
    f0_out, fwhm_out = best["f0"], best["fwhm"]
    r2 = best["r2"]
    fitted_contrast = best["contrast"]
    popt = best["popt"]
    fit_win_lo, fit_win_hi = best["f_win"]
else:
    # Fit unusable/ran away: keep dip frequency; do not grade success_shape
    # from a discarded Lorentzian (avoids success_shape=True with NaN popt).
    f0_out, fwhm_out = f0_init, fwhm_init
    r2 = best["r2"] if best is not None else 0.0
    fitted_contrast = depth_d / np.median(smoothed) if np.median(smoothed) > 0 else np.nan
    popt = nan5.copy()
    fit_win_lo, fit_win_hi = np.nan, np.nan
success_shape = bool(
    fit_trusted
    and r2 >= r2_threshold
    and 0 < fwhm_out <= max_fwhm_mhz * 1e6
    and fwhm_out / span_hz <= 0.50
    and fitted_contrast >= min_contrast
    and fit_win_lo <= f0_out <= fit_win_hi
)
```

Deepakkhurrana (changes requested):
Couple of comments, good to go after they are handled

qualibration_graphs/superconducting/calibrations/1Q_calibrations/02a_resonator_spectroscopy.py
ohad-lev-qm: Done

qualibration_graphs/superconducting/calibration_utils/resonator_spectroscopy/analysis.py
ohad-lev-qm: Fixed that. Also refactored this function for it to be clear when a human reads it (at least when I read it).

Also - refactored all returned dummy `dict[str, Any]` objects and transformed them to documented dataclasses (much better practice & improves readability)

Deepakkhurrana:
Made a little change to group parameters to not make too much noise in the UI

Screenshot 2026-09-08 at 10 16 10

Deepakkhurrana (approved):
LGTM

ohad-lev-qm:
> Made a little change to group parameters to not make too much noise in the UI
> 
> Screenshot 2026-09-08 at 10 16 10

Cool!

PR #570, opened 2026-08-21
Fix state buffer bug and add fit-quality gates to power Rabi calibration

ohad-lev-qm:
* Fixed a buffer-size bug in `04b_power_rabi.py`'s error-amplification path: `state_st` was buffered by `ceil(max_number_pulses_per_sweep / 2)` instead of `len(N_pi_vec)`.

* Gave the EF node (`13_power_rabi_ef`) its own explicit operation parameter, so the shared analysis/plotting code keys off `node.parameters.operation instead` of branching on `node.name`.

* Simplified how the pi-pulse amplitude is derived from the fit: it's now just half the oscillation period.

* Added oscillation fit-quality gates (SNR, periods spanned, points per period, and a fit-free raw-peak cross-check) that a fit must now also pass to be marked successful, on top of the existing amplitude-range check.

* Removed a readout_frequency 2D best-contrast fitting/plotting path that was dead code - neither `04b_power_rabi.py` nor `13_power_rabi_ef.py` ever sweeps that dimension.

* Added full modern type hints (`dict` / `tuple` / `X | None` / `NDArray`) and a one-line docstring under every `FitParameters` field.

### DIFF
--- a/qualibration_graphs/superconducting/calibration_utils/power_rabi/analysis.py
+++ b/qualibration_graphs/superconducting/calibration_utils/power_rabi/analysis.py
@@ -228,7 +228,9 @@ def fit_raw_data(ds: xr.Dataset, node: QualibrationNode) -> tuple[xr.Dataset, di
             _raw_pi_prefactor(fit_var.sel(qubit=q), float(fit_vals.sel(qubit=q, fit_vals="f").values))
             for q in fit_var.qubit.values
         ]
-        ds_fit = xr.merge([ds, fit_vals.rename("fit")])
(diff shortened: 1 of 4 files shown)


Deepakkhurrana:
Minor Changes pushed: 
-analysis.py — xr.merge([ds, ...]) → xr.merge([ds_fit, ...]) so the length-1 nb_of_pulses axis is not put back; SNR quality gating actually runs on default 04b.
- plotting.py — if fit.success → if fit is not None and bool(fit.success) so 2D overlay doesn’t crash if fit is missing.

LGTM

Deepakkhurrana (approved):

PR #571, opened 2026-08-27
fix old qualibrate imports and incorrect bool type

galimbarQM:
a. updated the imports on the four calibration graphs to ones supported on qualibrate 1.x,
b. updated the analysis.py of two specific nodes- cast np.bool to a general bool (np.bool is not supported in JSON and it produced a bug).

both were tested on IQCC usng my local environment.

### DIFF
--- a/qualibration_graphs/superconducting/calibration_utils/power_rabi/analysis.py
+++ b/qualibration_graphs/superconducting/calibration_utils/power_rabi/analysis.py
@@ -122,7 +122,7 @@ def _rabi_fit_quality(fit: xr.Dataset, use_state_disc: bool) -> dict[str, dict[s
 
 def _gate_ok(value: float, threshold: float, minimum: bool) -> bool:
     """True if `value` clears `threshold`; a NaN value (metric not computed) is never gated."""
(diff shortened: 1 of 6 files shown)


Deepakkhurrana:
Looks good, approve from my side.

### Imports
The four calibration graphs now use the qualibrate 1.x paths (`qualibrate.core.orchestration…` and top-level `QualibrationGraph` / `QualibrationLibrary`). Matches what the rest of the nodes already do.

### `bool()` casts
Sensible fix for the JSON serialization issue. `np.isnan` / `np.isfinite` comparisons yield `np.bool_`, which isn’t JSON-serializable. Casting in `_gate_ok` / `raw_fit_consistent` (power rabi) and the ramsey success criterion is the right place. These nodes were updated in recent PRs and the bugs sneaked in somehow. 

### Nits (non-blocking)
- `99_CZ_calibration_graph.py` still has the old import paths, worth fixing here
- After `_gate_ok` returns `bool(...)`, `raw_fit_consistent=bool(_raw_gap_ok(...))` is slightly redundant (harmless).

Deepakkhurrana (approved):

PR #573, opened 2026-09-02
Upgrade long time qubit flux distortions calibrations

Deepakkhurrana:
## Summary

Replace the legacy `17_pi_vs_flux_long_distortions` / `pi_flux` long-cryoscope path with two qubit flux long-distortion calibrations, backed by shared helpers:

- **17a** (`qubit_flux_long_distortion_qubitspec`) — π / spectroscopy probe: extract `f(t)`, map to flux via spectroscopy → Ramsey-vs-Z → `quad_term`, multi-exp fit → IIR taps.
- **17b** (`qubit_flux_long_distortion_ramsey`) — Ramsey probe with co-measured reference amp sweep for phase→flux inversion, then the same multi-exp / IIR path.
- **Shared** `calibration_utils/common_utils/flux_distortions/` — multi-exp fitting (`FitParameters`), freq↔flux curve loaders / amp resolution, LO-shift planning for detuned sweeps.

Retires `calibration_utils/pi_flux/` and renames the old node to `17a_…`. Coupler / short-distortion nodes are **not** included here; this shared groundwork is intended to support upcoming coupler flux-distortion calibrations built on the same helpers.

## What’s new vs old 17

| | Old `17` / `pi_flux` | 17a | 17b |
|---|---|---|---|
| Probe | π + freq sweep | same (evolved) | Ramsey + frame rotation |
| Flux amp | mostly `quad_term` | spec → Ramsey → quad cascade + `flux_branch` | manual + in-run ref amp sweep |
| Fit | sequential peel | shared `multi_exp_fit_global` | same (+ finite-pulse `t_pulse_ns`) |
| Plots | basic | default + optional `debug_plots` | default + optional `debug_plots` / branch-risk watermark |

## Test plan

- [ ] Hardware validation on IQCC once the chips are cold


Deepakkhurrana:
@paulQM, one issue but not sure how to handle it  

### `exponential_filter.extend` stacks taps on repeat runs

In both **17a** and **17b**, `update_state` does:

```python
z.opx_output.exponential_filter.extend(list(zip(A_list, tau_list)))
```
Each successful run with `update_state=True` appends a new cascade instead of replacing the previous one. After two calibrations the Z line carries old taps + new taps, which compounds predistortion rather than updating it.

Suggest assigning (or clear-then-set) instead, e.g.:

```python
z.opx_output.exponential_filter = list(zip(A_list, tau_list))
```

Happy to fix 17a/17b together in a follow-up so they stay consistent. Worth confirming whether replace is always correct, or whether we want an explicit “append cascade stage” mode for intentional multi-pass fitting.

qualibration_graphs/superconducting/calibration_utils/common_utils/flux_distortions/__init__.py
ohad-lev-qm, line 1: @Deepakkhurrana So we start dropping shared utils into this `common_utils` directory? Sounds good, I suggest to separate it from the other util (per-node) directories somehow.. Maybe call it `aa_calibration_utils` or something similar

qualibration_graphs/superconducting/calibration_utils/common_utils/flux_distortions/curves.py
```diff
@@ -0,0 +1,645 @@
...
+#: Priority order used by ``source="auto"``: measured curves first (Ramsey is the
+#: most accurate freq↔flux map), quadratic term only as a last resort.
+AUTO_SOURCE_ORDER: Tuple[str, ...] = ("ramsey", "spectroscopy")
+
+#: ``qubit.extras`` keys written by the source calibrations.
+RAMSEY_EXTRAS_KEY = "ramsey_vs_flux_calibration_load_id"  # 09a, when save_load_id=True
+SPECTROSCOPY_EXTRAS_KEY = "qubit_spectroscopy_vs_flux_load_id"  # 03b, when save_load_id=True
+
+
+def extras_run_id(qubit, key: str) -> Optional[int]:
```
ohad-lev-qm: What is this function? I didn't understand

qualibration_graphs/superconducting/calibration_utils/common_utils/flux_distortions/curves.py
```diff
@@ -0,0 +1,645 @@
...
+        print(f"  WARNING: {getattr(qubit, 'name', '?')}: extras['{key}']={value!r} is not a run ID")
+        return None
+
+
+# ---------------------------------------------------------------------------
+# Loaders
+# ---------------------------------------------------------------------------
+
+
+def _read_node_data_dict(run_id: int) -> dict:
```
ohad-lev-qm: This function should be somewhere else I think. It is not related to "curves"

qualibration_graphs/superconducting/calibration_utils/common_utils/flux_distortions/curves.py
```diff
@@ -0,0 +1,645 @@
...
+    from qualibrate.core.utils.node.content import read_node_data
+    from qualibrate.core.utils.node.path_solver import get_node_dir_path
+    from qualibrate_config.resolvers import get_qualibrate_config, get_qualibrate_config_path
+
+    base_path = get_qualibrate_config(get_qualibrate_config_path()).storage.location
+    node_dir = get_node_dir_path(run_id, base_path)
+    return read_node_data(node_dir, run_id, base_path)
+
+
+def load_spectroscopy_curve(qubit, run_id: Optional[int] = None) -> Optional[Tuple[np.ndarray, np.ndarray]]:
```
ohad-lev-qm: I strongly recommend to add proper & modern type hinting everywhere. Should be an easy task for an agent, can do that if you want. For example, here `qubit` should be... `str`? some object?

qualibration_graphs/superconducting/calibration_utils/common_utils/flux_distortions/curves.py
```diff
@@ -0,0 +1,645 @@
...
+        order = np.argsort(flux_m)
+        flux_m, freq_m = flux_m[order], freq_m[order]
+        print(
+            f"  Loaded spectroscopy curve for {qubit_name} from run #{rid} "
+            f"(ds_fit.peak_freq): {len(flux_m)} pts, "
+            f"flux=[{flux_m[0]:.4f}, {flux_m[-1]:.4f}] V"
+        )
+        return flux_m, freq_m
+    except Exception as e:
+        print(f"  WARNING: Failed to load spectroscopy curve for {qubit_name} from run #{rid}: {e}")
```
ohad-lev-qm: Switching this print to a proper logging is better

ohad-lev-qm: And everywhere else..

qualibration_graphs/superconducting/calibration_utils/common_utils/flux_distortions/curves.py
```diff
@@ -0,0 +1,645 @@
...
+    run_id: Optional[int] = None
+    quad_term: Optional[float] = None
+
+    @property
+    def is_measured(self) -> bool:
+        """True when a measured freq-vs-flux ``curve`` is available."""
+        return self.curve is not None
+
+
+def _load_source_curve(qubit, kind: str) -> Optional[Tuple[np.ndarray, np.ndarray]]:
```
ohad-lev-qm: This function looks redundant. I would drop it

qualibration_graphs/superconducting/calibration_utils/common_utils/flux_distortions/fitting.py
```diff
@@ -0,0 +1,281 @@
...
+TAU_MAX_FACTOR = 100.0
+TAU_PULSE_FACTOR = 20.0  # keep (1 - exp(-T/tau)) above ~5% so deconvolution stays stable
+ADC_PADDING_FRAC = 0.1
+
+
+@dataclass
+class FitParameters:
+    """Long-distortion multi-exp fit parameters for one channel."""
+
+    success: bool
```
ohad-lev-qm, line 83: I would add here one-liner docstrings to explain each param

qualibration_graphs/superconducting/calibrations/1Q_calibrations/17a_qubit_flux_long_distortion_qubitspec.py
```diff
@@ -220,8 +235,17 @@ def create_qua_program(node: QualibrationNode[Parameters, Quam]):
 # %% {Simulate_qua_program}
 @node.run_action(skip_if=node.parameters.load_data_id is not None or not node.parameters.simulate)
 def simulate_qua_program(node: QualibrationNode[Parameters, Quam]):
+    """Connect to the QOP and simulate the QUA program."""
     qmm = node.machine.connect()
     config = node.machine.generate_config()
+    debug = False
```
ohad-lev-qm: This probably should be removed

ohad-lev-qm (commented):
Hi,

I am not familiar with these Cryoscope variants, would love to try them experimentally. 

Anyway, I left a few comments. I would suggest to clean up a little the code, see the comments. Would love to help if needed.

Thanks!

qualibration_graphs/superconducting/calibration_utils/common_utils/flux_distortions/__init__.py
Deepakkhurrana: Good idea for longer term. For this PR we’re keeping calibration_utils/common_utils/flux_distortions/ as-is so 17a/17b stay focused and we don’t churn imports across the repo. The intent is: per-node packages (qubit_flux_long_distortion_*) vs cross-node shared helpers under common_utils/flux_distortions/. Happy to revisit naming in a follow-up if we add more shared modules.

qualibration_graphs/superconducting/calibration_utils/common_utils/flux_distortions/curves.py
Deepakkhurrana: Added a fuller docstring. In short: prior calibrations (09a Ramsey vs flux, 03b spec vs flux) can write a Qualibrate snapshot index into qubit.extras when save_load_id=True. This helper reads that index so 17a never needs a hand-typed run ID — it loads the saved curve from disk via read_node_data_dict(run_id).

qualibration_graphs/superconducting/calibration_utils/common_utils/flux_distortions/curves.py
Deepakkhurrana: Agreed — moved to common_utils/flux_distortions/node_storage.py as public read_node_data_dict(run_id). Curve loaders import from there; nothing storage-related left in curves.py except the call site.

qualibration_graphs/superconducting/calibration_utils/common_utils/flux_distortions/curves.py
Deepakkhurrana: Added modern hints across flux_distortions/: qubit: AnyTransmon (from quam_builder), MeasuredCurve, NDArray, optional run IDs, return types on loaders/resolvers. Same pattern on resolve_freq_flux_curve, resolve_flux_amplitudes, etc.

qualibration_graphs/superconducting/calibration_utils/common_utils/flux_distortions/curves.py
Deepakkhurrana: Replaced print with an optional log_callable threaded from the node. In 17a/17b we pass node.log, so curve-load / warning messages show in the Qualibrate UI log panel (not stdout). When helpers are used without a node, they stay silent unless log_callable is provided.

Deepakkhurrana: Same treatment in curves.py, fitting.py (multi_exp_fit_global verbose path), and lo_shift.py (LO update message). All wired to node.log from 17a analysis / create_qua_program and 17b analysis where relevant.

qualibration_graphs/superconducting/calibration_utils/common_utils/flux_distortions/curves.py
Deepakkhurrana: Done — removed the helper and inlined Ramsey/spectroscopy dispatch directly in resolve_freq_flux_curve.

qualibration_graphs/superconducting/calibration_utils/common_utils/flux_distortions/fitting.py
Deepakkhurrana: Added an Attributes section on FitParameters documenting success, n_components_requested/used, a_tau_tuple, a_dc, and rms_error (including that IIR taps are a_i / a_dc).

qualibration_graphs/superconducting/calibrations/1Q_calibrations/17a_qubit_flux_long_distortion_qubitspec.py
Deepakkhurrana: Removed — the debug = False / generate_qua_script block was dead code (never enabled in the repo). simulate_qua_program now goes straight to simulate_and_plot. If we need to dump QUA scripts locally again, we can add it behind a proper node parameter or keep it as a one-off in custom_param.

Deepakkhurrana:
17a results: 

image
image

17b results: 

image
image
image
image

Deepakkhurrana:
@ohad-lev-qm @paulQM All done from my side, please approve  if you are happy as well

paulQM (approved):

PR #580, opened 2026-09-11
Three tone spectroscopy to find coupler frequency

Deepakkhurrana:
## Summary
- Add **22a** (`22a_three_tone_coupler_spectroscopy_flux_pulse`): drive the coupler through the control qubit while probing the target, at a fixed coupler flux pulse. Fit the three-tone dip and optionally write `coupler.RF_frequency`.
- Add **22b** (`22b_three_tone_coupler_spectroscopy_vs_coupler_flux`): same drive, sweeping coupler flux-pulse amplitude vs drive frequency.
- Sweep centre is `rf_frequency_startpoint_in_hz`, else `coupler.RF_frequency` from state, else an idle guess: **1.2 GHz above `max(f)` or below `min(f)`** (`coupler_band` / `coupler_idle_detuning_in_ghz`).
- Extend existing `plan_lo_shift_for_frequency_window` with optional `if_base_hz` so the coupler-drive IF (not the qubit XY IF) stays in MW-FEM reach. 17a/21a are unchanged if they omit the argument.

Independent of the 21a/21b/21c flux-distortion branch.

## Test plan
- Nodes were run end to end on a tunable-coupler chip (program, fetch, analysis, plots).
- Physics is still untested: we did not check that fitted coupler frequencies or flux maps are fully accurate.

### DIFF
--- a/qualibration_graphs/superconducting/calibration_utils/common_utils/flux_distortions/lo_shift.py
+++ b/qualibration_graphs/superconducting/calibration_utils/common_utils/flux_distortions/lo_shift.py
@@ -67,16 +67,21 @@ def plan_lo_shift_for_frequency_window(
     qubits: Sequence[AnyTransmon],
     dfs: NDArray[np.integer] | NDArray[np.floating],
     *,
+    if_base_hz: Optional[Sequence[float]] = None,
     log_callable: Optional[LogCallable] = None,
 ) -> LoShiftPlan:
-    """Decide LO shifts so ``intermediate_frequency + dfs`` stays in usable IF reach.
+    """Decide LO shifts so ``if_base + dfs`` stays in usable IF reach.
 
     Parameters
     ----------
     qubits
-        Qubit-like objects with ``xy`` (and ``name``).
+        Qubit-like objects with ``xy`` (and ``name``). LO shifts apply to these channels.
     dfs
         Relative frequency sweep axis in Hz (same array used in QUA ``from_array``).
+    if_base_hz
+        Optional per-qubit IF centre for the sweep (same length as ``qubits``).
+        Defaults to each qubit's ``xy.intermediate_frequency``. Use this for three-tone
+        spectroscopy where the drive IF is ``coupler_RF - LO`` rather than the qubit idle IF.
 
     Returns
     -------
@@ -89,8 +94,8 @@ def plan_lo_shift_for_frequency_window(
     dfs_mid = int((dfs.min() + dfs.max()) / 2)
     plan = LoShiftPlan()
 
-    for q in qubits:
-        if_lo = q.xy.intermediate_frequency
+    for i, q in enumerate(qubits):
+        if_lo = int(if_base_hz[i]) if if_base_hz is not None else q.xy.intermediate_frequency
         if_low = if_lo + int(dfs.min())
         if_high = if_lo + int(dfs.max())
 
--- a/qualibration_graphs/superconducting/calibration_utils/three_tone_coupler_spectroscopy_flux_pulse/analysis.py
+++ b/qualibration_graphs/superconducting/calibration_utils/three_tone_coupler_spectroscopy_flux_pulse/analysis.py
@@ -1,0 +1,88 @@
+"""Analysis for three-tone coupler spectroscopy with flux pulse (22a)."""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Callable, Dict, Optional
+
+import numpy as np
+import xarray as xr
+from qualibrate import QualibrationNode
+
+from .parameters import resolve_coupler_rf_centers_by_pair
+
+LogCallable = Callable[[str], None]
+
+
+@dataclass
+class FitResults:
+    """Per-pair coupler frequency extracted from the three-tone dip."""
+
+    success: bool
+    coupler_frequency_hz: float
+    coupler_flux_v: float
+
+
+def process_raw_dataset(ds: xr.Dataset, node: QualibrationNode) -> xr.Dataset:
+    """Add derived coordinates and IQ amplitude when needed."""
+    qubit_pairs = node.namespace["qubit_pairs"]
+    dfs = node.namespace["dfs"]
+    coupler_rf_centers = node.namespace.get("coupler_rf_centers") or resolve_coupler_rf_centers_by_pair(
+        qubit_pairs,
+        node.parameters.rf_frequency_startpoint_in_hz,
+        coupler_band=node.parameters.coupler_band,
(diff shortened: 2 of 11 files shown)


qualibration_graphs/superconducting/calibration_utils/common_utils/flux_distortions/lo_shift.py
```diff
@@ -89,8 +94,8 @@ def plan_lo_shift_for_frequency_window(
     dfs_mid = int((dfs.min() + dfs.max()) / 2)
     plan = LoShiftPlan()
 
-    for q in qubits:
-        if_lo = q.xy.intermediate_frequency
+    for i, q in enumerate(qubits):
+        if_lo = int(if_base_hz[i]) if if_base_hz is not None else q.xy.intermediate_frequency
```
paulQM, line 98: `if_base_hz` changes the IF centre used for the range check, but the shift amount further down is still `dfs_mid` only, which is ~-0.5 MHz for a sweep symmetric around zero. So for a coupler drive the function detects that the window is out of reach, forces thermal reset, moves the LO by half a megahertz, and leaves the IF where it was.

Concretely, with control LO 5.1 GHz and a coupler centre of 6.4 GHz, `coupler_ifs` is 1.3 GHz. The plan shifts the LO to 5.0995 GHz and sets `if_update = -0.5e6`, so `update_frequency(df + coupler_if - if_update)` in 22a/22b still asks for 0.9-1.7 GHz of IF, well past the +/-500 MHz `IF_MAX_HZ` the MW-FEM can reach.

The shift needs to absorb the base offset too, i.e. `shift = if_lo + dfs_mid` (rather than `dfs_mid`) for both `plan.if_update` and the new `lo_frequency`, so the LO lands on the centre of the coupler window and the residual IF is back near zero. The `band_floor` comparison should use that same LO. As written, the nodes only work when the coupler already sits within ~500 MHz of the control LO, which is the case the new argument was added to avoid.

qualibration_graphs/superconducting/calibrations/1Q_calibrations/22a_three_tone_coupler_spectroscopy_flux_pulse.py
```diff
@@ -0,0 +1,329 @@
...
+                    else target.xy.operations[node.parameters.target_drive_operation].length * u.ns
+                ) // 4
+
+            with for_(n, 0, n < node.parameters.num_shots, n + 1):
+                save(n, n_st)
+                with for_(*from_array(df, dfs)):
+                    for ii, qp in multiplexed_qubit_pairs.items():
+                        control = qp.qubit_control
+                        target = qp.qubit_target
+                        control.reset(node.parameters.reset_type, node.parameters.simulate)
```
paulQM, line 154: `control.reset(...)` runs before `control.xy.update_frequency(...)` on the next line, so the reset uses whatever frequency the element was left at by the previous sweep point. For the first point that is the idle IF, but from the second point onward the active-reset x180 plays at the previous point's coupler-drive frequency and does not address the qubit.

The default `reset_type` is `"thermal"`, so this is latent rather than always-on, but it bites silently whenever someone selects `"active"` (and the LO planner only forces thermal when it actually shifts an LO). Moving the `update_frequency` call above the two `reset` calls, or resetting to the idle IF first, would fix it.

Same ordering in 22b at line 163.

qualibration_graphs/superconducting/calibration_utils/three_tone_coupler_spectroscopy_flux_pulse/analysis.py
```diff
@@ -0,0 +1,88 @@
...
+    success: bool
+    coupler_frequency_hz: float
+    coupler_flux_v: float
+
+
+def process_raw_dataset(ds: xr.Dataset, node: QualibrationNode) -> xr.Dataset:
+    """Add derived coordinates and IQ amplitude when needed."""
+    qubit_pairs = node.namespace["qubit_pairs"]
+    dfs = node.namespace["dfs"]
+    coupler_rf_centers = node.namespace.get("coupler_rf_centers") or resolve_coupler_rf_centers_by_pair(
```
paulQM, line 30: `coupler_rf_centers` only lives in `node.namespace`, which is not persisted, so on a `load_data_id` run this falls through to `resolve_coupler_rf_centers_by_pair` against the loaded state instead of the centres the sweep actually used.

That recomputation is not guaranteed to return the same numbers. After a 22a run with `update_state=True`, `coupler.RF_frequency` in the saved state holds the fitted value, so reloading centres the axis on the fit result rather than on the original sweep centre and `freq_full_control` comes out shifted by the fit offset. Anything else that touches `coupler.RF_frequency` between runs has the same effect.

Saving the centres with the data would make reloads exact, e.g. attach `freq_full_control` (or the per-pair centre) as a coordinate on `ds_raw` in `create_qua_program`/`execute_qua_program` and read it back here when present. Same pattern in the 22b analysis at line 34.

paulQM (commented):
Read through the two nodes and the `lo_shift` extension. Three correctness points below, inline. Physics/analysis nits aside, the LO one looks like it blocks the intended use case.

Deepakkhurrana:
Thanks for the comments @paulQM! 

**LO shift:** agreed, `if_base` was only used for the range check. The shift is now `if_lo + dfs_mid` for the new LO, `if_update`, band floor, and `RF_frequency`. 22a/22b still do `df + coupler_if - if_update` with `coupler_if` taken from the pre-shift LO, so no node change there.

**Reset vs `update_frequency`:** restore idle IF, then reset, then set the coupler-drive IF (same pattern as 04a). 

**Reload axis:** `freq_full_control` is written on `ds_raw` in execute. Analysis uses that coord and no longer rebuilds centres from state.

qualibration_graphs/superconducting/calibration_utils/common_utils/flux_distortions/lo_shift.py
Deepakkhurrana: Done!

qualibration_graphs/superconducting/calibrations/1Q_calibrations/22a_three_tone_coupler_spectroscopy_flux_pulse.py
Deepakkhurrana: Corrected!

qualibration_graphs/superconducting/calibration_utils/three_tone_coupler_spectroscopy_flux_pulse/analysis.py
Deepakkhurrana: Corrected as well!

paulQM (approved):

PR #581, opened 2026-09-11
Coupler flux distortions calibrations

Deepakkhurrana:
## Summary
- Add shared coupler flux helpers: `coupler_curves.py`, coupler filter updates, and rename `curves.py` → `qubit_curves.py` (qubit 17a/17b/17c imports follow).
- Add **21a** coupler long-distortion qubitspec, **21b** coupler long-distortion Ramsey, and **21c** coupler short distortion, aligned with 17a/17b/17c.
- Point 03c/09b coupler-curve extras keys at the shared loaders.
- Clarify Ramsey branch-alias warnings (widen `ramsey_flux_sweep_range_in_v` or lower amplitude) and shorten the 21b settle-time docstring.

## Test plan
- Nodes were run end to end on a tunable-coupler chip (program, fetch, analysis, plots).
- Physics is still untested: we did not check that fitted distortion filters or flux-response curves are fully accurate.


qualibration_graphs/superconducting/calibration_utils/common_utils/flux_distortions/coupler_curves.py
```diff
@@ -0,0 +1,418 @@
...
+        if key is None:
+            if log_callable is not None:
+                log_callable(
+                    f"{qubit_name}/{coupler.name}: could not resolve coord in run #{rid}; "
+                    f"skipping spectroscopy path"
+                )
+            return None
+
+        ds_fit = read_node_data_dict(rid).get("ds_fit")
+        if ds_fit is None or "peak_freq" not in getattr(ds_fit, "data_vars", {}):
```
paulQM, line 119: 03c's analysis never writes a `peak_freq` variable. `qubit_spectroscopy_vs_coupler/analysis.py` only stores `peak_frequency` (relative to RF) and `abs_peak_frequency`. So this check is always true, the loader always returns `None`, and the spectroscopy source never gets used. `auto` silently falls back to 09b, and `freq_to_flux_source="spectroscopy"` always misses.

Since the next lines add `qubit_rf`, I think you want `peak_frequency`:
```python
if ds_fit is None or "peak_frequency" not in getattr(ds_fit, "data_vars", {}):
    ...
peak = np.asarray(ds_fit.peak_frequency.sel(qubit=key).values, dtype=float)
```
Another option is to read `abs_peak_frequency` directly and drop the `qubit_rf +` offset.

qualibration_graphs/superconducting/calibration_utils/common_utils/flux_distortions/coupler_curves.py
```diff
@@ -0,0 +1,418 @@
...
+                    if pair.coupler.name == coupler.name and str(pair_name) in qp_names:
+                        pair_key = str(pair_name)
+                        break
+                if pair_key is None:
+                    if log_callable is not None:
+                        log_callable(f"No qubit_pair match for {qubit.name} / {coupler.name} in run #{rid}")
+                    return None
+                frequency = ds_fit["qubit_frequency"].sel(qubit_pair=pair_key).values
+            else:
+                frequency = ds_fit["qubit_frequency"].values
```
paulQM, line 183: 09b renames `qubit_pair` to `qubit` before analysis (`09b_ramsey_vs_coupler_flux.py` L184–186), so `ds_fit["qubit_frequency"]` has dims `(qubit, coupler_flux)` and `"qubit_pair" in ds_fit.dims` is never true. We always land in this `else`, which returns the full 2-D `(n_pairs, n_flux)` array as the frequency. `frequency_to_coupler_flux` then indexes rows with flux-length indices (IndexError or wrong data), and `coupler_flux_amp_from_curve` works on the wrong shape.

Suggest selecting the pair on the `qubit` coord, e.g. by reusing `_resolve_pair_coord_key`:
```python
key = _resolve_pair_coord_key(qubit, coupler, node, ds_fit.qubit.values)
if key is None:
    ...  # log + return None
frequency = ds_fit["qubit_frequency"].sel(qubit=key).values
```
The same applies to the fallback path below (L186–196), which does `.sel(qubit_pair=...)` and takes the first pair with a matching coupler, whichever qubit was measured.

qualibration_graphs/superconducting/calibrations/1Q_calibrations/21a_coupler_flux_long_distortion_qubitspec.py
```diff
@@ -0,0 +1,407 @@
...
+    node.namespace["coupler_flux_amps"] = coupler_flux_amps
+    node.results["coupler_flux_center"] = coupler_flux_amps[0]
+    node.results["decouple_offsets"] = [qp.coupler.decouple_offset for qp in qubit_pairs]
+
+    # Build dfs array: signed, centered at detuning relative to idle.
+    # If freq_at_decouple differs from RF_frequency (idle), dfs must account
+    # for that offset so the XY sweep is centered correctly.
+    ref_qubit = measured_qubits[0]
+    f_dec_ref = resolved.freq_at_decouple[0]
+    idle_to_decouple_offset_hz = f_dec_ref - ref_qubit.xy.RF_frequency if f_dec_ref is not None else 0.0
```
paulQM, line 129: The XY sweep is centred using only the first pair: `f_dec_ref = resolved.freq_at_decouple[0]`, and one `dfs` array is shared by every pair in the program. With several pairs, any pair whose qubit frequency at the decouple point differs from pair 0's gets an off-centre sweep, and its resonance can fall outside `frequency_span_in_mhz`. `coupler_flux_center` (L119/L121) also only records pair 0.

Options: build `dfs` per pair, e.g. a per-pair offset added in `update_frequency(df + offset[ii] + ...)`, or warn/raise when the per-pair offsets differ by more than a fraction of the span.

qualibration_graphs/superconducting/calibration_utils/common_utils/flux_distortions/coupler_curves.py
Deepakkhurrana: Corrected!  The loader now reads that and no longer adds the current RF.

qualibration_graphs/superconducting/calibration_utils/common_utils/flux_distortions/coupler_curves.py
Deepakkhurrana: Agreed! 09b renames qubit_pair → qubit before analysis and sets those labels to pair names, so ds_fit.qubit_frequency is (qubit, coupler_flux). Checking "qubit_pair" in ds_fit.dims was always false, and the else returned the whole 2-D array. Inversion then indexed it as if it were one pair.

The loader now requires qubit_frequency vs coupler_flux, finds the pair whose coupler matches, and .sel(qubit=pair_name).

qualibration_graphs/superconducting/calibrations/1Q_calibrations/21a_coupler_flux_long_distortion_qubitspec.py
Deepakkhurrana: dfs is now centred on the signed detuning from idle. Each pair adds freq_at_decouple − RF in update_frequency (and in analysis when rebuilding absolute frequency). LO planning uses dfs + offset per qubit.

Deepakkhurrana:
@paulQM all the comments has been addressed.

paulQM (approved):

PR #582, opened 2026-09-14
fix imports on 99_CZ graph to match qualibrate 1.x 

galimbarQM:
I fixed the imports on the 2-qubit calib graph (99_CZ) to fit the updated qualibrate 1.x

I ran pytest on the updated project and there is only one unrelated error:
ERROR Quantum-Control-Applications/Optically addressable spin qubits/Electron Spin Resonance/03_signal_test.py - qm.exceptions.QmServerDetectionError: Failed to detect to QuantumMachines server, failed to connect to a cluster. Tried connecting to 127.0.0.1:80.


Deepakkhurrana:
The pytest is a different error due CS station address being wrong and does not concern this PR. LGTM

Deepakkhurrana (approved):

PR #583, opened 2026-09-14
Fix/ld consistency

KU-QM:
Some consistency fixes throughout the nodes.

### **Biggest Architectural Change**
- Removed `QubitQuam` entirely from `my_quam.py`

**The motivation:**
- This is already a LossDiVincenzo node library. In this case, it makes sense that the default Quam is `LossDiVincenzoQuam`. 
- The node imports are easy to get confused about. The original logic was that the QD level nodes import `Quam` and the qubit level nodes import `QubitQuam`. However, this makes it difficult/error-prone for the user to perform HW level re-calibrations AFTER upgrading to a qubit level Quam (e.g. qubit spectroscopy -> sensor retuning). Therefore it makes architectural sense to clamp all the nodes to a single `Quam` type. 
- For users who want to stay with the QD layer, they can simply change the `Quam` subclass to `BaseQuamQD` instead of `LossDiVincenzoQuam`. In reality, in practise, this will also make very little difference, since the `LossDiVincenzoQuam` is also capable of QD level calibrations

**NOTE**: This achitecture currently does not highlight the advantages of the 2-stage structure in the best way (since we only have `LossDiVincenzoQuam`). However, there are some notes for the future: 
- Once other modalities are released, we can split the node library into a QD calibrations section, a LD section, EO section, and so on. This will make it apparent that the QD calibrations are universal. 
- The quam_config directory should also be expanded to ensure that the user understands the full advantages of the 2-stage approach, also by adapting build_quam in quam-builder. 

### **Other Smaller Changes**
- Made simulated data generators for node family 11 and 14
- Added consistency to the ds_processed in nodes 05, 06 and 08
- Removed any timeout args in any node's connect fns
- Removed custom_params


TorbjornRasmussenQM (approved):
Looks good to me

PR #585, opened 2026-09-15
fix 04b_power_rabi 1Q node to have a sub-pixel fit

galimbarQM:
in the existing implementation, 04b_power_rabi node (in error amp mode) finds the optimal amplitude by looking for a pixel with extremum value.  Here I add a sinc fit around this extremum value to find the optimal amplitude with sub-pixel accuracy. in case this fit fails, the result returns to be the old pixel accuracy value. 
the output plot shows the fitted optimal amp, and explicitly states if it's a sinc fit or not. there is a toggle if the user wants to use the legacy method without the sinc fit. 
I tested it on IQCC/Arbel with different combinations of settings and it seems to be working as intended. 

pytest results with one unrelated error:
ERROR Quantum-Control-Applications/Optically addressable spin qubits/Electron Spin Resonance/03_signal_test.py - qm.exceptions.QmServerDetectionError: Failed to detect to QuantumMachines server, failed to connect to a cluster. Tried connecting to 127.0.0.1:80.


Deepakkhurrana (commented):
Looks good overall: averaging over N gives a narrow peak, so a sinc fit for sub-pixel amplitude makes sense, and falling back to min/max if the fit fails is the right idea.

One thing I’m unsure about: right now we keep the sinc result whenever `curve_fit` converges and the centre is not on the window edge. Could a noisy or neighbouring peak still pass that and get written?

Would it make sense to also drop the sinc (and keep the pixel value) unless:
- the fitted centre is close to the min/max point,
- the peak is clearly above the noise,
- the width matches the number of pulses we averaged?

That check would live in `_sinc_refined_prefactor` so a bad sinc doesn’t fail the qubit, we just don’t use it.

On the plot, would you be open to drawing both the sinc and the pixel line, and storing a flag for “sinc used” instead of comparing the two floats?

PR #586, opened 2026-09-16
feat(superconducting): add 08c_readout_quantum_efficiency node

paulQM:
## Summary
- Adds `08c_readout_quantum_efficiency`, which measures the readout chain quantum efficiency eta over a 2D sweep of measurement-pulse frequency and amplitude, following C.C. Bultink et al., arXiv:1711.05336 (Fig. 1b and 1c).
- Adds the matching `calibration_utils/readout_quantum_efficiency` module (parameters, analysis, plotting).
- Ported from CS_installations (`15c_readout_quantum_efficiency`) and renumbered to 08c so it sits with the other readout nodes here; prerequisite node numbers in the description were remapped to this repo's numbering.

## Notes
- `sweep_frequency = False` (the default) collapses the frequency axis to the calibrated readout frequency, which is the fast way to get eta at the current setpoint.
- Optimal integration weights and active photon depletion are not implemented, so the measured eta is a lower bound.
- The node makes no state update: eta characterises the amplification chain, not a setting of it.

## Test plan
- [ ] Run against hardware with `sweep_frequency = False` on one qubit.
- [ ] Run a frequency sweep and check `quantum_efficiency` and `efficiency_map` figures.
- [ ] Re-analyse a saved run via `load_data_id` with a changed `max_amp_for_fit`.


Deepakkhurrana:
Some point before physics review: 

**Numbering:** do not land this as `08c`. That slot is the third readout *optimization* node (`08a` frequency, `08b` power; duration-power is the missing `08c`). Several nodes already list “08a, 08b and 08c” as optional readout optimization. This node does not set frequency, amplitude, duration, or weights.

Rename to **`23b_readout_quantum_efficiency`** with 23a for dephasing. Keep the utils package name. Use this repo’s node numbers in the description (`07` / `08b` for the confusion matrix, not `15` / `16`).

**Also fix:**

1. The analysis still says “re-run `16_iq_blobs` or `15_readout_power_optimization`.” Here that is `07_iq_blobs` / `08b`.
2. Prerequisites claim x90/x180 from `04b` and `06a`. `06a` is Ramsey. Use `04b` (and `10b` if DRAG matters).

qualibration_graphs/superconducting/calibrations/1Q_calibrations/08c_readout_quantum_efficiency.py
```diff
@@ -0,0 +1,423 @@
...
+                    # The averaging loop is outside the phase loop, so every pass samples the
+                    # whole fringe: drift within the cell then scales all phases alike and
+                    # cancels in |rho01(eps)|/|rho01(0)|, instead of distorting the fringe shape.
+                    with for_(n, 0, n < n_avg, n + 1):
+                        with for_(*from_array(phi, phases_in_turns)):
+                            for i, qubit in multiplexed_qubits.items():
+                                reset_frame(qubit.xy.name)
+                                qubit.reset(node.parameters.reset_type, node.parameters.simulate)
+                            align()
+                            for i, qubit in multiplexed_qubits.items():
```
Deepakkhurrana, line 184: on the weak pulse, is the batch meant to be multiplexed?

Right now, every resonator in the batch plays the probe at the same time. Neighbor photons still dephase this qubit, but the SNR half only looks at this qubit’s IQ, so η comes out low for the wrong reason.

Did you intend that, or should the probe (Ramsey play and SNR measure) run one qubit at a time?

qualibration_graphs/superconducting/calibrations/1Q_calibrations/08c_readout_quantum_efficiency.py
```diff
@@ -0,0 +1,423 @@
...
+                                update_frequency(qubit.resonator.name, df + qubit.resonator.intermediate_frequency)
+                                qubit.xy.play("x90")
+                                qubit.align()
+                                # The weak measurement itself: played, not demodulated - the
+                                # information it carries is measured in the second half. It stays
+                                # at the detuning set above, which is the whole point of the
+                                # frequency axis: beta and the SNR must be measured at the SAME
+                                # measurement-pulse frequency for their ratio to mean anything.
+                                qubit.resonator.play("readout", amplitude_scale=a)
+                                qubit.resonator.wait(qubit.resonator.depletion_time // 4)
```
Deepakkhurrana, line 195: After the weak resonator pulse, we wait `depletion_time` and then play the second `x90`. If the cavity is not empty yet, leftover photons still dephase the qubit, but they never show up in the SNR (that half only integrates during the pulse). η would then be low because of the wait, I think

Is `depletion_time` enough here, or should this follow the paper and actively empty the resonator before the second `x90`? Even a note that this is another lower-bound on top of the weights would help.

qualibration_graphs/superconducting/calibrations/1Q_calibrations/08c_readout_quantum_efficiency.py
```diff
@@ -0,0 +1,423 @@
...
+The measurement-induced dephasing is exp(-beta) = |rho01(eps)| / |rho01(0)|, and the efficiency
+is eta = SNR^2 / (4 beta). Since eta is amplitude independent in the linear regime, the reported
+number per frequency comes from the paper's global fits over the amplitude axis,
+SNR = a*eps and beta = eps^2/(2 sigma_m^2), giving eta = a^2 sigma_m^2 / 2. The point-by-point
+ratio is also produced, as a check that the sweep stayed in the linear regime.
+
+Optimal integration weights and active photon depletion, the other two steps of the paper's
+method, are NOT implemented here: this node uses the existing readout weights and a passive
+depletion wait. eta measured this way is therefore a lower bound on what the chain can deliver.
+
```
Deepakkhurrana, line 68: can we make the description explicit about what this η is?

It is the information in the integrated IQ over the information the qubit lost, for this pulse, these box weights, and a passive depletion wait. It is not readout fidelity, not a QUAM setting, and not the TWPA’s efficiency by itself.

A low number is not enough to blame the amp: missing optimal weights, leftover photons, and a multiplexed probe all pull η down for other reasons. A sentence up front on that would help.

qualibration_graphs/superconducting/calibrations/1Q_calibrations/08c_readout_quantum_efficiency.py
paulQM: I put the depletion time here because all of the readout macros that we use contain the depletion time, but since we are using play here, the depletion time is not included. I added it here to make it consistent. Anyway, we should reconsider whether the depletion time is a parameter of the resonator object or if it's a macro-based parameter, because for many of these active depletion techniques, we wouldn't want to have this extra wait, so we should leave it as zero (but just for the DRACHMA CLEAR techniques, etc.). Regarding this node, we can either leave it to the user to be responsible for choosing the right depletion time in the state, or add it as a node parameter. Then one can play and see how the efficiency degrades or improves with the depletion time, which also might be interesting. What do you think?

qualibration_graphs/superconducting/calibrations/1Q_calibrations/08c_readout_quantum_efficiency.py
paulQM: Resolved in 420b9077: the node now forces `node.parameters.multiplexed = False` unconditionally at the top of `create_qua_program`, regardless of what the GUI/graph config sets. So the batch loop always runs one qubit at a time now, and `multiplexed` is documented in `Parameters` as a no-op for this node.

qualibration_graphs/superconducting/calibrations/1Q_calibrations/08c_readout_quantum_efficiency.py
paulQM: Resolved in 81bb47a2: rewrote the paragraph after the eta formula to state explicitly what eta measures - the fraction of the qubit's coherence loss recoverable from the digitized signal, i.e. how far the SNR falls short of the quantum-limited noise floor - and to say plainly that it is neither the amplifier's own gain/noise figure in isolation nor readout fidelity. It also spells out the paper's three conditions for the true chain efficiency (optimal weights, active depletion, single-qubit probing) versus what this node actually does (existing weights, passive depletion wait).

PR #587, opened 2026-09-16
Measurement-induced dephasing matrix node

paulQM:
## Summary
- Add the `23_measurement_induced_dephasing_matrix` calibration node under `1Q_calibrations`, plus its `calibration_utils/measurement_induced_dephasing_matrix` module (parameters, analysis, plotting).
- Implements the protocol of Fig. 6 of Phys. Rev. Applied 23, 054089 ([arXiv:2412.14853](https://arxiv.org/abs/2412.14853)): a Hahn echo on qubit Qi with a readout pulse of relative amplitude xi inserted in the first half of the echo on resonator Rj. The contrast decay `c(xi) = c0 * exp(-Gamma * tau_p * xi**2)` gives the dephasing rate for each (Qi, Rj) pair.
- The diagonal holds self-dephasing rates (MHz scale) and the off-diagonal holds readout crosstalk (Hz scale). Because the two scales differ by orders of magnitude, the amplitude axis differs per pair and is stored as a 2D `xi` coordinate rather than as a dimension.
- The node is diagnostic: it reports the matrix and does not update the QUAM state.

## Notes
- Only new files. Nothing existing is modified or deleted.
- Renumbered from `35_` to `23_` to fit this repo's 1Q sequence, and the prerequisite node references in the docstring were remapped to this repo's numbering (04b/10b, 07/08a/08b, 06b).

## Test plan
- [ ] Run in simulation mode against a flux-tunable transmon QUAM state.
- [ ] Run on hardware and confirm the contrast fit and dephasing matrix figures render.


qualibration_graphs/superconducting/calibrations/1Q_calibrations/23_measurement_induced_dephasing_matrix.py
```diff
@@ -0,0 +1,382 @@
...
+
+                    probe_cycles = probe_lengths_ns[driven_qubit.name] // 4
+                    # Delay between the end of the first half's ring-down and the start of the second
+                    # probe: the rest of the first half, then the x180 played by the measured qubit.
+                    second_probe_delay_cycles = idle_time_cycles - probe_cycles - depletion_cycles + x180_cycles
+
+                    with for_each_(xi, xi_values[i, j].tolist()):
+                        with for_each_(phi, phases.tolist()):
+                            # Qubit initialization
+                            reset_frame(qubit.xy.name)
```
Deepakkhurrana, line 234: On the off-diagonal, you only reset the measured qubit. The resonator being driven belongs to a different qubit, which is never put back in a known state. Photon number at a given ξ depends on whether that qubit is in |g⟩ or |e⟩ — isn’t that mixing the crosstalk you’re trying to measure? Should each shot also reset the driven qubit?

qualibration_graphs/superconducting/calibration_utils/measurement_induced_dephasing_matrix/parameters.py
```diff
@@ -0,0 +1,203 @@
...
+    single-shot readout amplitude. Swept linearly from 0. Default is 1.0."""
+    xi_min_diagonal: float = 1e-2
+    """Minimum non-zero readout amplitude scaling for self-dephasing (i == j) pairs. The diagonal is
+    swept logarithmically because Gamma_ii is orders of magnitude larger. Default is 1e-2."""
+    xi_max_diagonal: float = 0.3
+    """Maximum readout amplitude scaling for self-dephasing (i == j) pairs. The decay is only well
+    conditioned when the exponent Gamma_ii * tau_p * xi_max**2 reaches order unity; for a typical
+    Gamma_ii of tens of MHz and a microsecond readout pulse that means xi_max of a few tenths, not a
+    few hundredths. Default is 0.3."""
+    readout_len_in_ns: Optional[int] = None
```
Deepakkhurrana, line 34: If `"readout"` is a shaped waveform (non square pulse), does `readout_len_in_ns` actually keep the same photon number, or only change the envelope?

paulQM: it will stretch the envelope, to use carefully

qualibration_graphs/superconducting/calibrations/1Q_calibrations/23_measurement_induced_dephasing_matrix.py
paulQM: good point

paulQM: Good catch, you're right. As of 4fbdaff5, every qubit is reset at the start of each shot, so the qubit of the driven resonator always starts in |g⟩ and its dispersive shift no longer gets mixed into the off-diagonal terms. With active reset this adds one conditional reset per qubit to each shot, but the resets run in parallel.

Deepakkhurrana (approved):
LGTM

PR #588, opened 2026-09-16
Add joint readout duration x power optimization node (08c)

paulQM:
## Summary

Ports the 2D readout optimization node from `CS_installations` into this repo as `08c_readout_duration_power_optimization`.

A single acquisition sweeps the readout amplitude and, through accumulated demodulation, returns the running integral after every chunk. Every integration duration up to the maximum therefore comes out of the same shots the amplitude sweep already paid for, so the duration axis costs no extra acquisition time over `08b`.

The operating point per qubit is the fidelity maximum over grid points that pass two gates: a non-outlier fraction floor and a blob-variance-ratio ceiling. The variance gate catches the characteristic high-power failure, where the excited blob smears into an arc while the ground blob stays tight — the non-outlier fraction still looks healthy there, but the fidelity number is meaningless.

## Changes

New:
- `calibrations/1Q_calibrations/08c_readout_duration_power_optimization.py`
- `calibration_utils/readout_duration_power_optimization/` (`__init__`, `analysis`, `parameters`, `plotting`, `qua_sequence`)
- `calibration_utils/common_utils/path_signature.py` — `accumulated_demod_batches`, `declare_path_arrays`, `preflight_path_signature`
- `tests/test_readout_duration_power_optimization.py`

Modified:
- `calibration_utils/common_utils/__init__.py` — exports the three helpers above
- `pyproject.toml` — adds `scikit-learn>=1.5.0`; the per-grid-point blob fit uses `sklearn.mixture.GaussianMixture`
- `calibration_utils/iq_blobs/plotting.py` — adds `plot_histograms` with a `log_scale` option

## Adapted for this repo

- **Renumbered `15b` to `08c`** to match the `08a`/`08b` readout family here, and updated the docstring cross-references (`node 15` becomes `node 08b`).
- **Dropped the "re-run node 14b" instructions.** This repo has no readout-weights-optimization node. Custom integration weights are still reset to the defaults, which the length change requires either way.
- **`plot_historams` was renamed to `plot_histograms`** and gained `log_scale`. `plot_historams` is kept as an alias, so `07_iq_blobs.py` is unaffected.

## Test plan

- [x] `pytest qualibration_graphs/superconducting/tests/` — 21 passed
- [x] Every import in the node resolves
- [x] `black --line-length 120 --check` clean on all touched files
- [ ] Run on hardware

Coverage is analysis-only: the per-grid-point blob fit, the operating-point gates, the per-duration volts conversion, and the 4 ns chunk-grid validator. The execute path is not covered and the node has not been run against hardware in this repo.

## Notes for the reviewer

- This repo's `calibration_utils` is a namespace package with no `__init__.py`. If you have `CS_installations` installed in the same environment, its regular `calibration_utils` package will shadow this one entirely and you will silently test the wrong files. Use a clean environment.
- `uv pip install -e qualibration_graphs/superconducting` fails on `main`: `pyproject.toml` points `readme` at `../README.md`, outside the project directory, which hatchling rejects. Pre-existing and untouched here.
- `17c_qubit_flux_short_distortion.py` already fails `black` on `main`. Left alone.


qualibration_graphs/superconducting/calibrations/1Q_calibrations/08c_readout_duration_power_optimization.py
```diff
@@ -0,0 +1,399 @@
...
+            I_e_st = [declare_stream() for _ in range(num_qubits)]
+            Q_e_st = [declare_stream() for _ in range(num_qubits)]
+
+            for measurement_batch in measurement_batches:
+                # Initialize the QPU in terms of flux points (flux tunable transmons and/or tunable couplers)
+                for qubit in measurement_batch.values():
+                    node.machine.initialize_qpu(target=qubit)
+                align()
+
+                paths_g = {i: declare_path_arrays(num_durations) for i in measurement_batch}
```
Deepakkhurrana: `paths_g`, `paths_e`, and `scratch` are declared inside the batch loop, after `initialize_qpu` / `align()`. QUA does not allow that.

`multiplexed` defaults to `False`, so this is one batch per qubit: the second qubit’s `declare` lands after the first qubit’s `for_` loop and will fail compilation.

Declare those arrays once for `range(num_qubits)` before the batch loop, then only play/measure inside it, same as `Ig[i]` in 08b. The tests never compile the QUA program, so they will not catch this. The unit tests never build the QUA program, so they will not catch this.

qualibration_graphs/superconducting/calibrations/1Q_calibrations/08c_readout_duration_power_optimization.py
```diff
@@ -0,0 +1,399 @@
...
+            # The length goes first: every quantity below is derived AT that length, so
+            # writing it afterwards would leave the thresholds describing the old pulse. When
+            # the length is not updated the analysis has already pinned the search to the
+            # current length, so `optimal_duration` equals `operation.length` either way and
+            # the demod-unit conversions below are consistent with the pulse that will run.
+            if node.parameters.update_readout_length:
+                operation.length = int(fit_results["optimal_duration"])
+            # Custom weights span the previous length and no longer tile the pulse.
+            # Reset them to the defaults and say so in the log.
+            if has_custom_integration_weights(operation):
```
Deepakkhurrana: **Custom weights are reset even when length is not updated**

The sweep must flatten weights while the pulse is temporarily lengthened — `readout_config_override` already does that and restores the originals.

`update_state` then always writes the defaults whenever custom weights are present, including when `update_readout_length` is off and `operation.length` does not change. That throws away a prior weights calibration on an amplitude-only run.

Consider only resetting weights in the state when the length actually changes:

```python
if node.parameters.update_readout_length:
    operation.length = int(fit_results["optimal_duration"])
    if has_custom_integration_weights(operation):
        node.log(...)
        set_integration_weights(operation, DEFAULT_INTEGRATION_WEIGHTS)
```

qualibration_graphs/superconducting/calibrations/1Q_calibrations/08c_readout_duration_power_optimization.py
```diff
@@ -0,0 +1,399 @@
...
+                        for i, qubit in measurement_batch.items():
+                            qubit.reset(node.parameters.reset_type, node.parameters.simulate)
+                        align()
+                        # Excited state: qubit readout
+                        for i, qubit in measurement_batch.items():
+                            # Play the x180 gate to put the qubit in the excited state
+                            qubit.xy.play("x180")
+                            # Align the elements to measure after playing the qubit pulses.
+                            qubit.align()
+                            measure_accumulated_scaled(qubit, operation, paths_e[i], samples_per_chunk, a)
```
Deepakkhurrana: When `multiplexed=True` and qubits don’t all fit in one accumulated-demod batch, `accumulated_demod_batches` sizes the batch as if leftover resonators still play a readout (no demod). 08c only measures the qubits in the current batch.

Was idle play intentional so multiplexed crosstalk matches a real readout, or should unused resonators stay dark?

qualibration_graphs/superconducting/calibration_utils/readout_duration_power_optimization/qua_sequence.py
```diff
@@ -0,0 +1,126 @@
...
+        operation: Name of the resonator operation to acquire with.
+        path_arrays: The four arrays from ``declare_path_arrays``, filled with the
+            accumulated ``(II, IQ, QI, QQ)`` components.
+        samples_per_chunk: Chunk size in units of 4 ns.
+        amplitude_scale: QUA ``fixed`` prefactor applied to the readout pulse amplitude.
+    """
+    II, IQ, QI, QQ = path_arrays
+    pulse = qubit.resonator.operations[operation]
+    labels = list(pulse.integration_weights_mapping)
+    measure(
```
Deepakkhurrana: `measure_accumulated_scaled` reimplements accumulated dual-demod with a raw `measure(...)`. 08b/07 go through `qubit.resonator.measure(...)`, and QUAM already has `resonator.measure_accumulated(...)` with the same weight/output pairing.

Any reason not to call that helper instead, so this stays on the same channel API?

Deepakkhurrana (changes requested):
Some nits:

Remaining items that are worth posting. 1–4 are already covered.

- 08c only uses `accumulated_demod_batches`, `declare_path_arrays`, and `preflight_path_signature`. The rest (Levy area, `reduce_to_signature`, feature streams) is unused here. Can the unused path-signature surface stay out of `common_utils` until a node actually needs it?

- 07 types this as `Literal["readout", "readout_QND"]`. Worth matching so the GUI does not accept an arbitrary pulse name?

qualibration_graphs/superconducting/calibrations/1Q_calibrations/08c_readout_duration_power_optimization.py
paulQM: Addressed in 248c7a44.

`paths_g`, `paths_e` and `scratch` are now declared once for `range(num_qubits)` before the batch loop, indexed `paths_g[i]`, same as `Ig[i]` in 08b. The batch loop only plays and measures.

One correction though: this was not a compile failure. On qm-qua 1.4.1, the version the lock pins, every `declare()` is hoisted to program scope no matter where it sits in the Python source — including inside a `with for_(...)` body. I built both versions and `generate_qua_script` gives byte-identical programs. The batch loop is also a plain Python loop rather than a QUA `for_`, so there was never an ordering problem for the compiler to see.

I kept the change anyway: it reads better and it stops the node depending on that hoisting behaviour. I just did not want to leave a comment in the file claiming QUA forbids something it allows.

On the testing gap you flagged, I did compile the program rather than rely on the unit tests. I generated a throwaway 4-qubit MW-FEM QUAM and ran the node through `create_qua_program` and `generate_config()` in both multiplexed modes. That harness is not committed — if you want the repo to carry a QUAM fixture so the test suite can build the program, say so and I will open it as a separate piece of work.

qualibration_graphs/superconducting/calibrations/1Q_calibrations/08c_readout_duration_power_optimization.py
paulQM: Addressed in 248c7a44, exactly as you suggested.

The `has_custom_integration_weights` block moved inside `if node.parameters.update_readout_length:`, so the defaults are only written when `operation.length` actually changes. An amplitude-only run now leaves a previous weights calibration alone.

The sweep itself is unchanged: `readout_config_override` still flattens the weights while the pulse is temporarily lengthened and still restores the originals on the way out.

I also updated the two places in the node docstring that described the old behaviour — the "Notes" bullet on custom weights, and the "State update" line, which now reads `reset to the defaults (only if update_readout_length)`.

qualibration_graphs/superconducting/calibration_utils/readout_duration_power_optimization/qua_sequence.py
paulQM: Addressed in 248c7a44 — you are right, it is equivalent, so `measure_accumulated_scaled` is gone and the node calls the channel method:

```python
qubit.resonator.measure_accumulated(
    operation,
    amplitude_scale=a,
    segment_length=samples_per_chunk,
    qua_vars=paths_g[i],
)
```

What I checked before swapping:

- **Pairing is identical.** QUAM emits `(labels[0], out1)`, `(labels[1], out2)`, `(labels[2], out1)`, `(labels[0], out2)` — the same four `demod.accumulated` calls in the same order as the hand-rolled version.
- **Units match.** `segment_length` is in units of 4 ADC samples and goes straight to `demod.accumulated`, so it is the same quantity as `samples_per_chunk`.
- **Amplitude scaling matches.** `add_amplitude_scale_to_pulse_name` resolves a QUA `fixed` to `pulse * amp(scale)`.
- **Availability.** It lives on `_InComplexChannel`, so both `InOutIQChannel` and `InOutMWChannel` resonators have it, and it is present across the whole `quam>=0.5.0a1` range in `pyproject.toml`.

Passing our own `qua_vars` is what lets the arrays stay declared outside the batch loop; QUAM only declares its own when the argument is omitted. I confirmed on a generated MW-FEM QUAM that the emitted `measure` statement carries the right weights, outputs and `amplitude_scale`.

paulQM:
Thanks for the review. Five of the six items are addressed in 248c7a44; the two from the review body are below, and the remaining one needs a decision from you.

**Unused path-signature surface — removed.** Deleted `measure_path`, `reduce_to_signature`, `path_quadratures`, `save_path`, `declare_feature_streams`, `feature_handle_name` and `path_handle_name`. I checked first: nothing outside the module referenced any of them. What is left is the three helpers 08c uses plus their internals (`flat_weights_reason`, `n_chunks_for`, `_fem_key`, `_block_limit`, `max_measured_per_batch`).

I went one small step further than you asked, so flag it if you disagree. With the signature machinery gone, nothing in the module computes a signature, so I renamed `path_signature.py` to `accumulated_demod.py` and `preflight_path_signature` to `preflight_accumulated_demod`, and cut the Levy-area and Stratonovich material out of the module docstring. `declare_path_arrays` keeps its name — the accumulated demod does still return a path. Happy to revert just the rename if you would rather keep the old names for the eventual port back.

**Pulse name typing — matched.** `operation: Literal["readout", "readout_QND"] = "readout"`, same as 07.

**Still open: the idle-readout question.** Your reading is right, and the mismatch is real. `max_measured_per_batch` solves `4m + (n − m) ≤ 16`, which charges one block to every selected qubit the batch is not measuring — only true if those resonators still play their tone. The node loops over `measurement_batch.items()` only, so they are silent and cost nothing.

With eight qubits on one MW-FEM that caps the batch at two measured qubits and reserves six blocks nobody spends, so the batch uses 8 of 16 blocks and the run takes four batches instead of two. At eleven qubits it is eleven batches instead of three, and at thirteen the "no room for even one accumulated demodulation" error fires when under the dark reading it should not.

Two consistent repairs, and I do not have the chip knowledge to pick:

1. **Change the node to match the helper.** Play the readout, without demodulating, on the selected qubits outside the batch. Every batch then sees the full feedline, which matches a production readout and keeps the batches consistent with each other. Acquisition time unchanged.
2. **Change the helper to match the node.** Size the batch at `limit // 4`, drop `BLOCKS_PER_IDLE_READOUT`, and rewrite the docstring that currently promises the crosstalk environment is unchanged. Halves the run at eight qubits per FEM, but the calibration keeps happening on a quieter line than production, and a different one per batch.

The question I cannot answer from the code: on your chips, is the shift from a two-tone to an eight-tone feedline big enough to move the optimal amplitude or the discrimination threshold? If yes, option 1. If not, option 2 is free speed. Note none of it bites at the defaults, since `multiplexed` defaults to `False`.

**Verification on this push:** 25 tests pass, `black --line-length 120` clean. I also compiled the QUA program rather than trusting the unit tests — generated a throwaway 4-qubit MW-FEM QUAM and ran the node through `create_qua_program` and `generate_config()` in both multiplexed modes. Still not run against real hardware.

paulQM:
Item 3 (the idle-readout block accounting) is now addressed, in `7ab9bf35`. We went with option 2: only measured qubits are charged against the PPU budget.

**What changed.** `max_measured_per_batch` used to solve `4m + (n - m) <= limit`, charging one block to every selected qubit the batch was *not* measuring, on the assumption that those resonators still play an idle readout. Node 08c never plays them — it loops over `measurement_batch.items()` only — so they are silent and cost nothing. The batch is now sized at `limit // 4`, which is 4 qubits per MW-FEM and 5 per OPX+ however many qubits are selected. `BLOCKS_PER_IDLE_READOUT` is gone.

**What it buys.** On one MW-FEM:

| qubits on the FEM | batches before | batches now |
|---|---|---|
| 4 | 1 | 1 |
| 5 | 2 | 2 |
| 8 | 4 | 2 |
| 11 | 11 | 3 |
| 13+ | `ValueError` | 4 |

Eight qubits on a FEM halves the run, and a fully-loaded FEM no longer refuses to run at all.

**The cost, written down rather than fixed.** The out-of-batch resonators stay dark, so the feedline carries fewer tones than a production multiplexed readout. That is a real difference in the crosstalk environment, and if it matters on your chip the optimum this node finds can sit slightly off the one a full multiplexed readout would give. It is now a `Notes` bullet in the node docstring and a paragraph in the module docstring, so nobody has to rediscover it. If you'd rather have the production environment than the shorter run, the other repair is to play idle readouts on the out-of-batch qubits and keep the old arithmetic — say the word and I'll swap it.

**Tests.** Six new tests in `tests/test_readout_duration_power_optimization.py` pin the batch shape: the batch size no longer shrinks as the selection grows, eight qubits on one FEM give two batches of four, every selected qubit is measured exactly once, two FEMs pack into one batch, and `multiplexed=False` still measures one at a time. The whole file passes (31 tests), and the node still builds its QUA program against a generated QUAM in both multiplexed modes.

One related thing I noticed while in here, not part of this PR: `calibration_utils/readout_optimization_3d/measurement_batching.py` carries the same idle-readout assumption in `get_max_accumulated_readouts`. No calibration node imports it today, so nothing is broken, but whoever writes that node will want the same decision made explicitly.

paulQM:
@Deepakkhurrana heads up: we went with option 2 for the block-accounting point, implemented in `7ab9bf35` (CI green).

Batches are now sized by measured qubits only, `limit // 4`, so 4 per MW-FEM and 5 per OPX+ however many qubits are selected. The out-of-batch resonators stay dark rather than playing an idle readout. Eight qubits on one FEM now run in 2 batches instead of 4, and 13 or more no longer raise.

The trade-off — fewer tones on the feedline than a production multiplexed readout — is recorded in the node and module docstrings rather than fixed, and six new tests pin the batch shape. Details and the before/after table are in the comment just above: https://github.com/qua-platform/qua-libs/pull/588#issuecomment-5729056944

All six review items are now addressed, so this is ready for another look when you have a moment.

qualibration_graphs/superconducting/calibrations/1Q_calibrations/08c_readout_duration_power_optimization.py
Deepakkhurrana: I think we dont need to do the QUAM fixture fir now as test suite is uncooked anyway. I will leave it as it is now.

Deepakkhurrana:
> @Deepakkhurrana heads up: we went with option 2 for the block-accounting point, implemented in `7ab9bf35` (CI green).
> 
> Batches are now sized by measured qubits only, `limit // 4`, so 4 per MW-FEM and 5 per OPX+ however many qubits are selected. The out-of-batch resonators stay dark rather than playing an idle readout. Eight qubits on one FEM now run in 2 batches instead of 4, and 13 or more no longer raise.
> 
> The trade-off — fewer tones on the feedline than a production multiplexed readout — is recorded in the node and module docstrings rather than fixed, and six new tests pin the batch shape. Details and the before/after table are in the comment just above: [#588 (comment)](https://github.com/qua-platform/qua-libs/pull/588#issuecomment-5729056944)
> 
> All six review items are now addressed, so this is ready for another look when you have a moment.

Option 2 is fine.

It is not production multiplexing: out-of-batch resonators stay dark, so the feedline is quieter than a full multiplexed readout and the optimum can sit slightly off. That is acceptable here, especially with multiplexed defaulting to False.

Thanks for documenting it.

Deepakkhurrana (approved):
LGTM!

PR #595, opened 2026-09-18
Update/port GEF readout calibrations (15, 14a, 14b)

Deepakkhurrana:
## Summary
- Port CS IQ-blob GEF fixes into `15_iq_blobs_gef`: `ensure_gef_readout_pulse`, thermal or `active_gef` with IF restore, `abs(anharmonicity)` + `keep_phase` on the EF hop, and `gef_centers` inverted with `readout.length` (not `readout_GEF.length`).
- Replace `14_gef_readout_frequency_optimization` with `14a` (CS 30a): thermal-only frequency sweep, graded argmax, write `GEF_frequency_shift` and informational `f_12`.
- Add `14b_gef_readout_power_optimization` (CS 30b): 1D `readout_GEF` amplitude sweep using min{d_ge, d_ef, d_gf}.
- Shared helpers: `gef_readout_pulse.py` (pulse / IF / reset) and `curve_quality.argmax_with_quality` so a flat/dead curve is not written to state.


paulQM (approved):
LGTM!

